PR #16, opened 2021-05-22
fix: show a better error message when server is not available

jlipps:


KazuCocoa (approved):

app/renderer/actions/Session.js
```diff
@@ -117,7 +117,7 @@ export function showError (e, methodName, secs = 5) {
   } else {
     errMessage = i18n.t('Could not start session');
   }
-  if (errMessage === 'ECONNREFUSED') {
+  if (errMessage === 'ECONNREFUSED' || errMessage.includes('Failed to fetch')) {
```
mykola-mokhnach, line 120: this might fail if errMessage is a number (`errMessage = e.code`). I would rather use _.include

Issue #59, opened 2021-06-28
Can not move inspector window, if the the inspector window has focus.


stephenjg24-git:
## Appium or Appium Desktop?

You are reporting an issue at the Appium Desktop repository. Appium Desktop is a wrapper around Appium. If you are having trouble running tests, it is much more likely that the problem you are encountering is not a problem with Appium Desktop but with Appium. For that reason we require you to have tried your testcase with Appium CLI before reporting issues. Check the checkbox below to confirm that you have proven your issue does _not_ reproduce on Appium itself:

- [ ] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## The problem

_Replace this section with a description of your issue_

## Environment

- I am running Appium Desktop version _<insert version here>_.
- I am on (pick one):
    - [ x] Mac
    - [ ] Windows
    - [ ] Linux

I am running beta 5, this is a UI issue with inspector, not related to anything appium.

Launch the Inspector
click in the top area of the window, keep holding the left ouse button down and try to drag the window to a new area.
This fails

Workaround
Click on an open area on the screen, to move focus away from the inspector
Then repeat the above steps,
This works.

jlipps:
Which mac version do you have? I can't replicate this bug.

stephenjg24-git:
I have the beta 5 version.

sme81:
Getting same bug, workaround doesnt seem to be working. same app version. Mac Catalina

jlipps:
Can you record a video of the bug, with the devtool console log open so I can see any logging that comes through? (Right click > inspect on any element, then select the console).

sme81:
Gist [here](https://gist.github.com/sme81/fc2993a7eb83db32eca4b4c13e369d00) of devtools console after initial opening.

Video here, cleared console, doesnt log anything new when moving...


https://user-images.githubusercontent.com/48491431/126275450-ff2ba880-9410-4324-a659-8e767536fd93.mov

jlipps:
OK thanks. I'm going to try to publish a new version of the inspector. Let's see if it gets rid of the issue.

jlipps:
@sme81 for some reason our publishing flow isn't working. Could you try this instead: https://inspector.appiumpro.com/ (if you need to you can open it up in its own browser window and move and resize to your heart's content).

sme81:
@jlipps yeah opening in browser loads and can move that window about. 

[...]

PR #61, opened 2021-06-29
fix: can not move window.

youngfreeFJS:
#59 fix this issue.
Also, Fixed a typo for reference errors.


youngfreeFJS:
@jlipps I have the same problem. 
#59 he said the **_inspector window_**, means the _**Session**_ page.

PR #62, opened 2021-06-29
fix: can not move window.

youngfreeFJS:
@jlipps  I have the same problem.
#59 he said the **_inspector window_**, means the **_Session_** page.


jlipps (approved):
hmm. which platform are you on? on Mac I can drag already with the current code.

PR #66, opened 2021-07-07
feat: add a bundle target for web browsers

jlipps:
This allows us to run e.g. `npm run build:prod:browser`, and get output in `dist-browser` which can be served as a set of static files from a web server, enabling Appium Inspector to run in a browser rather than only as a standalone Electron app.

To accomplish this I had to implement some of the node-only or electron-only interfaces using web technologies, but otherwise it was fairly straightforward.


app/renderer/polyfills/index.js
```diff
@@ -0,0 +1,37 @@
+let log, settings, clipboard, shell, remote, ipcRenderer,
+    i18NextBackend, i18NextBackendOptions;
+
+function buildForBrowser () {
+  if (process.env.BUILD_BROWSER) {
+    return true;
+  }
+
+  if (typeof navigator !== 'undefined') {
```
mykola-mokhnach, line 9: this could be simplified to `return typeof navigator !== 'undefined' && !/electron/i.test(navigator.userAgent)`

mykola-mokhnach (approved):

mykola-mokhnach:
shall we do something about tests?

KazuCocoa (commented):
Great idea, yey
This change itself seems good to me, but CI is failing. I guess something electron's `app`...?

app/renderer/polyfills/index.js
jlipps: ah yes good point. i was debugging something and left the debugging structure.

app/renderer/polyfills/index.js
```diff
@@ -0,0 +1,35 @@
+let log, settings, clipboard, shell, remote, ipcRenderer,
+    i18NextBackend, i18NextBackendOptions;
+
+function buildForBrowser () {
+  if (process.env.BUILD_BROWSER) {
```
mykola-mokhnach, line 5: we could reduce this whole function body to a single return

mykola-mokhnach (approved):

app/renderer/polyfills/index.js
jlipps: i know, i found this a bit easier to follow

PR #79, opened 2021-07-21
feat: enable state init and auto-start-session via the query string

jlipps:
This will enable us to automatically set state for ADI via a deep link or a regular URL (in browser mode)

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -56,11 +56,15 @@ export const SET_PROVIDERS = 'SET_PROVIDERS';
 
 export const SET_ADD_VENDOR_PREFIXES = 'SET_ADD_VENDOR_PREFIXES';
 
+export const SET_STATE_FROM_URL = 'SET_STATE_FROM_URL';
+
 
 const CAPS_NEW_COMMAND = 'appium:newCommandTimeout';
 const CAPS_CONNECT_HARDWARE_KEYBOARD = 'appium:connectHardwareKeyboard';
 const CAPS_NATIVE_WEB_SCREENSHOT = 'appium:nativeWebScreenshot';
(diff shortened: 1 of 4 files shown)


app/renderer/actions/Session.js
```diff
@@ -831,3 +835,41 @@ export function setAddVendorPrefixes (addVendorPrefixes) {
...
+export function initFromQueryString () {
+  return async (dispatch, getState) => {
+    if (!isFirstRun) {
+      return;
+    }
+
+    isFirstRun = false;
+
+    let url = window.location.href;
+    if (url.indexOf('?') === -1) {
```
KazuCocoa, line 848: `.includes`?

app/renderer/actions/Session.js
```diff
@@ -831,3 +835,41 @@ export function setAddVendorPrefixes (addVendorPrefixes) {
...
+      return;
+    }
+    // remove the weird anchor thing react or the router puts at the end
+    url = url.replace(/#\/$/, '');
+    const query = url.slice(url.indexOf('?') + 1);
```
mykola-mokhnach, line 853: why should we do the same operation twice?

mykola-mokhnach: I also assume it would be more appropriate to use [URL](https://url.spec.whatwg.org/#api) class

app/renderer/actions/Session.js
```diff
@@ -831,3 +835,41 @@ export function setAddVendorPrefixes (addVendorPrefixes) {
...
+    if (initialState) {
+      try {
+        const state = JSON.parse(initialState);
+        dispatch({type: SET_STATE_FROM_URL, state});
+      } catch (e) {
+        showError(new Error('Could not parse initial state from URL'), null, 0);
+      }
+    }
+
+    if (autoStartSession === '1') {
```
mykola-mokhnach, line 867: magic strings?

PR #100, opened 2021-08-24
Misc updates and fixes

jlipps:
Trying to fix a bunch of small issues I've noticed


app/renderer/components/Session/CapabilityEditor.js
```diff
@@ -52,10 +57,18 @@ export default class NewSessionForm extends Component {
     setCapabilityParam(index, 'value', translatedValue);
   }
 
+  componentDidUpdate () {
+    if (!this.latestCapField.current.input.value) {
```
mykola-mokhnach, line 61: is current always defined? should we use optional chaining instead?

mykola-mokhnach (approved):
it would be easier to review the PR if split to smaller parts

PR #118, opened 2021-08-31
fix: ensure connection auth details are sent to w2d correctly (fix #22)

jlipps:

PR #129, opened 2021-09-07
Fix conns use key

KazuCocoa:
I guess https://github.com/appium/appium-inspector/pull/118#issuecomment-914027306 did not use `username` and `accessKey` as their basic auth.


KazuCocoa:
I merged this change into jlipps/fix-conns cc @jlipps

PR #139, opened 2021-09-15
Use bstack:options for setting the 'source' capability for BrowserStack to comply with W3C

huzaifaiftikhar:
**Issue**
Currently, on using the appium-inspector we get the following error.

image

**Fix**
Use `bstack:options` for setting the 'source' capability for BrowserStack to comply with W3C spec.


jlipps (approved):

jlipps:
thanks!

Issue #154, opened 2021-09-26
Fix Hybrid Element positon determination


wswebcreation:
## Appium or Appium Desktop?

You are reporting an issue at the Appium Desktop repository. Appium Desktop is a wrapper around Appium. If you are having trouble running tests, it is much more likely that the problem you are encountering is not a problem with Appium Desktop but with Appium. For that reason we require you to have tried your testcase with Appium CLI before reporting issues. Check the checkbox below to confirm that you have proven your issue does _not_ reproduce on Appium itself:

- [x] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## Environment

- I am running Appium Desktop version _2021.9.1_.
- I am on (pick one):
    - [x] Mac
    - [ ] Windows
    - [ ] Linux

## The problem
HTML elements in an Hybrid app are not positioned properly, this has been fixed in Appium Desktop, see
https://github.com/appium/appium-desktop/pull/1906. This needs to be ported to Appium Inspector

wswebcreation:
Fixed

avenstewart:
I have updated to the latest 1.22 Appium Server GUI as well as the latest inspector patch 2021.9.2, and I am still experiencing this issue [Android 9, Chrome 96] on a hybrid Cordova application. Appium components deleted from android device, app cache cleared, Appium server and inspector restarted, device restarted, app reinstalled, error still persists - along with https://github.com/appium/appium-inspector/issues/150.

Issue #153, opened 2021-09-26
Hybrid mode is not working for Android Hybrid apps


wswebcreation:
## Appium or Appium Desktop?

You are reporting an issue at the Appium Desktop repository. Appium Desktop is a wrapper around Appium. If you are having trouble running tests, it is much more likely that the problem you are encountering is not a problem with Appium Desktop but with Appium. For that reason we require you to have tried your testcase with Appium CLI before reporting issues. Check the checkbox below to confirm that you have proven your issue does _not_ reproduce on Appium itself:

- [x] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## Environment

- I am running Appium Desktop version _2021.9.1_.
- I am on (pick one):
    - [ ] Mac
    - [ ] Windows
    - [ ] Linux

## The problem
Android hybrid apps can't be inspected because the dropdown is not shown

## Root Cause:
This [line](https://github.com/appium/appium-inspector/blob/main/app/renderer/lib/appium-client.js#L235) of code returns the contexts. The problem is that it returns different data for Android as it does for iOS

**iOS result**
```
    [
      {
        "id": "NATIVE_APP"
      },
      {
        "id": "WEBVIEW_70122.3",
        "title": "WebdriverIO · Next-gen browser and mobile automation test framework for Node.js | WebdriverIO",
        "url": "https://webdriver.io/",
        "bundleId": "org.wdioNativeDemoApp"
      }
    ]
```

**Android result**
```
  [
      {
        "proc": "@webview_devtools_remote_7414",
        "webview": "WEBVIEW_7414",
        "info": {
          "Android-Package": "com.wdiodemoapp",
          "Browser": "Chrome/74.0.3729.185",
          "Protocol-Version": "1.3",
          "User-Agent": "Mozilla/5.0 (Linux; Android 10; Android SDK built for x86 Build/QSR1.200403.001; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/74.0.3729.185 Mobile Safari/537.36",
          "V8-Version": "7.4.288.28",
          "WebKit-Version": "537.36 (@22955682f94ce09336197bfb8dffea991fa32f0d)",
          "webSocketDebuggerUrl": "ws://127.0.0.1:10900/devtools/browser"
        },
        "pages": [
          {
[...]

Issue #152, opened 2021-09-26
Older versions of Chrome are not W3C compatible


wswebcreation:
## Appium Inspector
You are reporting an issue at the Appium Desktop repository. Appium Desktop is a wrapper around Appium. If you are having trouble running tests, it is much more likely that the problem you are encountering is not a problem with Appium Desktop but with Appium. For that reason we require you to have tried your testcase with Appium CLI before reporting issues. Check the checkbox below to confirm that you have proven your issue does _not_ reproduce on Appium itself:

- [x] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## Environment

- I am running Appium Inspector version _2021.9.1_.
- I am on (pick one):
    - [x] Mac
    - [ ] Windows
    - [ ] Linux

## The problem
Appium detects which version of of ChromeDriver needs to be used based on the Webview / Chrome browser and that might result in an older version of ChromeDriver which doesn't support W3C commands like the `executeScript`.
This only happens with Hybrid Android apps or with Chrome

wswebcreation:
Decided not support this anymore , see https://github.com/appium/appium-inspector/pull/155#issuecomment-929366327

Issue #151, opened 2021-09-26
`executeScript` command doesn't accept a function


wswebcreation:
## Appium Inspector
You are reporting an issue at the Appium Desktop repository. Appium Desktop is a wrapper around Appium. If you are having trouble running tests, it is much more likely that the problem you are encountering is not a problem with Appium Desktop but with Appium. For that reason we require you to have tried your testcase with Appium CLI before reporting issues. Check the checkbox below to confirm that you have proven your issue does _not_ reproduce on Appium itself:

- [x] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## Environment

- I am running Appium Inspector version _2021.9.1_.
- I am on (pick one):
    - [x] Mac
    - [ ] Windows
    - [ ] Linux

## The problem
 The [`executeScript`](https://github.com/appium/appium-inspector/blob/main/app/renderer/lib/appium-client.js#L250)-command doesn't accept a function, it needs to be a string

wswebcreation:
@jlipps @KazuCocoa 

I'll also try to fix this in my PR

wswebcreation:
Fixed

PR #155, opened 2021-09-26
Fix hybrid app support

wswebcreation:
This PR fixes support for (Android) Hybrid apps, see https://github.com/appium/appium-inspector/issues/153. It also fixes the following issues:
- [x] Fix `Hybrid mode iOS not highlighting elements in screenshot`, see https://github.com/appium/appium-inspector/issues/150
- [x] Fix `executeScript command doesn't accept a function`, see https://github.com/appium/appium-inspector/issues/151
- [x] Port `Hybrid element position determination`, see https://github.com/appium/appium-inspector/issues/154

**Open discussion points:**
@mykola-mokhnach / @jlipps , I'd love to get some feedback on the following open points.

- How to handle older versions of Chrome(Driver). Emulators (and even some older Real Devices) don't always have Chrome 75+ versions which support W3C. This means that this might break, see TODO
- ChromeDriver can have 1 webview, but multiple pages. Now the `SelectElement.js` only switches context, and is not switching to the correct window handle in that webview. This can lead to incorrect data in the tree (of an inactive page). I thought this could be done with the `active`-property from the `description`, but that only works for Webviews, not for Chrome Tabs.

**TODO:**
- [ ] Fix `Older versions of Chrome are not W3C compatible`, see https://github.com/appium/appium-inspector/issues/152


app/renderer/lib/appium-client.js
```diff
@@ -202,8 +201,23 @@ export default class AppiumClient {
 
   async getWindowUpdate () {
     let windowSize, windowSizeError;
+    const {client:{capabilities:{deviceScreenSize, platformName}}} = this.driver
     try {
-      windowSize = await this.driver.getWindowRect();
+      // The call doesn't need to be made for Android for two reasons
+      // - when appMode is hybrid Chrome driver doesn't know this this command
```
mykola-mokhnach, line 207: typo in `this this`?

app/renderer/lib/appium-client.js
```diff
@@ -202,8 +201,23 @@ export default class AppiumClient {
...
+      // - the data is already on the driver
+      //   this.driver.client.capabilities.deviceScreenSize => "{width}x{length}"
+      if (platformName.toLowerCase() === 'android') {
```
mykola-mokhnach, line 210: can platformName be not a string?

app/renderer/lib/appium-client.js
```diff
@@ -202,8 +201,23 @@ export default class AppiumClient {
...
+        const widthHeight = deviceScreenSize.split('x');
```
mykola-mokhnach, line 211: const [width, height] = ...

app/renderer/lib/appium-client.js
```diff
@@ -202,8 +201,23 @@ export default class AppiumClient {
...
+        windowSize = {
+          width: widthHeight[0],
+          height: widthHeight[1],
+          x:0,
```
mykola-mokhnach, line 215: Please add spaces before zeroes

app/renderer/lib/appium-client.js
```diff
@@ -292,6 +306,55 @@ export default class AppiumClient {
...
+   * {
+   *   id: string;             // NATIVE_APP or the webview name
+   *   title?: string;         // the name of the page
+   *   url?: string;           // the url
+   *   bundleId?: string;      // for iOS
+   *   packageName?: string;   // for Android
+   *   handle?: string;        // the id of the active page in the webview of Android
+   * }[];
+   */
+  parseAndroidContexts (contexts) {
```
mykola-mokhnach, line 323: it would be useful to have a unit test for this helper

app/renderer/lib/appium-client.js
wswebcreation: Thanks

app/renderer/lib/appium-client.js
wswebcreation: Thanks

app/renderer/lib/appium-client.js
wswebcreation: Thanks

app/renderer/lib/appium-client.js
wswebcreation: not sure about that, it comes back from the capabilities, should I do a type check here?

mykola-mokhnach: I usually use lodash to make it safe: `_.lower(platformName) === 'android'`

wswebcreation: Done, thanks

app/renderer/lib/appium-client.js
```diff
@@ -229,15 +244,45 @@ export default class AppiumClient {
...
+            y: isAndroid ? viewportRect.top : Math.round(viewportRect.top / pixelRatio),
+          };
+        } else {
+          // Fallback
+        const el = await this.driver.findElement(
+          isAndroid ? 'xpath' : '-ios class chain',
+          isAndroid ? '//android.webkit.WebView' : '**/XCUIElementTypeWebView'
+        );
+          if ('elementId' in el) {
+            // Strange is that this command is failing, isn't it exposed by Web2Driver
```
wswebcreation, line 275: @jlipps 

Do you know why this command is not working, is this an issue with Web2Driver?

app/renderer/lib/appium-client.js
```diff
@@ -229,15 +244,45 @@ export default class AppiumClient {
...
+        if (viewportRect) {
+          // The viewport rectangles are based on the screen density,
+          // iOS needs CSS pixels
+          webViewPosition = {
+            x: isAndroid ? viewportRect.left : Math.round(viewportRect.left / pixelRatio),
+            y: isAndroid ? viewportRect.top : Math.round(viewportRect.top / pixelRatio),
+          };
+        } else {
+          // Fallback
+        const el = await this.driver.findElement(
```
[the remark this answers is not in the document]
wswebcreation: Fixed

app/renderer/lib/appium-client.js
```diff
@@ -229,15 +244,45 @@ export default class AppiumClient {
...
+          isAndroid ? 'xpath' : '-ios class chain',
+          isAndroid ? '//android.webkit.WebView' : '**/XCUIElementTypeWebView'
+        );
+          if ('elementId' in el) {
```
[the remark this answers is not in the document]
wswebcreation: Fixed

app/renderer/lib/appium-client.js
wswebcreation: Thanks, totally missed that, my bad, fixed it now

app/renderer/lib/appium-client.js
wswebcreation: Done

app/renderer/lib/webview-helpers.js
```diff
@@ -82,6 +82,11 @@ export function parseSource (source) {
   const scripts = $('script');
   scripts.remove();
 
-  // remove all `data-appium-desktop-` prefixes
-  return $.html().replace(/data-appium-desktop-/g, '');
+  return $.html()
+    // Remove all html entity / special chars that could break the parsing
+    .replace(/(&(.*?);)/ig, '')
```
mykola-mokhnach, line 87: I would still prefer to avoid using regexps over dom traversal

app/renderer/lib/webview-helpers.js
```diff
@@ -82,6 +82,11 @@ export function parseSource (source) {
   const scripts = $('script');
   scripts.remove();
 
-  // remove all `data-appium-desktop-` prefixes
-  return $.html().replace(/data-appium-desktop-/g, '');
+  return $.html()
```
mykola-mokhnach, line 85: would it be a good idea to also cover this function with a unit test?

wswebcreation: Yes, that's a good one, will create that later today

app/renderer/lib/webview-helpers.js
wswebcreation: Agree with that, but wonder how I could walk over the dom and remove that from the content, need to dive into this

app/renderer/lib/webview-helpers.js
mykola-mokhnach: the $ object is a [cheerio](https://github.com/cheeriojs/cheerio) parser instance. You could check https://stackoverflow.com/questions/21003384/how-to-remove-all-attributes-from-html for example on how to remove attributes from a document

app/renderer/lib/webview-helpers.js
mykola-mokhnach: the $ object is a [cheerio](https://github.com/cheeriojs/cheerio) parser instance. You could check https://stackoverflow.com/questions/21003384/how-to-remove-all-attributes-from-html for example on how to remove attributes from a document

wswebcreation: @mykola-mokhnach 

I used cheerio and htmlparser2 to make it safer, can you take a look again?

app/renderer/lib/webview-helpers.js
wswebcreation: Added this

test/unit/webview-helpers-specs.js
```diff
@@ -0,0 +1,26 @@
...
+import chaiAsPromised from 'chai-as-promised';
+import {parseSource} from '../../app/renderer/lib/webview-helpers';
+
+chai.use(chaiAsPromised);
+
+describe('webview-helpers.js', function () {
+  describe('#parseSource', function () {
+    it('should parse html to proper xml', function () {
+      const output = parseSource(
+        readFileSync(
```
mykola-mokhnach, line 13: can we use async reader instead?

mykola-mokhnach (approved):

test/unit/webview-helpers-specs.js
wswebcreation: Fixed it, thanks

jlipps (approved):
LGTM to me now!

wswebcreation:
Thanks for merging this, but there were 2 open points @jlipps 

did you also had the time to lok at those (see the description)

jlipps:
sorry, i went through the code and saw mykola's approval and thought we were good.

as for your open points:
1. i'm fine to not support older chromes that don't speak w3c
2. for chrome, is there a way to add another dropdown below the context dropdown which populates with windows, so that the user can also select windows in the same way they select contexts? this would make it clear what they have to do with appium as well in the automated version of their test.

jlipps:
(and the window thing could be generic, but only chrome would ever show windows in that list)

wswebcreation:
Rhanks @jlipps 

the pages thing is something for Android webviews and Chrome tabs.

In an ideal situation this should work with the default context command, but for specific situations it would need an extra drop down (I like the idea)

Should we first release this PR and fix the Android webview selection later?

jlipps:
> Should we first release this PR and fix the Android webview selection later?

Yes, I'll work on a release

jlipps:
OK, release in progress. Should be up at https://inspector.appiumpro.com momentarily

PR #225, opened 2021-12-07
fix: being blocked by webdriver client when using bitbar

lastverb:
New webdriver client blocks capabilities not compliable with W3C spec before sending anything anywhere.

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -310,8 +310,9 @@ export function newSession (caps, attachSessId = null) {
           });
           return;
         }
-        desiredCapabilities.testdroid_source = 'appiumdesktop';
-        desiredCapabilities.testdroid_apiKey = accessKey;
+        desiredCapabilities['bitbar:options'] = {};
+        desiredCapabilities['bitbar:options'].source = 'appiumdesktop';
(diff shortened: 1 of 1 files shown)


app/renderer/actions/Session.js
```diff
@@ -310,8 +310,9 @@ export function newSession (caps, attachSessId = null) {
           });
           return;
         }
-        desiredCapabilities.testdroid_source = 'appiumdesktop';
-        desiredCapabilities.testdroid_apiKey = accessKey;
+        desiredCapabilities['bitbar:options'] = {};
```
mykola-mokhnach, line 313: this could be simplified to
```js
desiredCapabilities['bitbar:options'] = {
  source: 'appiumdesktop',
  apiKey: accessKey,
}
```

lastverb: looking at most other object inits in this file, I assumed this is preffered

KazuCocoa:
Could you sign the cla?

dsaner:
Can someone who's signed the CLA make these changes?  Would be nice to be able to use this without having to jump through all the developer hoops to get a build that would work with bitbar.

jlipps:
OK, added in a separate PR

PR #270, opened 2022-01-09
fix: prevent TypeError when SAVED_SESSIONS setting is undeinfed

enkryptor:
getSetting(SAVED_SESSIONS) can return undefined:
image

Since we have no direct control over the saved data, it's reasonable to introduce a basic guard clause:
`let savedSessions = await getSetting(SAVED_SESSIONS) || []`

otherwise, `savedSessions.push()` down below throws an unprocessed exception


jlipps (approved):

jlipps:
@enkryptor thanks for the fix! looks like you need to amend the author of your commit so it matches an e-mail address on file with github, so that the CLA will work.

PR #285, opened 2022-01-30
include sauce:options by default and give job a name

dpgraham:
* If "sauce:options" is not set, set it so that it defaults to W3C
* Give the Appium Desktop job a name so that it's easier for Sauce Labs users to know which job came from Appium Desktop

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -92,6 +92,8 @@ export const DEFAULT_SERVER_PATH = '/';
 export const DEFAULT_SERVER_HOST = '127.0.0.1';
 export const DEFAULT_SERVER_PORT = 4723;
 
+const SAUCE_OPTIONS_CAP = 'sauce:options';
+
 const JSON_TYPES = ['object', 'number', 'boolean'];
 
 export function getCapsObject (caps) {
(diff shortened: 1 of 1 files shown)


app/renderer/actions/Session.js
```diff
@@ -250,6 +251,13 @@ export function newSession (caps, attachSessId = null) {
           return;
         }
         https = false;
+        if (!desiredCapabilities['sauce:options']) {
+          desiredCapabilities['sauce:options'] = {};
+        }
+        if (!desiredCapabilities['sauce:options'].name) {
+          const dateTime = moment().format('MMM DD -- h:mma');
```
mykola-mokhnach, line 258: would it be possible to format the string according to active user time format settings?

app/renderer/actions/Session.js
```diff
@@ -250,6 +251,13 @@ export function newSession (caps, attachSessId = null) {
           return;
         }
         https = false;
+        if (!desiredCapabilities['sauce:options']) {
```
mykola-mokhnach, line 254: `if (!_.isPlainObject(...))` ?

app/renderer/actions/Session.js
```diff
@@ -250,6 +251,13 @@ export function newSession (caps, attachSessId = null) {
...
+          desiredCapabilities['sauce:options'] = {};
+        }
+        if (!desiredCapabilities['sauce:options'].name) {
```
mykola-mokhnach, line 257: does it make sense to make `'sauce:options'` to a constant?

app/renderer/actions/Session.js
dpgraham: @mykola-mokhnach could you be specific, I'm not familiar with what that is.

mykola-mokhnach: check `Multiple Locale Support` at https://momentjs.com/

PR #288, opened 2022-02-02
chore: bump web2driver

KazuCocoa:


mykola-mokhnach:
shouldn't this be done by dependabot?

KazuCocoa:
Should be regularly.
I have a fix that I'd like to check in this version, so manually created this PR.

PR #290, opened 2022-02-02
fix: add connectionRetryTimeout in attaching to an existing session

KazuCocoa:
Current requests after attach to an existing session is default each end client's default timeout instead of webdriverio's default one.
As our future work, we could make them customizable, but let me make the timeout as same as webdriverio's default value for now.

For example, ky (browser) one is 10 sec. `/source` endpoint can take over the time easily.


jlipps (commented):
👍  thanks

PR #298, opened 2022-02-11
feat: call mobile command for ios page source

KazuCocoa:
at least this helps to reduce iOS's execution time. (but I know it drops `visible` attribute, so this is a kind of trade-off.)
It's probably good to add a checkbox preference.


mykola-mokhnach:
I'd also say it makes sense to make this option configurable (and disabled by default as it would break the backward compatibility)

KazuCocoa:
yea. it makes sense. let me create a checkbox now as this pr.
Then, i'm thinking to add something input text place so that users can specify excluding items (default is visible)

jlipps:
yeah I like this idea. Not sure where a checkbox would go. I don't want to clutter up the UI too much. But let's see if you have an idea.

app/renderer/actions/Inspector.js
```diff
@@ -587,13 +587,14 @@ export function callClientMethod (params) {
   return async (dispatch, getState) => {
     console.log(`Calling client method with params:`); // eslint-disable-line no-console
     console.log(params); // eslint-disable-line no-console
-    const {driver, appMode} = getState().inspector;
+    const {driver, appMode, enableMobileSauceCmd} = getState().inspector;
```
jlipps, line 590: i assume this is a typo and should be `enableMobileSourceCmd`? otherwise i don't see what this has to do with sauce labs?

KazuCocoa:
Probably... we should keep `/source` and create a new settings api's item to exclude some attribute if needed as inspector's concept. let me close this pr

KazuCocoa:
need to re-construct this idea, but I'd love to keep considering how to change the "mode" of calling `mobile:source` instead of `/source` endpoint for iOS

KazuCocoa:
-> https://github.com/appium/appium-inspector/issues/1062

app/renderer/lib/appium-client.js
```diff
@@ -309,10 +314,26 @@ export default class AppiumClient {
...
-  async getSourceUpdate () {
+  /**
+   *
+   * @param {string} appMode Current app mode
+   * @returns {object} Current page source result or error message
+   */
+  async getSourceUpdate (appMode) {
+    let pageSource;
+    const {client: {capabilities: {platformName}}} = this.driver;
+    if (this.enableMobileSauceCmd === true && ['ios', 'tvos'].includes(_.toLower(platformName)) && appMode === APP_MODE.NATIVE) {
```
KazuCocoa, line 325: basically, this switch will be 'mode' change via `sourceOptions` or something instead of the platformName

PR #323, opened 2022-02-24
chore: remove  --no-content-hash from browser build

KazuCocoa:
I observed a broken layout by browser build. In the case, the broken one has gone after clearing the browser cache or open it via a private mode.

I guess this was because content has stuff.

I tested a few environment such as behind Nginx. Then, this change could prevent the broken layout by cache.

~~It seems good to go with this. Instead, the build size raises (in my local, dist-browser size will be around 70MB from 55MB), but its better than layout broken.~~ <- this may be wrong. Possibly I haven't cleanup the directory properly


KazuCocoa:
yea.. we should find a better way

KazuCocoa:
close for now

KazuCocoa:
This change was already applied, closing

PR #342, opened 2022-03-16
README.md

leonidaa:
### Installing on macOS

If you're using the desktop app on macOS, when you run it you will likely be greeted with some error about the app not being able to be opened, or not verified by Apple, or something similar. The easiest way to get around this is to run `xattr -cr` on the file you downloaded. So let's say you downloaded `appium-inspector.dmg` and copy `Appium Inspector.app` in `/Applications` inside the disk image. Then you would run `xattr -cr "/Applications/Appium Inspector.app"` before opening it. The same goes for the zip version (or the .app itself).

The reason for this is that the Appium team does not have a paid Apple developer account and so our app isn't signed by a certificate they recognize, so the OS attempts to prevent it from running. Running the command above gets around this security limitation.

### Installing on Linux 

If you're using  Linux . 
Download the file Appium-Inspector-linux-2022.2.1.AppImage
Making it executable
You can make the appImage executable as follows:
chmod a+x Appium-Inspector-linux-2022.2.1.AppImage
Executing it
You can execute an appImage as follows:
./Appium-Inspector-linux-2022.2.1.AppImage

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Both apps have the exact same set of features, so you might find that simply ope
 
 ### Installing on macOS
 
-If you're using the desktop app on macOS, when you run it you will likely be greeted with some error about the app not being able to be opened, or not verified by Apple, or something similar. The easiest way to get around this is to run `xattr -cr` on the file you downloaded. So let's say you downloaded `appium-inspector.dmg`. Then you would run `xattr -cr appium-inspector.dmg` before opening the disk image. The same goes for the zip version (or the .app itself).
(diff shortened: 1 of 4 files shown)


KazuCocoa:
Sorry, but would you like to add `Installing on Linux` section in the `main` branch's README?

This PR seems a mistake.

leonidaa:
sorry , yes I wanted to add that readme to the main branch and it was a mistake , thank you .

KazuCocoa:
No problem.
Thank for the contribution!

PR #347, opened 2022-03-21
Feat: Add Sauce Labs Manual Testing for Real Devices

wswebcreation:
Most of the Sauce Labs Real Device users use Appium Inspector to inspect their app. This PR will add the option for those users to easily switch between App _source code_ and manual testing in Sauce Labs, see the movie below.

This PR also fixes a few small things in Appium Inspector

https://user-images.githubusercontent.com/11979740/159254509-2ea928a3-dff4-4bc8-9f71-f57a129eaec1.mp4

**Contains:**
- adds Sauce Labs manual testing video stream screen for only Sauce Labs Real Devices
- fix tap duration issue, the previous value for the duration caused the tap to be a long press
- fix setting the "Web/Hybrid App Mode" state when the session is started with a browser


app/main/sauce.js
```diff
@@ -0,0 +1,57 @@
+import WebSocket from 'ws';
+
+const SAUCE_IPC_TYPES = {
```
mykola-mokhnach, line 3: I usually enclose constant objects into Object.freeze

app/main/sauce.js
```diff
@@ -0,0 +1,57 @@
...
+  WS_STARTED: 'websocketStarted',
+};
+
+let wss = null;
+
+/**
+ * Start the websocket and process the data
+ *
+ * @param {object} event
+ * @param {object} websocketData
```
mykola-mokhnach, line 19: Please create a custom typedef for this parameter

app/main/sauce.js
```diff
@@ -0,0 +1,57 @@
...
+ * @param {string} websocketData.accessKey
+ * @param {string} websocketData.dataCenter
+ * @param {string} websocketData.sessionId
+ * @param {string} websocketData.username
+ */
+function runWebSocket (event, { accessKey, dataCenter, sessionId, username }) {
+  const wsUrl = `wss://api.${dataCenter}.saucelabs.com/v1/rdc/socket/alternativeIo/${sessionId}`;
+  wss = new WebSocket(wsUrl, {
+    headers: {
+      Authorization: `Basic ${new Buffer.from(
```
mykola-mokhnach, line 29: `new Buffer` constructor is deprecated. Use Buffer.from instead

app/main/sauce.js
```diff
@@ -0,0 +1,57 @@
...
+        `${username}:${accessKey}`
+      ).toString('base64')}`,
+    },
+  });
+  wss.onopen = () => {
```
mykola-mokhnach, line 34: do we need opening and closing brackets for oneline hanlders?

app/main/sauce.js
```diff
@@ -0,0 +1,57 @@
...
+  wss.onmessage = (e) => {
+    event.sender.send(SAUCE_IPC_TYPES.WS_MESSAGE, e.data);
+    wss.send('n/');
+  };
+}
+/**
+ * Close the websocket
+ */
+function closeWebsocket () {
+  if (wss) {
```
mykola-mokhnach, line 52: I'd say checking `wss?.readyState` would be more precise

app/renderer/actions/Inspector.js
```diff
@@ -152,30 +175,53 @@ export function unselectHoveredElement (path) {
...
+      getState().inspector.isRecording;
     try {
-      dispatch({type: METHOD_CALL_REQUESTED});
+      dispatch({ type: METHOD_CALL_REQUESTED });
       const callAction = callClientMethod(params);
-      const {contexts, contextsError, currentContext, currentContextError,
-             source, screenshot, windowSize, result, sourceError,
-             screenshotError, windowSizeError, variableName,
-             variableIndex, strategy, selector} = await callAction(dispatch, getState);
+      const {
```
mykola-mokhnach, line 185: I understand this is the work of an autoformatter, but it really makes it super hard to review this PR, because it is not clear what changes are new and what was just formatting stuff. Would it be possible to revert autoformatting changes to simplify the review process and to reduce the cognitive load?

wswebcreation: I also understand that, it's pretty hard to revert it and the problem/challenge is that we don't have real editor format rules.
I can check what is possible, but it might be reverted all the time I store the file :(

README.md
```diff
@@ -12,7 +12,7 @@ Appium Inspector is basically just an Appium client (like WebdriverIO, Appium's
 
 This version of Appium Inspector is designed to work with Appium 2.0 as a default. So if you are migrating from Appium Desktop (which is designed to work with Appium 1.x as a default), you need to be aware of some changes:
 
-* The default remote server path has changed from `/wd/hub` to `/` to reflect Appium 2.0's default server path. If you're using Appium Inspector with an Appium 1.x server, you'll likely need to update the path information in the New Session form back to `/wd/hub`.
+- The default remote server path has changed from `/wd/hub` to `/` to reflect Appium 2.0's default server path. If you're using Appium Inspector with an Appium 1.x server, you'll likely need to update the path information in the New Session form back to `/wd/hub`.
```
boneskull, line 15: I feel like this may be premature?

also... I'd hope that appium would respond with a 404 mentioning that the default base path changed. guessing it doesn't. @jlipps would that be a good idea?

app/main/sauce.js
boneskull: if we omit them, it's an implicit return, and while that may be OK in this case, I find it better to be explicit

boneskull (commented):
regarding the docstrings... they can probably stay for now, but "function" and "object" usually need specific typing. I can provide guidance on this, but not here, and it shouldn't hold up the PR

README.md
jlipps: yeah that would probably be a good idea. and it may have been premature, but we made the call to have the inspector's defaults work with appium 2's defaults, so changing it back would cause even more confusion.

jlipps (commented):
cool feature! i haven't looked at the code. we should probably have a team conversation about whether this feature is appropriate in the open source version of Appium Inspector. is the sauce manual feature following an open source control method (e.g., it's just using plain VNC) or using proprietary APIs? It seems like at least for the biometric feature or image injection feature that proprietary APIs are used.

I'm not sure if I want to build divergent vendor-specific features into ADI. But if we can agree on a standard set of websocket-based control mechanisms and other APIs, then each vendor could implement as necessary and the features would show up if the vendors implement them.

But in general I don't like the idea of building custom UI for a vendor in ADI. For that kind of thing I would expect a vendor to fork ADI and release a version with their custom UI added to it.

jlipps (commented):
I read through the code a little bit. Looks like user actions are sent via Appium, which is good. Could Sauce expose an mjpeg stream instead of a bespoke websocket stream? Then my other branch that tries to show the mjpeg stream instead of the static screenshot would work for Sauce as well.

Also, I wouldn't want to add anything to the main process (e.g. a websocket client), since that would mean this wouldn't work in the web version.

jlipps:
One other thought: if we all agree that a jpeg websocket stream is superior to an mjpeg stream, then maybe we should build support for a standard jpeg websocket stream in appium itself, that any vendor could implement.

wswebcreation:
Hi @jlipps 

Thanks for the feedback, I also understand the concerns regarding vendor specific screens, on the other hand, I didn't wanted to make a Sauce Specific fork because that would, imho: 

- confuse the users, we now already see that people are having challenges with the difference between Appium Desktop and Appium Inspector. A third app specifically for Sauce would complicate it even more
- doesn't feel appropriate to the open source community to fork it and release it ourselves

Regarding the websocket being implemented in the main process. I've chosen this approach because we need to authenticate the user to be able to allow them to see the stream. The websocket in the browser doesn't allow additional headers, the main process (node.js) does. The advantage is that no extra authentication layer needs to be build. 
I can exclude the manual testing in the browser if you can point me into a direction to distinguish web and electron.

> Could Sauce expose an mjpeg stream instead of a bespoke websocket stream? Then my other branch that tries to show the mjpeg stream instead of the static screenshot would work for Sauce as well.

We already have the websocket stream and currently don't have the mjpeg stream on our roadmap so I wanted to build on top of what we already supported.

Besides from the small adjustments in the inspector screen, it only adds new screens which should not _harm_ AI or the support that the Appium all already gives. By building this feature I realise that Sauce Labs or I need to support this specific screen and I will dedicate time for it to do so.

I hope this answers some questions/concerns you have or maybe raises some more 😉 , if needed I can jump on a call to discuss this.

jlipps:
Hi Wim. Thanks for that. Everything makes sense. However I believe that this change goes against 2 significant design goals for the inspector, so I don't believe we should evaluate and merge as is. The 2 design goals are:

1. Provide a universal interface for Appium that integrates with cloud services along the lines of Appium's officially supported features and modes of operation.
2. Build the Inspector so that it can always be run in a browser.

The second problem seems solvable technologically. For example [this SO post](https://stackoverflow.com/questions/4361173/http-headers-in-websockets-client-api) documents a solution whereby authentication doesn't happen via custom headers. 

The first problem is more conceptually significant. I don't believe that vendor-specific features or technologies should be integrated into ADI (beyond the slight differences in server connection information which we already support). This creates the possibility of too much divergence in UI design, code maintenance, and scope for the project. It would be great if somehow we had a plugin system for ADI, but I'm not sure how that would be implemented.

However, if the Appium team (and hopefully also representatives from other vendors) can agree on open APIs for the features you want to add on behalf of Sauce, then we can implement those features in ADI, so long as Sauce's APIs conform to the open APIs which we have agreed on. For example, if we decide that we want a socket-based jpeg stream reading capability for Appium in addition to the existing mjpeg stream reading capability, we should spec that out in conversation with all interested parties (and who knows, maybe Sauce's existing implementation or HeadSpin's existing implementation will be selected as the reference based on the relative merits of each). Then we can implement support for these features in ADI in a vendor-agnostic way, and each vendor will gain access to the feature so long as it has done the work to support the server side implementation of the standard. (Ideally, all these things would take place through existing Appium channels, so that additional knowledge of vendor details is not required. For example, the Appium server could optionally respond with standard websocket connection details in its New Session response).

For me, that is the appropriate path to adding more vendor-supported features. It prevents the divergence of features and UI and ensures correct maintenance incentives.

I understand it means there's not a quick path to merging the features you want to provide to your customers. In this case, I do think forking and releasing your own version of the Inspector would be the quickest path forward. The fork doesn't need to be public, if that lessens the confusion for the community; you can simply maintain a forked version and publish releases on your website, marketed as Appium Inspector for Sauce users. Then, if Sauce users want the Sauce-specific capabilities, they can use the Sauce-specific Inspector download.

Meanwhile, I'm happy to work with you and the rest of the team on the process of standardization around features that we want to be officially supported in ADI.

Happy to also hear thoughts from @KazuCocoa, @mykola-mokhnach, @boneskull, and anyone else.

boneskull:
- I don't think vendor-specific stuff belongs in Appium Inspector.
- I don't want to see a fork of Appium Inspector.  Appium Inspector should discourage forks.
- Standardization is generally a good thing.  However, vendors may want to move faster than standardization allows, and (for whatever potentially-misguided reason) may not wish to standardize.  If the project doesn't consider these possibilities, it _does_ encourage forking (which, again, is bad).
- Note: if the functionality in this PR was standardized and the PR merged, the project has taken on a sizable maintenance burden  (there's a whole discussion about Appium's approach to standardization which I think we should probably have, but maybe not in this PR).
- Also note: vendors can add extensions to Appium 2 via Plugins.  If _Appium Inspector_ is to support Appium Plugins--and it should, IMO--then Appium Inspector _itself_ needs to be extensible.
- So I see the long-term solution here to be twofold: 
    1. Project/vendor dialogue on standardization strategy, and further work on standardizing specific functionality where appropriate.
    2. Design and implement a plugin system for Appium Inspector. This PR would be implemented as a plugin, instead.

jlipps:
Appreciate those thoughts @boneskull. I agree with you that, long-term, a plugin/extension system for ADI would be a huge win. It's a very popular project and there's so much potential for it. It's unclear how much work a plugin system would be or how ready ADI is as it stands to begin that work. It's probably a project that we'd need to plan out and discuss when we have some more folks working with us who have bandwidth for this kind of thing (or at least, after Appium 2's tasks are all done).

wswebcreation:
Thanks for all the feedback. I’ll close the PR and will look for a different option to provide this to at least our Sauce Labs customers as a short term solution.

I’m still open for generic solution, so please ping me how I can help with this

PR #371, opened 2022-04-12
feature: add .appiumsession file opener

dpgraham:
(Note: this isn't a merge request into "main". This is ongoing)

* Add a file association for any file with a `.appium` extension
* Makes it so when user tries opening a `.appium` file, it will be opened by Appium Inspector
* Makes use of Redux state


package.json
```diff
@@ -56,6 +56,14 @@
     "directories": {
       "output": "release"
     },
+    "fileAssociations": [
+      {
+        "ext": "appium",
```
jlipps, line 61: I wonder if '.appium' is the best extension name. It's really more like an appium session. maybe .appiumsession? What if we, for example, end up saving other things down the road?

app/main/helpers.js
```diff
@@ -12,3 +13,24 @@ export function setSavedEnv () {
   const set = makeSetSavedEnv(settings);
   return set();
 }
+
+export function readAppiumFile (argv, isPackaged, isDev) {
+  if (process.platform === 'darwin') {
+    return {success: false};
```
jlipps, line 19: why? comments would be helpful to note any platform limitations or what's going on

app/renderer/actions/Session.js
```diff
@@ -724,6 +725,32 @@ export function setSavedServerParams () {
   };
 }
 
+export function setStateFromAppiumJson (appiumJson) {
```
jlipps, line 728: if these significantly duplicate logic or functionality from setting the state via URL, perhaps they could be merged.

app/main/helpers.js
```diff
@@ -12,3 +12,23 @@ export function setSavedEnv () {
...
+    // packaged macOS apps do not receive args from process.argv, they
+    // receive the filepath argument from the `electron.app.on('open-file', cb)` event
+    return false;
+  }
+  let argvIndex = isPackaged ? 1 : 2;
+  return argv[argvIndex];
+}
+
+// get the slice of the redux state that's needed for the .appiumsession files
+export function getSaveableState (reduxState) {
```
jlipps, line 27: why do we care about just getting the slice? is it because it's serialized and passed as a message or something?

app/renderer/actions/Session.js
```diff
@@ -724,6 +727,54 @@ export function setSavedServerParams () {
   };
 }
 
+export function setStateFromAppiumFile (newFilepath = null) {
+  return (dispatch) => {
+    // no "fs" means not a browser. do nothing
+    if (!fs) {
```
jlipps, line 733: when would this function ever be called in a non-browser context? it's in the renderer after all

app/renderer/actions/Session.js
```diff
@@ -724,6 +727,54 @@ export function setSavedServerParams () {
...
+          return;
+        }
+        filePath = lastArg.split('=')[1];
+      }
+      const filePathStorageKey = 'last_opened_file';
+      if (sessionStorage.getItem(filePathStorageKey) === filePath) {
+        // file was opened already, do nothing
+        return;
+      }
+      const appiumJson = JSON.parse(fs.readFileSync(filePath, 'utf8'));
```
jlipps, line 750: how is it possible to call `fs` from the renderer?

app/renderer/polyfills/electron.js
```diff
@@ -3,6 +3,7 @@ import { clipboard, shell, remote, ipcRenderer } from 'electron';
 import log from 'electron-log';
 import settings from 'electron-settings';
 import i18NextBackend from 'i18next-node-fs-backend';
+import fs from 'fs';
```
jlipps, line 6: weird does electron make this available to the renderer?

app/renderer/reducers/Session.js
```diff
@@ -336,6 +336,19 @@ export default function session (state = INITIAL_STATE, action) {
         ...omit(action.state, ['server']),
       };
 
+    case SET_STATE_FROM_SAVED:
+      if (!state.visibleProviders.includes(action.state.serverType) &&
+        action.state.serverType !== ServerTypes.local &&
+        action.state.serverType !== ServerTypes.remote
+      ) {
+        state.visibleProviders.push(action.state.serverType);
```
jlipps, line 344: do we want to check that the provider actually is valid?

test/fixtures/corrupted.appiumsession
```diff
@@ -0,0 +1,7 @@
+This is not valid JSON!
```
jlipps, line 1: i don't see any of these test fixtures being used; should they be committed?

jlipps:
did you pull latest gui-common after the recent merge? also looks like there are rebase conflicts

app/renderer/polyfills/electron.js
dpgraham: @jlipps it does yes! It simplified things quite a bit for me, actually. I'm glad it's there.

app/renderer/actions/Session.js
dpgraham: I worded that incorrectly. The comment should say `no "fs" means it's not an electron renderer` to indicate that it's being run from a browser.

app/main/helpers.js
dpgraham: @jlipps good question. Do you think it would make more sense to just save the entire state?

I was actually thinking a potential use-case would be to save recorded actions too.

app/renderer/actions/Session.js
dpgraham: Magic!

test/fixtures/corrupted.appiumsession
dpgraham: How about I move them over to a new folder called "sample-session-files"?

app/renderer/reducers/Session.js
dpgraham: Good thinking. Done.

app/main.js
```diff
@@ -31,7 +37,21 @@ app.on('ready', async () => {
...
+    },
+  });
+
+  ipcMain.on('save-file-as', () => {
+    dialog.showSaveDialog(mainWindow, {
+      title: 'Save Appium File',
+      filters: [
+        {name: 'Appium Session Files', extensions: ['appiumsession']},
+      ]
+    }).then(({canceled, filePath}) => {
```
mykola-mokhnach, line 50: cannot we use await instead?

app/main/helpers.js
```diff
@@ -12,3 +12,23 @@ export function setSavedEnv () {
...
   return set();
 }
+
+export function getAppiumSessionFilePath (argv, isPackaged, isDev) {
+  if (process.platform === 'darwin' && !isDev) {
+    // packaged macOS apps do not receive args from process.argv, they
+    // receive the filepath argument from the `electron.app.on('open-file', cb)` event
+    return false;
+  }
+  let argvIndex = isPackaged ? 1 : 2;
```
mykola-mokhnach, line 22: const

app/renderer/actions/Session.js
```diff
@@ -724,6 +727,54 @@ export function setSavedServerParams () {
...
+    try {
+      let filePath = newFilepath;
+      if (!newFilepath) {
+        const lastArg = process.argv[process.argv.length - 1];
+        if (!lastArg.startsWith('filename=')) {
+          return;
+        }
+        filePath = lastArg.split('=')[1];
+      }
+      const filePathStorageKey = 'last_opened_file';
```
mykola-mokhnach, line 745: this could be a global constant

app/renderer/actions/Session.js
```diff
@@ -724,6 +727,54 @@ export function setSavedServerParams () {
...
+      if (sessionStorage.getItem(filePathStorageKey) === filePath) {
+        // file was opened already, do nothing
+        return;
+      }
+      const appiumJson = JSON.parse(fs.readFileSync(filePath, 'utf8'));
+      sessionStorage.setItem(filePathStorageKey, filePath);
+      dispatch({type: SET_STATE_FROM_SAVED, state: appiumJson, filePath});
+    } catch (e) {
+      notification.error({
```
mykola-mokhnach, line 754: shall we write the actual stack trace somewhere to the log, so it could be debugged?

app/renderer/actions/Session.js
```diff
@@ -724,6 +727,54 @@ export function setSavedServerParams () {
...
+}
+
+export function saveFile (filepath) {
+  return (dispatch, getState) => {
+    const state = getState().session;
+    const filePath = filepath || state.filePath;
+    const filePathStorageKey = 'last_opened_file';
+    if (filePath) {
+      const appiumFileInfo = getSaveableState(state);
+      fs.writeFileSync(filePath, JSON.stringify(appiumFileInfo, null, 2), 'utf8');
```
mykola-mokhnach, line 768: why it must be sync?

app/renderer/actions/Session.js
```diff
@@ -724,6 +727,54 @@ export function setSavedServerParams () {
...
+      });
+    }
+  };
+}
+
+export function saveFile (filepath) {
+  return (dispatch, getState) => {
+    const state = getState().session;
+    const filePath = filepath || state.filePath;
+    const filePathStorageKey = 'last_opened_file';
```
mykola-mokhnach, line 765: can we just have a global constant?

app/renderer/reducers/Session.js
```diff
@@ -336,6 +337,25 @@ export default function session (state = INITIAL_STATE, action) {
...
 
+    case SET_STATE_FROM_SAVED:
+      if (!Object.keys(ServerTypes).includes(action.state.serverType)) {
+        notification.error({
+          message: `Failed to load session: ${action.state.serverType} is not a valid server type`,
+        });
+        return state;
+      }
+      if (!state.visibleProviders.includes(action.state.serverType) &&
+        action.state.serverType !== ServerTypes.local &&
```
mykola-mokhnach, line 348: ![ServerTypes.local, ServerTypes.remote].includes(action.state.serverType)

sample-session-files/fake.appiumsession
```diff
@@ -0,0 +1,56 @@
+{
```
mykola-mokhnach, line 1: I would rather make it versioned. It might be necessary to alter/change this format in the future, so we must know which version is supported and which is not (or probably needs to be upgraded)

app/renderer/actions/Session.js
dpgraham: Screen Shot 2022-04-28 at 1 01 42 PM

^ I'll just add it directly to the notification. It's a bit messy, but this is likely a rare case anyways because it's not going to be common for users to meddle with the file contents.

sample-session-files/fake.appiumsession
dpgraham: I updated `getSaveableState` to include version as a property.

app/renderer/actions/Session.js
dpgraham: And showing it in the notification would be helpful for users to submit bugs.

dpgraham:
Fixes made. Thanks for the feedback @jlipps and @mykola-mokhnach !

app/renderer/actions/Session.js
```diff
@@ -724,6 +729,52 @@ export function setSavedServerParams () {
...
+        if (!lastArg.startsWith('filename=')) {
+          return;
+        }
+        filePath = lastArg.split('=')[1];
+      }
+      if (sessionStorage.getItem(FILE_PATH_STORAGE_KEY) === filePath) {
+        // file was opened already, do nothing
+        return;
+      }
+      const appiumJson = JSON.parse(await util.promisify(fs.readFile)(filePath, 'utf8'));
```
KazuCocoa, line 751: What should here do when `filePath` was invalid or not existed?
Just I clone this change and started with `npm run dev`, then I got the below error.

image

The `filePath` was `./`. Is it when a user open an appium inspector without selecting any appiumsession files before?

app/main/helpers.js
```diff
@@ -3,6 +3,8 @@ import settings from '../shared/settings';
 import i18n from '../configs/i18next.config';
 import { makeOpenBrowserWindow, makeSetSavedEnv } from '../../gui-common/util';
```
KazuCocoa, line 5: Should avoid importing this module from shared places for browser/electron? (The L1's electron reference)

app/renderer/actions/Session.js
dpgraham: I'll change it so that it doesn't launch a file from the command line when it's dev. The problem is that dev is launched as `<electron> --inspect ./` and it's picking up `./`.

app/main/helpers.js
dpgraham: I don't know. Wasn't this here already though?

KazuCocoa: Maybe no files loaded this file...?
Just grepped current main branch:
```
(headspin) kazu$ git grep 'helpers'
app/renderer/lib/appium-client.js:import {getWebviewStatusAddressBarHeight, parseSource, setHtmlElementAttributes} from './webview-helpers';
package-lock.json:        "@babel/helpers": "^7.17.2",
package-lock.json:    "node_modules/@babel/helpers": {
package-lock.json:      "resolved": "https://registry.npmjs.org/@babel/helpers/-/helpers-7.17.2.tgz",
package-lock.json:        "@babel/helpers": "^7.17.2",
package-lock.json:    "@babel/helpers": {
package-lock.json:      "resolved": "https://registry.npmjs.org/@babel/helpers/-/helpers-7.17.2.tgz",
test/e2e/pre-e2e.test.js:import { e2eBefore, e2eAfter } from '../../gui-common/test-helpers';
test/unit/webview-helpers-specs.js:import {parseSource} from '../../app/renderer/lib/webview-helpers';
test/unit/webview-helpers-specs.js:describe('webview-helpers.js', function () {
```

dpgraham: I don't think I added this block of code. I think it as already there.

KazuCocoa: i mean probably like https://github.com/appium/appium-inspector/pull/371/files#diff-cc0a8b1e7efd7e5b9ee01a00c256283251315b2c0c18cca08911fef1c01fa0b9R16 is the first time to load this file. No dependencies to load this file existed before, so "already there" is correct but no references were there.

```
index.js:8 Uncaught TypeError: fs.existsSync is not a function
    at getElectronPath (index.js:8:10)
    at Object.parcelRequire.../../node_modules/electron/index.js.fs (index.js:21:18)
    at newRequire (renderer.e31bb0bc.js:47:24)
    at localRequire (renderer.e31bb0bc.js:53:14)
    at Object.parcelRequire.../main/helpers.js.electron (helpers.js:1:1)
    at newRequire (renderer.e31bb0bc.js:47:24)
    at localRequire (renderer.e31bb0bc.js:53:14)
    at Object.parcelRequire.actions/Session.js.../../shared/settings (Session.js:16:1)
    at newRequire (renderer.e31bb0bc.js:47:24)
    at localRequire (renderer.e31bb0bc.js:53:14)
    at Object.parcelRequire.containers/SessionPage.js.redux (SessionPage.js:4:1)
    at newRequire (renderer.e31bb0bc.js:47:24)
    at localRequire (renderer.e31bb0bc.js:53:14)
    at Object.parcelRequire.routes.js.react (routes.js:4:1)
    at newRequire (renderer.e31bb0bc.js:47:24)
    at localRequire (renderer.e31bb0bc.js:53:14)
```

app/main/helpers.js
jlipps: i dunno, the whole state might not make sense. i was just trying to make sure i understood what you were doing. i think we can always add more state in later versions of save files

app/renderer/components/Inspector/Inspector.js
```diff
@@ -163,6 +163,7 @@ export default class Inspector extends Component {
             </Card>
           </TabPane>
         </Tabs>
+        <div>FUDGE!</div>
```
jlipps, line 166: seems strange to commit this, at least without a comment explaining what it's doing there

sample-session-files/fake.appiumsession
```diff
@@ -0,0 +1,57 @@
...
+  "version": "v1.0",
```
jlipps, line 2: seems strange to include `v` in the version number here?

sample-session-files/sample.appiumsession
```diff
@@ -0,0 +1,51 @@
+{
```
jlipps, line 1: does this file need to be regenerated using the version aware code?

jlipps (changes requested):
ok few tiny questions/changes, then i think we'll be good to go

app/renderer/components/Inspector/Inspector.js
dpgraham: :cringe:

dpgraham: I was doing some hacking to test another feature...

My word choice could have been much worse 😌

dpgraham:
@jlipps changes made. As for "getSaveableState", I do think it's possible we'll add to it in the future. I just wanted to lean on the conservative side and only save what we needed.

dpgraham:
🤦 wth is this eslint error?

dpgraham:
Fixed the lint issue. Turns out we never actually had "eslint' installed. it only worked because it was a downstream dependency 🤷

jlipps:
ok looks good to me, seems like conflicts need to be resolved (most likely package-lock related, just blow it away and recreate it)

PR #385, opened 2022-05-04
DRAFT: feat: make the Appium Inspector openable from command line

dpgraham:
Add a feature where you can open Appium Inspector from command line:

Example:

```
node . --session-file ./sample-session-files/fake.appiumsession
```

This will open the inspector in the user's default browser and populate the state with the contents of the session file.

TODO: This will require a future pull request that publishes this as an NPM package.


main.js
```diff
@@ -0,0 +1,80 @@
+const open = require('open');
+const yargs = require('yargs/yargs');
+const { hideBin } = require('yargs/helpers')
+const { fs } = require('appium-support');
```
mykola-mokhnach, line 4: this module is not supported anymore. Consider switching to `@appium/support` one

main.js
```diff
@@ -0,0 +1,80 @@
...
+    } catch (e) {
+      throw new Error(`Session file ${sessionFile} is not valid JSON`);
+    }
+  }
+
+  if (stateJson) {
+    try {
+      state = {...state, ...JSON.parse(stateJson)};
+    } catch (e) {
+      throw new Error(`State JSON is not valid JSON`);
```
mykola-mokhnach, line 44: strings or arrays are also valid JSON

main.js
```diff
@@ -0,0 +1,80 @@
...
+module.exports.openAppiumInspector = async function openAppiumInspector ({
+  sessionFile, stateJson, autoStart = false,
+}) {
+  // start a static server to serve the dist-browser files
+  const distBrowser = path.join(__dirname, 'dist-browser');
+  const serve = serveStatic(distBrowser, {index: ['index.html']});
+  const server = http.createServer(function onRequest (req, res) {
+    serve(req, res, finalhandler)
+  });
+  const port = await portfinder.getPortPromise({port: 8888});
```
mykola-mokhnach, line 23: should port number be configurable?

jlipps:
seems like this is including a lot of irrelevant commits and should be rebased on master.

also i'm not entirely sure of the purpose of this change. can't the inspector already be started from the command line (`open Inspector.app`?)

dpgraham:
This is just a draft. The purpose of this is so that you can open the inspector from a client. Something like webdriverio could call `debug` and then that would open the running session in the inspector.

Issue #384, opened 2022-05-03
when selecting element, source is not unfolded to show selected element


jlipps:
On a first load of the inspector, the page source starts folded up. When selecting an element in the screenshot box, I would expect the page source to unfurl so that that element is shown as selected in the source as well.

This _does_ work if I've interacted with the page source already, but not if I haven't. It should work regardless of whether I have clicked anything in the source box.

PR #390, opened 2022-05-05
fix: when selecting element, source is not unfolded to show selected element #384

emaan-c:
This PR fixes the page source not folding down when selecting an element in the screenshot box on the first load of the inspector.

### DIFF
--- a/app/renderer/components/Inspector/Source.js
+++ b/app/renderer/components/Inspector/Source.js
@@ -77,6 +77,7 @@ class Source extends Component {
     const treeData = source && recursive(source);
 
     return <div id='sourceContainer' className={InspectorStyles['tree-container']}>
+      {/* Must switch to a new antd Tree component when there's changes to treeData  */}
       {treeData ?
         <Tree
           defaultExpandAll={true}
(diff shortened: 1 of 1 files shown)


app/renderer/components/Inspector/Source.js
```diff
@@ -77,14 +77,17 @@ class Source extends Component {
     const treeData = source && recursive(source);
 
     return <div id='sourceContainer' className={InspectorStyles['tree-container']}>
-      {source &&
+      {treeData ?
```
jlipps, line 80: can we add a comment here to explain what we're doing? it's strange enough we might want to leave a note for our future selves or developers, something like `// the antd Tree component doesn't handle changes in treeData well, so it's better to switch to a new component when we have treedata` (or whatever it is that you think is actually the issue, I could have worded it wrongly)

jlipps (approved):

app/renderer/components/Inspector/Source.js
emaan-c: Will do!

jlipps (approved):

PR #393, opened 2022-05-09
fix: restore scrollbars to source pane (fix #383; fix #194)

### LINKED ISSUE #194: No longer possible to scroll App Source using keyboard buttons
## Appium or Appium Desktop?

- [x] I have verified that my issue does not occur with Appium and should be investigated as an Appium Inspector issue

## The problem

In Appium Inspector (App Source) view is no longer possible to scroll by left/right/up/down/End/Home/Page Down/Page Up buttons. It's really useful when App Source is huge.

It used to be possible when Inspector was the part of Appium Desktop.

## Environment

- I am running Appium Inspector version _<v2021.10.1>_.
- I am on (pick one):
    - [x] Mac

### LINKED ISSUE #383: Appium Inspector - The App Source doesn't have a vertical scroll bar. Which Version does?
Appium Inspector 2022.2.1 for Windows

- [ ] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## The problem

App Source inside of the Appium Inspector won't allow me to scroll down to see any elements that may pass the page.

## Environment

- I am running Appium Desktop version2022.2.1
- I am on (pick one):
    - [ ] Mac
    - [x ] Windows
    - [ ] Linux
Appium Scrolling

jlipps:
seems like it's confusing not to have visible system scrollbars when necessary.
Screen Shot 2022-05-09 at 12 05 29

### DIFF
--- a/app/renderer/components/Inspector/Inspector.css
+++ b/app/renderer/components/Inspector/Inspector.css
@@ -74,11 +74,12 @@
 
 .inspector-main :global(.ant-card-head) {
     max-height: 48px;
+    padding: 0px 12px;
 }
 
 .inspector-main :global(.ant-card-body) {
     height: calc(100% - 48px);
-    padding-bottom: 48px;
+    padding: 12px;
 }
 
 .inspector-main .interaction-tab-container :global(.ant-tabs) {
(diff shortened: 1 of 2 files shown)


jlipps:
also fixed a minor css issue that left the two right panes at uneven heights

PR #396, opened 2022-05-09
Improved Appium-Inspector Search Feature

emaan-c:
In this PR, the inspector search feature has been improved in terms of, elements in the landing page are highlighted when searched. The front-end feature of the Locator Modal has been improved, specifically the back button and the no elements found page.


jlipps (changes requested):
i also notice that CI looks unhappy, maybe these changes broke an integration test or something?

jlipps (approved):

PR #411, opened 2022-05-11
feat: use monospace font in capability names and values so whitespace is obvious

### LINKED ISSUE #188: feature request: use monospace font in capability names and values so whitespace is obvious
## Appium or Appium Desktop?

## The problem

I use a lot of time to get the reason of the error:
image
finally, I find the beginning whitespace!
image

_So I advise to auto remove the whitespace from the beginning and end of the string in Capability value_

## Environment

- I am running Appium Desktop version _1.22.0_.
- I am on (pick one):
    - [ ] Mac
    - [x] Windows
    - [ ] Linux

emaan-c:
This PR Closes #188 

Changed the font in the capabilities text boxes to monospace so that whitespaces are more apparent!

image

### DIFF
--- a/app/renderer/components/Session/CapabilityEditor.js
+++ b/app/renderer/components/Session/CapabilityEditor.js
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { Button, Checkbox, Input, Modal, Form, Row, Col, Select } from 'antd';
+import { Button, Checkbox, Input, Modal, Form, Row, Col, Select, Tooltip } from 'antd';
 import FormattedCaps from './FormattedCaps';
 import CapabilityControl from './CapabilityControl';
 import SessionStyles from './Session.css';
@@ -11,6 +11,16 @@ import { ROW } from '../../../../gui-common/components/AntdTypes';
 
 const {Item: FormItem} = Form;
 const {Option} = Select;
+const whitespaces = /^\s|\s$/;
+
+function whitespaceMsg (value) {
+  const leadingSpace = /^\s/.test(value);
+  const trailingSpace = /\s$/.test(value);
+
+  if (leadingSpace && trailingSpace) {return 'Contains Leading & Trailing Whitespace';}
+  if (leadingSpace) {return 'Contains Leading Whitespace';}
+  if (trailingSpace) {return 'Contains Trailing Whitespace';}
+}
 
 export default class CapabilityEditor extends Component {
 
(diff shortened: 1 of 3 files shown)


jlipps:
the trick would be to figure out which characters to consider "invalid" or "dangerous". any ideas there?

mykola-mokhnach:
I would make it simple.
We may simply make whitespaces wisible by for example switching the font as described in https://stackoverflow.com/questions/7900655/visualize-whitespaces-using-css or via css  https://discourse.wicg.io/t/visualizing-whitespace-characters/1394

jlipps:
if there's an easy way to incorporate a font that will do this, that seems like a good idea

emaan-c:
I used Jonathan's suggestion in terms of warning the user if they are using any non-ASCII characters when the user is entering their capabilities:

image

mykola-mokhnach:
> I used Jonathan's suggestion in terms of warning the user if they are using any non-ASCII characters when the user is entering their capabilities:
> 
> image

I'm not quite sure such limitation makes sense for values, but keys

emaan-c:
> > I used Jonathan's suggestion in terms of warning the user if they are using any non-ASCII characters when the user is entering their capabilities:
> > image
> 
> I'm not quite sure such limitation makes sense for values, but keys

Shouldn't this limitation be for values as well, since appium won't be able to run any set of capabilities that have any non-ASCII characters in it?

jlipps:
> Shouldn't this limitation be for values as well, since appium won't be able to run any set of capabilities that have any non-ASCII characters in it?

I think my suggestion of non-ascii is too aggressive. and @mykola-mokhnach is right that we care less about values, since maybe this is like some user generated thing.

I kinda think we're overthinking this somewhat. Is there a simple way to just warn if there is leading or trailing whitespace? that would resolve the concrete issue that has been faced in the past. I'm not sure if other scenarios are likely to actually be problematic. what do you think @mykola-mokhnach ?

emaan-c:
> I think my suggestion of non-ascii is too aggressive. and @mykola-mokhnach is right that we care less about values, since maybe this is like some user generated thing.
> 
> I kinda think we're overthinking this somewhat. Is there a simple way to just warn if there is leading or trailing whitespace? that would resolve the concrete issue that has been faced in the past. I'm not sure if other scenarios are likely to actually be problematic. what do you think @mykola-mokhnach ?

Sounds great, I made a quick change where the tooltip only warns the user when the text box contains any leading/trailing whitespace!

image

jlipps:
I think this works pretty well! @mykola-mokhnach are you good with this?

mykola-mokhnach:
> I think this works pretty well! @mykola-mokhnach are you good with this?

Yes, it's better now. Could we improve that and be more explicit about when leading/training/both whitespaces are present?

emaan-c:
> @emaan-c can we do that, like update the tooltips to say one of the following
> 
> * `Contains Leading Whitespace`
> * `Contains Trailing Whitespace`
> * `Contains Leading and Trailing Whitespace`
> 
> based on the situation?

Yup, I have made those changes!

image

app/renderer/components/Session/CapabilityEditor.js
```diff
@@ -11,6 +11,16 @@ import { ROW } from '../../../../gui-common/components/AntdTypes';
 
 const {Item: FormItem} = Form;
 const {Option} = Select;
+const whitepaces = /^\s|\s$/;
```
jlipps, line 14: `whitespaces`

jlipps (approved):
one typo then we're good to go

PR #417, opened 2022-05-13
feat: save remote connection info alongside capabilities in saved sets

### LINKED ISSUE #398: feature request: save remote connection info alongside capabilities in saved sets
More often than not, I think we want to recall the remote connection details for a saved capability set

emaan-c:
This PR Closes #398 

In this PR whenever the user saves their capabilities, the remote connection details (this includes any cloud provider details or advanced settings) are saved as well. So the user can have many different capabilities saved, and when they click on the saved capabilities the remote session/cloud provider connection information associated with that particular CAP set is loaded as well.

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -27,7 +27,7 @@ export const SET_CAPABILITY_PARAM = 'SET_CAPABILITY_PARAM';
 export const ADD_CAPABILITY = 'ADD_CAPABILITY';
 export const REMOVE_CAPABILITY = 'REMOVE_CAPABILITY';
 export const SWITCHED_TABS = 'SWITCHED_TABS';
-export const SET_CAPS = 'SET_CAPS';
+export const SET_CAPS_AND_SERVER = 'SET_CAPS_AND_SERVER';
 export const SAVE_AS_MODAL_REQUESTED = 'SAVE_AS_MODAL_REQUESTED';
 export const HIDE_SAVE_AS_MODAL_REQUESTED = 'HIDE_SAVE_AS_MODAL_REQUESTED';
 export const SET_SAVE_AS_TEXT = 'SET_SAVE_AS_TEXT';
@@ -155,11 +155,11 @@ export function showError (e, methodName, secs = 5) {
 }
 
 /**
- * Change the caps object and then go back to the new session tab
+ * Change the caps object, along with the server details and then go back to the new session tab
  */
-export function setCaps (caps, uuid) {
+export function setCapsAndServer (server, serverType, caps, uuid) {
   return (dispatch) => {
-    dispatch({type: SET_CAPS, caps, uuid});
+    dispatch({type: SET_CAPS_AND_SERVER, server, serverType, caps, uuid});
   };
 }
 
@@ -200,8 +200,9 @@ export function removeCapability (index) {
 }
 
 function _addVendorPrefixes (caps, dispatch, getState) {
+  const {server, serverType, capsUUID} = getState().session;
   const prefixedCaps = addVendorPrefixes(caps);
-  setCaps(prefixedCaps, getState().session.capsUUID)(dispatch);
+  setCapsAndServer(server, serverType, prefixedCaps, capsUUID)(dispatch);
   return prefixedCaps;
 }
 
@@ -556,9 +557,9 @@ export function newSession (caps, attachSessId = null) {
 
 
 /**
- * Saves the caps
+ * Saves the caps and server details
  */
-export function saveSession (caps, params) {
+export function saveSession (server, serverType, caps, params) {
   return async (dispatch) => {
     let {name, uuid} = params;
     dispatch({type: SAVE_SESSION_REQUESTED});
@@ -572,6 +573,8 @@ export function saveSession (caps, params) {
         name,
         uuid,
         caps,
+        server,
+        serverType,
       };
       savedSessions.push(newSavedSession);
     } else {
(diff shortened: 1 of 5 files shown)


app/renderer/actions/Session.js
```diff
@@ -201,7 +201,10 @@ export function removeCapability (index) {
 
 function _addVendorPrefixes (caps, dispatch, getState) {
   const prefixedCaps = addVendorPrefixes(caps);
```
jlipps, line 203: you can just do `const {server, serverType, capsUUID} = getState().session` to get a short reference to each of these and avoid so many chars.

app/renderer/components/Session/SavedSessions.js
```diff
@@ -19,9 +19,9 @@ export default class SavedSessions extends Component {
   onRow (record) {
     return {
       onClick: () => {
-        const {setCaps} = this.props;
+        const {setCapsWithServer} = this.props;
         let session = this.sessionFromUUID(record.key);
-        setCaps(session.caps, session.uuid);
+        setCapsWithServer(session.server, session.serverType, session.caps, session.uuid);
```
jlipps, line 24: we need to assume that someone who has an existing version of the app won't have any server or serverType defined, so let's make sure there's handling for that case somewhere (the default action should be to just do nothing in that case; keep the current server info as is; then when the user saves it again, the server info will get written)

app/renderer/actions/Session.js
```diff
@@ -27,7 +27,7 @@ export const SET_CAPABILITY_PARAM = 'SET_CAPABILITY_PARAM';
 export const ADD_CAPABILITY = 'ADD_CAPABILITY';
 export const REMOVE_CAPABILITY = 'REMOVE_CAPABILITY';
 export const SWITCHED_TABS = 'SWITCHED_TABS';
-export const SET_CAPS = 'SET_CAPS';
+export const SET_CAPS_SERVER = 'SET_CAPS_SERVER';
```
jlipps, line 30: can we change this to `SET_CAPS_AND_SERVER`

app/renderer/actions/Session.js
```diff
@@ -155,11 +155,11 @@ export function showError (e, methodName, secs = 5) {
 }
 
 /**
- * Change the caps object and then go back to the new session tab
+ * Change the caps object, along with the server details and then go back to the new session tab
  */
-export function setCaps (caps, uuid) {
+export function setCapsWithServer (server, serverType, caps, uuid) {
```
jlipps, line 160: maybe here also `setCapsAndServer` to keep it consistent

jlipps (changes requested):

app/renderer/actions/Session.js
```diff
@@ -580,13 +585,15 @@ export function saveSession (caps, params) {
       for (let session of savedSessions) {
         if (session.uuid === uuid) {
           session.caps = caps;
+          session.server = server;
+          session.serverType = serverType;
```
mykola-mokhnach, line 589: would that still work for older saves?

app/renderer/components/Session/SavedSessions.js
mykola-mokhnach: I was already mentioning this once that our save files must have versioned schemas. Not sure if it was ever implemented though

emaan-c: Thanks for the help! I just added an if statement in SavedSessions.js file to check if capability set has a server and serverType defined before calling the setCapsAndServer function

mykola-mokhnach: I assume we are going to have more and more `if`'s when new fields are added to saved files or existing ones are changed. A proper way of doing that stuff would be to have versioning and schemas/migrations.

jlipps: I'm not sure how much we want to future-proof this. A couple if statements are easier to read/understand than a schema migration system. I'd probably vote for keeping as-is for now and then revisiting if we find ourselves making lots of changes to this format (this is the first time in like 5 years we have done so).

app/renderer/components/Session/SavedSessions.js
```diff
@@ -31,6 +30,18 @@ export default class SavedSessions extends Component {
     return capsUUID === record.key ? SessionCSS.selected : '';
   }
 
+  handleCapsAndServer (session) {
+    const {setCapsAndServer, server, serverType } = this.props;
+
+    // Incase user has CAPS saved from older version of Inspector which
+    // doesn't contain server and serverType within the session object
+    if (('server' in session) && ('serverType' in session)) {
```
jlipps, line 38: more often we use the idiom `if (session.server)`; if it's undefined, null, or an empty string it will evaluate to false, which is often what we want.

app/renderer/components/Session/SavedSessions.js
```diff
@@ -31,6 +30,18 @@ export default class SavedSessions extends Component {
...
+      setCapsAndServer(session.server, session.serverType, session.caps, session.uuid);
```
jlipps, line 39: this could all be one line:

```js
setCapsAndServer(session.server || server, session.serverType || serverType, session.caps, session.uuid);
```

app/renderer/actions/Session.js
jlipps: just adding new fields on the object, so it should work. i'll pull it down and check on my current saves to be sure before merging.

jlipps (approved):

app/renderer/components/Session/SavedSessions.js
jlipps: @mykola-mokhnach are you OK with this approach for now?

PR #418, opened 2022-05-13
feat: show mjpeg stream instead of static screenshot

jlipps:
if the mjpegScreenshotUrl cap is included, we can use that
for a more real time screen view, and forego getting screenshot

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -73,6 +73,8 @@ const FILE_PATH_STORAGE_KEY = 'last_opened_file';
 
 const AUTO_START_URL_PARAM = '1'; // what should be passed in to ?autoStart= to turn it on
 
+const MJPEG_CAP = 'mjpegScreenshotUrl';
+
 // Multiple requests sometimes send a new session request
 // after establishing a session.
 // This situation could happen easier on cloud vendors,
@@ -538,17 +540,28 @@ export function newSession (caps, attachSessId = null) {
       } catch (ign) {}
     }
 
+
+    const mjpegScreenshotUrl = desiredCapabilities[`appium:${MJPEG_CAP}`] ||
+      desiredCapabilities[MJPEG_CAP] ||
+      null;
+
+
     // pass some state to the inspector that it needs to build recorder
     // code boilerplate
-    const action = setSessionDetails(driver, {
-      desiredCapabilities,
-      host,
-      port,
-      path,
-      username,
-      accessKey,
-      https,
-    }, mode);
+    const action = setSessionDetails({
+      driver,
+      sessionDetails: {
+        desiredCapabilities,
+        host,
+        port,
+        path,
+        username,
+        accessKey,
+        https,
+      },
+      mode,
+      mjpegScreenshotUrl,
+    });
     action(dispatch);
     dispatch(push('/inspector'));
   };
--- a/app/renderer/components/Inspector/HighlighterRects.js
+++ b/app/renderer/components/Inspector/HighlighterRects.js
@@ -1,5 +1,4 @@
 import React, { Component } from 'react';
-import { debounce } from 'lodash';
 import HighlighterRect from './HighlighterRect';
 import B from 'bluebird';
 import { SCREENSHOT_INTERACTION_MODE } from './shared';
@@ -10,27 +9,6 @@ const {TAP, SWIPE, SELECT} = SCREENSHOT_INTERACTION_MODE;
  * Shows screenshot of running application and divs that highlight the elements' bounding boxes
  */
 export default class HighlighterRects extends Component {
-
-  constructor (props) {
-    super(props);
-    this.state = {
-      scaleRatio: 1,
-    };
-    this.updateScaleRatio = debounce(this.updateScaleRatio.bind(this), 1000);
-  }
-
-  /**
-   * Calculates the ratio that the image is being scaled by
-   */
-  updateScaleRatio () {
-    const screenshotEl = this.props.containerEl.querySelector('img');
-
-    // now update scale ratio
-    this.setState({
-      scaleRatio: (this.props.windowSize.width / screenshotEl.offsetWidth)
-    });
-
-  }
 
   async handleScreenshotClick () {
     const {screenshotInteractionMode, applyClientMethod,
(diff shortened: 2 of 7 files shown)


app/renderer/components/Inspector/Screenshot.js
```diff
@@ -166,4 +165,4 @@ class Screenshot extends Component {
   }
 }
 
-export default withTranslation(Screenshot);
```
mykola-mokhnach, line 169: is the wrapper not necessary anymore?

app/renderer/actions/Session.js
```diff
@@ -538,17 +540,28 @@ export function newSession (caps, attachSessId = null) {
       } catch (ign) {}
     }
 
+
+    const mjpegScreenshotUrl = desiredCapabilities[`appium:${MJPEG_CAP}`] ||
```
mykola-mokhnach, line 544: this feels a bit hacky to me that we need to fetch the value from capabilities. I would rather move this functionality to a separate inspector setting, so the mjpeg server url could be provided separately. Also, do we want to fallback to vanilla screenshots if the url is not accessible?

jlipps: > this feels a bit hacky to me that we need to fetch the value from capabilities

i thought about this, but it seemed straightforward to just use the existing capability feature. if we create a separate control, then the user would have to put it in both places (or we'd have to auto-use the capability under the hood, which seems dangerous). but i am open to it if you think this way will really confuse users.

> Also, do we want to fallback to vanilla screenshots if the url is not accessible?

i don't think so. if the user provides a cap with incorrect details, better that they see that something is broken?

another (more difficult) option would be to bundle an mjpeg client, which would give us finer grained control and be able to tell if we're actually getting any image data, rather than relying on chrome to show a broken image.

but i figured this was a good start.

mykola-mokhnach: what will be shown in case we don't get anything from the mjpeg server?

KazuCocoa:
Seems good overall. Locally tested a bit. let me do a bit more later

app/renderer/components/Inspector/Screenshot.js
jlipps: I didn't see it doing anything, and it was preventing me from accessing the component from another component, so I removed it.

app/renderer/actions/Session.js
jlipps: hmm yeah, right now it's just a broken image. not sure if there's a way in javascript to do some kind of detection to see whether there is actually data there. maybe an API to convert to base64 and see if it's empty? if so we could show a placeholder or a spinner.

app/renderer/components/Inspector/Screenshot.js
mykola-mokhnach: It was needed for localization to work. Does it still look ok if you change the app language?

jlipps: oh, I didn't try. I just saw that the component didn't seem to have any text in it so I figured it was an extraneous i18n wrapper. but i don't want to break it if it is doing something

jlipps: i was able to refactor this PR so that i could replace the `withTranslation` wrapper.

app/renderer/actions/Session.js
jlipps: ok I now have an interval-based checking function that will detect if the image stream stops working and will show the spinner until it starts working again (i was able to see this in action and it works pretty well)

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -11,27 +11,6 @@ const {TAP, SWIPE, SELECT} = SCREENSHOT_INTERACTION_MODE;
  */
 export default class HighlighterRects extends Component {
 
-  constructor (props) {
-    super(props);
-    this.state = {
-      scaleRatio: 1,
-    };
-    this.updateScaleRatio = debounce(this.updateScaleRatio.bind(this), 1000);
```
mykola-mokhnach, line 19: do we still need to import `debounce`?

app/renderer/components/Inspector/Inspector.js
```diff
@@ -57,10 +59,27 @@ export default class Inspector extends Component {
...
+
+  /**
+   * Calculates the ratio that the image is being scaled by
+   */
+  updateScaleRatio () {
+    const screenshotImg = this.screenshotEl.querySelector('img');
+
+    // now update scale ratio
+    this.setState({
+      scaleRatio: (this.props.windowSize.width / screenshotImg.offsetWidth)
```
mykola-mokhnach, line 81: can screenshotImg.offsetWidth be equal to zero?

app/renderer/components/Inspector/Inspector.js
```diff
@@ -108,6 +129,28 @@ export default class Inspector extends Component {
...
+    try {
+      await img.decode();
+      imgReady = true;
+    } catch (ign) {}
+    if (imgReady && isAwaitingMjpegStream) {
+      setAwaitingMjpegStream(false);
+      this.updateSourceTreeWidth();
+    } else if (!imgReady && !isAwaitingMjpegStream) {
+      setAwaitingMjpegStream(true);
+    }
```
mykola-mokhnach, line 153: what if none of the branches above matches?

app/renderer/components/Inspector/Screenshot.js
```diff
@@ -18,23 +18,9 @@ class Screenshot extends Component {
     super(props);
     this.containerEl = null;
     this.state = {
-      scaleRatio: 1,
       x: null,
       y: null,
     };
-    this.updateScaleRatio = debounce(this.updateScaleRatio.bind(this), 1000);
```
mykola-mokhnach, line 25: shall we still import `debounce`?

app/renderer/components/Inspector/Inspector.js
jlipps: then the state will not change. `isAwaitingMjpegStream` starts out as true, so a spinner will be shown from the start.

this code handles the situations where:
1. we move from no image data to having image data
2. we move from having image data to no image data

I think that's all the cases we need to handle. I'm basically trying to eliminate spamming the redux state change since this is being called frequently.

app/renderer/components/Inspector/HighlighterRects.js
jlipps: oops, that caused lint to fail. fixed.

jlipps:
@mykola-mokhnach @KazuCocoa based on my testing this is ready to go. any objections? i'm also happy to release as is and then see how people use it, and if we want to make the streaming url more of a first class feature we can add a UI control specifically for it later on.

mykola-mokhnach (approved):
lets try it

KazuCocoa (approved):
👍

PR #442, opened 2022-05-27
fix: add connection retry timeout

dpgraham:
The connection retry timeout we have currently is too short for some cloud providers.

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -512,9 +512,6 @@ export function newSession (caps, attachSessId = null) {
         // assume the device is mobile so that Appium protocols are included
         // in the userPrototype.
         serverOpts.isMobile = true;
-        // Need to set connectionRetryTimeout as same as the new session request.
-        // TODO: make configurable?
-        serverOpts.connectionRetryTimeout = 5 * 60 * 1000;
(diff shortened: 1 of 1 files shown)


wswebcreation:
@dpgraham 

I think you should now remove https://github.com/appium/appium-inspector/blob/0b30b91e76dd1bb5d22f6cfd96c4310d152e0b63/app/renderer/actions/Session.js#L517
This has no value

KazuCocoa (approved):
lgtm

app/renderer/actions/Session.js
```diff
@@ -481,6 +481,7 @@ export function newSession (caps, attachSessId = null) {
       protocol: https ? 'https' : 'http',
       path,
       connectionRetryCount: CONN_RETRIES,
+      connectionRetryTimeout: 10 * 60 * 1000,
```
jlipps, line 484: let's put magic constants up top with a descriptive variable name.

and we're sure we want to make it a 10 minute connection timeout in all cases, not just the cloud provider case? not sure this makes sense for local servers.

jlipps (changes requested):

jlipps:
closing due to no response and conflicts

PR #443, opened 2022-05-27
fix: session panel info and padding

emaan-c:
This PR fixes some bugs and frontend features of the session information panel. 

This includes: 

1.) Updating Session URL to display the correct URL: http://<host>:<port><path>/session/<session-id>
2.) Fixing some padding issues 
3.) Setting the session timer to update live instead of updating when user interacts with the UI
4.) Combining the server details into one row
5.) Removed the session CAPS row with session details, which now retrieves all session info from the webdriver.io command 'getSession'

However, I could not determine a way to display the scroll bar on default, still looking into antd docs to determine what properties need to change in order to accomplish that.

image

### DIFF
--- a/app/renderer/components/Inspector/SessionInfo.js
+++ b/app/renderer/components/Inspector/SessionInfo.js
@@ -1,44 +1,67 @@
 import React, { Component } from 'react';
-import { Table, Card } from 'antd';
+import { Table, Row, Col } from 'antd';
 import SessionCodeBox from './SessionCodeBox';
 import { withTranslation } from '../../util';
+import InspectorStyles from './Inspector.css';
+import formatJSON from 'format-json';
 
 const SESSION_OBJ = {
   session_id: 'Session ID', session_url: 'Session URL',
   server_details: 'Server Details', session_length: 'Session Length',
-  session_caps: 'Session Capabilities', server_response: 'Server Details from Server',
-  active_appId: 'Currently Active App ID'
+  session_details: 'Session Details', active_appId: 'Currently Active App ID'
 };
 
-const SCROLL_DISTANCE_Y = 105;
-const SCROLL_DISTANCE_X = 100;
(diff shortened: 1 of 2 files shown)


app/renderer/components/Inspector/SessionInfo.js
```diff
@@ -38,7 +64,9 @@ class SessionInfo extends Component {
     const secDiff = (minDiff - minutes) * 60;
     const seconds = Math.floor(secDiff);
 
-    return `${minutes}:${(seconds < 10) ? '0' : ''}${seconds}`;
+    const checkTime = (time) => `${(time < 10) ? '0' : ''}${time}`;
```
mykola-mokhnach, line 67: can we use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart instead?

app/renderer/components/Inspector/SessionInfo.js
```diff
@@ -99,20 +147,18 @@ class SessionInfo extends Component {
...
-      case 'Session URL': return intoHTML(sessionUrl);
-      case 'Server Details': return this.getTable(serverDetailsArray, 'serverDetails', isOuterTable);
-      case 'Session Length': return intoHTML(this.generateSessionTime());
-      case 'Session Capabilities': return this.getTable(capsArray, 'caps', isOuterTable);
-      case 'Server Details from Server': return this.getTable(serverStatusArray, 'serverStatus', isOuterTable);
-      case 'Currently Active App ID': return intoHTML(appId);
-      default: return intoHTML('Invalid Header');
+      case 'Session ID': return sessionId;
+      case 'Session URL': return sessionUrl;
+      case 'Server Details': return this.getTable([...serverDetailsArray, ...serverStatusArray], SERVER_TABLE_KEY, isOuterTable);
```
mykola-mokhnach, line 157: this line seems to be too long

PR #452, opened 2022-06-03
feat: improve element selection

### LINKED ISSUE #276: feature request: mechanism for handling overlapping element regions in the screenshot element selection
## Appium or Appium Desktop?

- [X] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## The problem
The is the problem with visualizing the highlighter areas at least for iOS (because view hierarchies tend to co-exist and overlap on iOS)

Here is video of the problem
screencast 2022-01-16 22-16-05

I first interact with the Reminders app and then open the details of a particular reminder  (opens in a modal on top) — and can't interact with it, because highlighter rects remain those of the screen underneath the modal. 

After playing a bit with electron debugger I found that there are actually all the highlighter rects for modal, but because all these highlighter rects are plain and ordered by zIndex - it happens that zIndex of those highlighter rects underneath the modal are greater than those of the modal elements. It happens because of how zIndex is calculated here https://github.com/appium/appium-inspector/blob/main/app/renderer/components/Inspector/HighlighterRects.js#L128
<pre>
-> Contrainer1 (zIndex:0)
  -> subContainer (zIndex:1)
    -> subSubContainer (zIndex: 2)
      -> element1 (zIndex: 3)
      -> element2 (zIndex: 3)
  -> subContainer (zIndex: 1)
    -> element3 (zIndex: 2)
</pre>

with hierarchy like this, if element1 or element2 happens to overlap with element3 — they will stay on top. And whats worth — if subSubContainer (zIndex: 2) happens to be fullscreen — it will make element3 completely inaccessible. That's what happens on the video example. 

I have a fix for this problem, which changes how zIndex is being calculated. The idea is that zIndex gets carried over when stepping across sub-hierarchies. My example hierarchy above would become: 
<pre>
-> Contrainer1 (zIndex:0)
  -> subContainer (zIndex:1)
    -> subSubContainer (zIndex: 2)
      -> element1 (zIndex: 3)
      -> element2 (zIndex: 3)
  -> subContainer (zIndex: 4)
    -> element3 (zIndex: 5)
</pre>
This way we can ensure that elements that appear "on top" in the hierarchy are guaranteed to be "on top" in the zIndex meaning.  
What do you think of this idea? I have it implemented and can push a PR with the fix. 

## Environment

- I am running Appium Desktop version 2022.1.2
- I am on (pick one):
    - [X] Mac
    - [ ] Windows
    - [ ] Linux

emaan-c:
This PR closes #276 

This PR seeks to improve the element selection within Appium Inspector by rendering all the elements as small centroids so users can see and highlight certain difficult elements. Square boxes represent non container elements, circle centroids represent containers, dashed green centroids/boxes represent elements not in the current hierarchy/overlapping elements, and the +/- circles are just expandable circles to show those overlapping elements.

image

### DIFF
--- a/app/renderer/components/Inspector/HighlighterCentroid.js
+++ b/app/renderer/components/Inspector/HighlighterCentroid.js
@@ -1,0 +1,127 @@
+import React, { Component } from 'react';
+import InspectorCSS from './Inspector.css';
+import { RENDER_CENTROID_AS } from './shared';
+
+const {CENTROID, OVERLAP, EXPAND} = RENDER_CENTROID_AS;
+const CENTROID_STYLES = {
+  VISIBLE: 'visible',
+  HIDDEN: 'hidden',
+  CONTAINER: '50%',
+  NON_CONTAINER: '0%',
+};
+
+// Generate new coordinates along a circlular trajectory
+// for overlapping elements only
+function getCentroidPos (type, angle, coord) {
+  return type === OVERLAP ?
+    `calc((${angle} * 2.6vh) + ${coord}px)`
+    :
+    coord;
+}
+
+/**
+ * Shows all element centroids
+ */
+export default class HighlighterCentroid extends Component {
+
+  onMouseEnter (path) {
+    const {selectHoveredElement, selectHoveredCentroid,
+           centroidType} = this.props;
+    if (centroidType === EXPAND) {
+      selectHoveredCentroid(path);
+    } else {
+      selectHoveredElement(path);
+    }
+  }
+
+  onMouseLeave () {
+    const {unselectHoveredElement, unselectHoveredCentroid,
+           centroidType} = this.props;
+    if (centroidType === EXPAND) {
+      unselectHoveredCentroid();
+    } else {
+      unselectHoveredElement();
+    }
+  }
+
+  onClickCentroid (path) {
+    const {selectElement, unselectElement,
+           selectCentroid, unselectCentroid, centroidType,
+           selectedCentroid, selectedElementPath} = this.props;
+    if (centroidType === EXPAND) {
+      if (path === selectedCentroid) {
+        unselectCentroid();
+      } else {
+        selectCentroid(path);
+      }
+    } else {
+      if (path === selectedElementPath) {
+        unselectElement();
+      } else {
+        selectElement(path);
+      }
+    }
+  }
+
+  render () {
+    const {selectedElementPath, hoveredElement = {}, element, elementProperties,
+           centroidType, hoveredCentroid, selectedCentroid} = this.props;
+    const {path: hoveredPath} = hoveredElement;
+    const {centerX, centerY, angleX, angleY,
+           keyCode, path, container} = elementProperties;
+    const centroidClasses = [InspectorCSS['centroid-box']];
+    centroidClasses.push(InspectorCSS[centroidType]);
+
+    // Highlght centroids that represent elements
+    if (centroidType !== EXPAND) {
+      if (hoveredPath === path) {
+        centroidClasses.push(InspectorCSS['hovered-element-box']);
+      }
+      if (selectedElementPath === path) {
+        centroidClasses.push(InspectorCSS['inspected-element-box']);
+      }
+    }
+
+    // Highlight +/- centroids
+    if (centroidType !== CENTROID) {
+      if (hoveredCentroid === keyCode) {
+        centroidClasses.push(InspectorCSS['hovered-element-box']);
+      }
+      if (selectedCentroid === keyCode && !element) {
+        centroidClasses.push(InspectorCSS['inspected-element-box']);
+      }
+    }
+
+    const overlapDivStyle = {
+      visibility:
+        keyCode === selectedCentroid ?
+          CENTROID_STYLES.VISIBLE : CENTROID_STYLES.HIDDEN,
+    };
+
+    const centroidDivStyle = {
+      left: getCentroidPos(centroidType, angleX, centerX),
+      top: getCentroidPos(centroidType, angleY, centerY),
+      borderRadius:
+        element && !container ?
+          CENTROID_STYLES.NON_CONTAINER : CENTROID_STYLES.CONTAINER,
+      ...(centroidType === OVERLAP ? overlapDivStyle : {})
+    };
+
+    const placeHolder = centroidType === EXPAND ?
+      <div className={InspectorCSS['plus-minus']}>
+        {keyCode === selectedCentroid ? '-' : '+'}
+      </div>
+      :
+      <div></div>;
+
+    return <div
+      className={centroidClasses.join(' ').trim()}
+      onMouseOver={() => this.onMouseEnter(path)}
+      onMouseOut={() => this.onMouseLeave()}
+      onClick={() => this.onClickCentroid(path)}
+      key={path}
+      style={centroidDivStyle}>
+      {placeHolder}
+    </div>;
+  }
+}
--- a/app/renderer/components/Inspector/HighlighterRect.js
+++ b/app/renderer/components/Inspector/HighlighterRect.js
@@ -1,6 +1,5 @@
 import React, { Component } from 'react';
 import InspectorCSS from './Inspector.css';
-import { parseCoordinates } from './shared';
 
 /**
  * Absolute positioned divs that overlay the app screenshot and highlight the bounding
@@ -10,7 +9,7 @@ export default class HighlighterRect extends Component {
 
   render () {
     const {selectedElement = {}, selectHoveredElement, unselectHoveredElement, hoveredElement = {}, selectElement, unselectElement, element,
-           zIndex, scaleRatio, xOffset, elLocation, elSize} = this.props;
+           scaleRatio, xOffset, elLocation, elSize, dimensions} = this.props;
     const {path: hoveredPath} = hoveredElement;
     const {path: selectedPath} = selectedElement;
 
@@ -18,12 +17,7 @@ export default class HighlighterRect extends Component {
     highlighterClasses = [InspectorCSS['highlighter-box']];
 
     if (element) {
-      // Calculate left, top, width and height coordinates
-      const {x1, y1, x2, y2} = parseCoordinates(element);
-      left = x1 / scaleRatio + xOffset;
-      top = y1 / scaleRatio;
-      width = (x2 - x1) / scaleRatio;
-      height = (y2 - y1) / scaleRatio;
+      ({width, height, left, top} = dimensions);
 
       // Add class + special classes to hovered and selected elements
       if (hoveredPath === element.path) {
@@ -47,7 +41,7 @@ export default class HighlighterRect extends Component {
       onMouseOut={unselectHoveredElement}
       onClick={() => key === selectedPath ? unselectElement() : selectElement(key)}
       key={key}
-      style={{zIndex, left: (left || 0), top: (top || 0), width: (width || 0), height: (height || 0)}}>
+      style={{left: (left || 0), top: (top || 0), width: (width || 0), height: (height || 0)}}>
       <div></div>
     </div>;
   }
--- a/app/renderer/components/Inspector/HighlighterRects.js
+++ b/app/renderer/components/Inspector/HighlighterRects.js
@@ -1,20 +1,159 @@
 import React, { Component } from 'react';
 import HighlighterRect from './HighlighterRect';
+import HighlighterCentroid from './HighlighterCentroid';
 import B from 'bluebird';
-import { SCREENSHOT_INTERACTION_MODE } from './shared';
+import { SCREENSHOT_INTERACTION_MODE, parseCoordinates, RENDER_CENTROID_AS } from './shared';
 
 const {TAP, SWIPE, SELECT} = SCREENSHOT_INTERACTION_MODE;
+const {CENTROID, OVERLAP, EXPAND} = RENDER_CENTROID_AS;
 
 /**
  * Shows screenshot of running application and divs that highlight the elements' bounding boxes
  */
 export default class HighlighterRects extends Component {
 
+  getElements (source) {
+    const elementsByOverlap = this.buildElementsWithProps(source, null, [], {});
+    let elements = [];
+
+    // Adjust overlapping elements
+    for (const key of Object.keys(elementsByOverlap)) {
+      if (elementsByOverlap[key].length > 1) {
+        const {centerX, centerY} = elementsByOverlap[key][0].properties;
+
+        // Create new element obj which will be a +/- centroid
+
+        const element = {
+          type: EXPAND,
+          element: null,
+          parent: null,
+          properties: {
+            left: null,
+            top: null,
+            width: null,
+            height: null,
+            centerX,
+            centerY,
+            angleX: null,
+            angleY: null,
+            path: key,
+            keyCode: key,
+            container: null,
+            accessible: null
+          }
+        };
+        elements = [...elements, element,
+          ...this.updateOverlapsAngles(elementsByOverlap[key], key)];
+      } else {
+        elements.push(elementsByOverlap[key][0]);
+      }
+    }
+
+    return elements;
+  }
+
+  // This func creates a new object for each element and determines its properties
+  // 'elements' is an array that stores all prev elements
+  // 'overlaps' is an object which organzies elements by their positions
+  buildElementsWithProps (source, prevElement, elements, overlaps) {
+    if (!source) {
+      return {};
+    }
+    const {scaleRatio} = this.props;
+    const {x1, y1, x2, y2} = parseCoordinates(source);
+    const xOffset = this.highlighterXOffset || 0;
+    const centerPoint = (v1, v2) =>
+      Math.round(v1 + ((v2 - v1) / 2)) / scaleRatio;
+    const obj = {
+      type: CENTROID,
+      element: source,
+      parent: prevElement,
+      properties: {
+        left: x1 / scaleRatio + xOffset,
+        top: y1 / scaleRatio,
+        width: (x2 - x1) / scaleRatio,
+        height: (y2 - y1) / scaleRatio,
+        centerX: centerPoint(x1, x2) + xOffset,
+        centerY: centerPoint(y1, y2),
+        angleX: null,
+        angleY: null,
+        path: source.path,
+        keyCode: null,
+        container: false,
+        accessible: source.attributes.accessible
+      }
+    };
+    const coordinates = `${obj.properties.centerX}.${obj.properties.centerY}`;
+    obj.properties.container = this.isElementContainer(obj, elements);
+
+    elements.push(obj);
+
+    if (source.path) {
+      if (overlaps[coordinates]) {
+        overlaps[coordinates].push(obj);
+      } else {
+        overlaps[coordinates] = [obj];
+      }
+    }
+
+    if (source.children) {
+      for (const childEl of source.children) {
+        this.buildElementsWithProps(childEl, source, elements, overlaps);
+      }
+    }
+
+    return overlaps;
+  }
+
+  isElementContainer (element1, elements) {
+    for (const element2 of elements) {
+      if (element2.element !== element1.element
+        && this.isElementOverElement(element1.properties, element2.properties)
+        && !this.isAncestor(element1.parent, element2.element, elements)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  isElementOverElement (element1, element2) {
+    return element1.left <= element2.left
+        && element1.width >= element2.width
+        && element1.top >= element2.top
+        && element1.height >= element2.height;
+  }
+
+  // Traverse through parent elements until we reach maybeAncestor
+  isAncestor (curElement, maybeAncestor, elements) {
+    if (elements.length > 0) {
+      while (curElement !== null) {
+        if (curElement === maybeAncestor) { return true; }
+
+        for (const elem of elements) {
+          if (elem.element === curElement) { curElement = elem.parent; }
+        }
+      }
+    }
+    return false;
+  }
+
+  // Generate angles for circular positioning for overlaping elements
+  updateOverlapsAngles (elements, key) {
+    const steps = elements.length;
+    for (let step = 0; step < steps; step++) {
+      const [el, elProps] = [elements[step], elements[step].properties];
+      el.type = OVERLAP;
+      elProps.keyCode = key;
+      elProps.angleX = Math.cos(2 * Math.PI * (step / steps));
+      elProps.angleY = Math.sin(2 * Math.PI * (step / steps));
+    }
+    return elements;
+  }
+
   async handleScreenshotClick () {
     const {screenshotInteractionMode, applyClientMethod,
            swipeStart, swipeEnd, setSwipeStart, setSwipeEnd} = this.props;
     const {x, y} = this.state;
-
     if (screenshotInteractionMode === TAP) {
       applyClientMethod({
         methodName: TAP,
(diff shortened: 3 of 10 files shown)


app/renderer/components/Inspector/HighlighterRect.js
```diff
@@ -1,38 +1,55 @@
...
   render () {
-    const {selectedElement = {}, selectHoveredElement, unselectHoveredElement, hoveredElement = {}, selectElement, unselectElement, element,
-           zIndex, scaleRatio, xOffset, elLocation, elSize} = this.props;
+    const {selectedElement = {}, selectHoveredElement, unselectHoveredElement, hoveredElement = {}, selectElement, unselectElement, element, elementProperties,
+           scaleRatio, xOffset, elLocation, elSize, renderElementBoxes } = this.props;
     const {path: hoveredPath} = hoveredElement;
     const {path: selectedPath} = selectedElement;
 
-    let width, height, left, top, highlighterClasses, key;
+    let highlighterClasses, centroidClasses, key, divStyle;
```
mykola-mokhnach, line 17: consider defining variables close to code blocks where they are used

app/renderer/components/Inspector/HighlighterRect.js
```diff
@@ -1,38 +1,55 @@
...
+    let {left, top, width, height, centerX, centerY} = elementProperties.dimensions;
```
mykola-mokhnach, line 18: can the variables be constants?

app/renderer/components/Inspector/HighlighterRect.js
```diff
@@ -1,38 +1,55 @@
...
       if (selectedPath === element.path) {
-        highlighterClasses.push(InspectorCSS['inspected-element-box']);
+        renderElementBoxes ?
+          highlighterClasses.push(InspectorCSS['inspected-element-box'])
+          :
+          centroidClasses.push(InspectorCSS['inspected-element-box']);
       }
-      key = element.path;
+
+      key = renderElementBoxes ? element.path : `C:${element.path}`;
```
mykola-mokhnach, line 40: what is this magic prefix?

app/renderer/components/Inspector/HighlighterRect.js
```diff
@@ -1,38 +1,55 @@
...
+
+      divStyle = !renderElementBoxes ?
+        {
+          left: (centerX - (10 / scaleRatio + (xOffset || 0)) || 0),
```
mykola-mokhnach, line 44: what is the meaning of these magic numbers?

app/renderer/components/Inspector/HighlighterRect.js
```diff
@@ -1,38 +1,55 @@
...
+          top: (centerY - (10 / scaleRatio) || 0),
+          width: (20 / scaleRatio),
+          height: (20 / scaleRatio),
+          ...(elementProperties.container && { borderRadius: '50%' }),
```
mykola-mokhnach, line 48: not sure if spread operator would work here if elementProperties.container is falsy

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,130 @@
...
 export default class HighlighterRects extends Component {
 
+  organizeElements (source) {
+    ELEMENTS_ARRAY = [];
+    OVERLAPS_ARRAY = [];
+    BESTOVERLAPS = [];
+
+    this.copyElementsIntoArray(source, null);
+
+    for (let element of ELEMENTS_ARRAY) {
```
mykola-mokhnach, line 28: use const where possible

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,130 @@
...
+      if (this.isMouseOverElement(element.element.attributes)) {
+        OVERLAPS_ARRAY.push(element);
```
mykola-mokhnach, line 30: this could be changed to filter->map

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,130 @@
...
+          && !this.isAncestor(element1.parent, element2.element)) {
+          element1.container = true;
+          break;
+        }
+      }
+    }
+
+    for (let element of OVERLAPS_ARRAY) {
+      const { container, dimensions: { centerX, centerY } } = element;
+      if ((container && distance(centerY, centerX) <= 12) || !container) {
```
mykola-mokhnach, line 47: what is the `12` magic number?

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,130 @@
...
+        BESTOVERLAPS.push(element);
```
mykola-mokhnach, line 48: this could be replaced with map->filter

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,130 @@
...
+
+  isElementOverElement (element1, element2) {
+    return element1.left <= element2.left
+        && element2.width <= element1.width
+        && element1.top >= element2.top
+        && element2.height <= element1.height;
+  }
+
+  isAncestor (parent, checkElement) {
+    while (parent !== null) {
```
mykola-mokhnach, line 88: could it be this will be an endless loop?

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,130 @@
...
+    const {scaleRatio} = this.props;
+    const {x1, y1, x2, y2} = parseCoordinates(element);
+    const left = (x1 / scaleRatio + (this.highlighterXOffset || 0)) || 0;
+    const top = (y1 / scaleRatio) || 0;
+    const width = ((x2 - x1) / scaleRatio) || 0;
+    const height = ((y2 - y1) / scaleRatio) || 0;
+    const centerX = ((x1 + ((x2 - x1) / 2)) / scaleRatio + (this.highlighterXOffset || 0)) || 0;
+    const centerY = ((y1 + ((y2 - y1) / 2)) / scaleRatio) || 0;
+
+    ELEMENTS_ARRAY.push({
```
mykola-mokhnach, line 110: I would rather prefer more explicit variant of this function, where it returns elements array, rather than mutates it implicitly:

```
ELEMENTS_ARRAY.push(...fetchElements());
```

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -67,33 +175,44 @@ export default class HighlighterRects extends Component {
   render () {
     const {source, screenshotInteractionMode, containerEl, searchedForElementBounds,
            isLocatorTestModalVisible, scaleRatio} = this.props;
+    const {x, y} = this.state || 0;
```
mykola-mokhnach, line 178: `0` cannot be destructed

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -67,33 +175,44 @@ export default class HighlighterRects extends Component {
...
+    MOUSE_X = x;
```
mykola-mokhnach, line 179: it is possible to destruct and assign in the same expression: 
```
const o = {p: 42, q: true};
const {p: foo, q: bar} = o;

console.log(foo); // 42
console.log(bar); // true
```

app/renderer/components/Inspector/HighlighterCentroid.js
```diff
@@ -0,0 +1,96 @@
...
+      left: centerX,
+      top: centerY,
+      width: '2.2vh',
+      height: '2.2vh',
+      borderStyle: 'solid',
+      borderColor: '#ff4141',
+      borderWidth: '2px',
+      borderRadius: !element || container ? '50%' : '0%',
+      ...(currentRender === RENDER.OVERLAP ?
+        {
```
jlipps, line 50: this extra options object could be a constant at the top of the file

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,152 @@
...
+  getElements (source) {
+    const elementsByOverlap = this.findElementProps(source, null, [], {});
+    let elements = [];
+
+    for (let key in elementsByOverlap) {
+      let element = elementsByOverlap[key][0];
+
+      //Create new element obj which will be a +/- centroid
+      if (elementsByOverlap[key].length > 1) {
+        element = {
```
jlipps, line 24: is there a reason we're mutating `element` rather than creating a new object? we typically never want to mutate unless there's a reason

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,152 @@
...
+
+  // Create a new object for each element and determine its properties
+  findElementProps (element, prevElement, elements, overlaps) {
+    if (!element) {
+      return {};
+    }
+    const { scaleRatio, containerEl } = this.props;
+    const {x1, y1, x2, y2} = parseCoordinates(element);
+    const xOffset = this.highlighterXOffset || 0;
+    const posOffset = 0.03 * containerEl.offsetWidth;
```
jlipps, line 60: what is this magic constant doing? it should have a descriptive name and comment up top

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,152 @@
...
+    const obj = {
+      type: RENDER.CENTROID,
+      element,
+      parent: prevElement,
+      properties: {
+        left: x1 / scaleRatio + xOffset || 0,
+        top: y1 / scaleRatio || 0,
+        width: (x2 - x1) / scaleRatio || 0,
+        height: (y2 - y1) / scaleRatio || 0,
+        centerX: ((Math.round(x1 + ((x2 - x1) / 2)) / scaleRatio) + xOffset - posOffset) || 0,
```
jlipps, line 70: if you create intermediate variables for left, top, width, height, you might be able to define centerpoints more easily, without as much repetition of math

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,152 @@
...
+    return element1.left <= element2.left
+        && element2.width <= element1.width
+        && element1.top >= element2.top
+        && element2.height <= element1.height;
+  }
+
+  isAncestor (parent, checkElement, elements) {
+    if (elements.length > 0) {
+      while (parent !== null) {
+        if (parent === checkElement) { return true; }
```
jlipps, line 117: so elements are considered ancestors of themselves?

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,152 @@
...
+  }
+
+  isElementOverElement (element1, element2) {
+    return element1.left <= element2.left
+        && element2.width <= element1.width
+        && element1.top >= element2.top
+        && element2.height <= element1.height;
+  }
+
+  isAncestor (parent, checkElement, elements) {
```
jlipps, line 114: it's a bit unclear what we're checking is an ancestor of what? is this function signature basically saying, "tell me if `parent` is an ancestor of `checkElement`"? or the other way around?

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,152 @@
...
+        for (let elem of elements) {
+          if (elem.element === parent) { parent = elem.parent; }
+        }
+      }
+    }
+    return false;
+  }
+
+  // Generate new coordinates for overlapping elements
+  newCoordinates (elements, key) {
```
jlipps, line 128: `newCoordinates` is not very descriptive. maybe something like `updateElementsCoords`? to make it clear that we are mutating the elements array

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,152 @@
...
+    const { containerEl } = this.props;
+    const radius = 0.06 * containerEl.offsetWidth;
```
jlipps, line 130: what is this magic number? let's describe/comment and move it up top (slash get rid of it if possible! where is this coming from?)

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,152 @@
...
+    const steps = elements.length;
+    for (let step = 0; step < steps; step++) {
```
jlipps, line 132: const; also if we are just doing a simple loop like this it's ok to use a counter variable like `i`

but more broadly, why not just do `for (const el of elements)`? typically we prefer that kind of iteration over for loops with counters.

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -66,38 +198,46 @@ export default class HighlighterRects extends Component {
...
+
+    let renderCentroids = (centroids) => {
+      for (let elem of centroids) {
+        highlighterCentroids.push(<HighlighterCentroid {...this.props}
+          currentRender={elem.type}
+          elementProperties={elem.properties}
+          element={elem.element}
+          scaleRatio={scaleRatio}
+          key={`C:${elem.properties.path}`}
+          RENDER={RENDER}
```
jlipps, line 235: why are we passing in a constant which could just be imported?

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -66,38 +198,46 @@ export default class HighlighterRects extends Component {
...
+    };
+
+    let renderCentroids = (centroids) => {
+      for (let elem of centroids) {
+        highlighterCentroids.push(<HighlighterCentroid {...this.props}
+          currentRender={elem.type}
+          elementProperties={elem.properties}
+          element={elem.element}
+          scaleRatio={scaleRatio}
+          key={`C:${elem.properties.path}`}
```
jlipps, line 234: it looks like `key` is not retrieved in this component and is always overwritten, so do we need to include it here?

jlipps (changes requested):
Looking really good, really close! Bunch of small code tweaks/requests and a few questions here.

app/renderer/reducers/Inspector.js
```diff
@@ -33,6 +34,7 @@ const INITIAL_STATE = {
   actionFramework: DEFAULT_FRAMEWORK,
   sessionDetails: {},
   isLocatorTestModalVisible: false,
+  centroidMode: false,
```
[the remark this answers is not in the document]
emaan-c: centroid mode is just supposed represent the state of the show/hide element handles switch. If the switch is turned on, the centroids are displayed on the screen. I changed the name to showCentroids to hopefully make it more clear!

app/renderer/components/Inspector/HighlighterCentroid.js
```diff
@@ -0,0 +1,96 @@
...
+    if (currentRender !== RENDER.CENTROID) {
+      if (hoveredCentroid === keyCode) {
+        centroidClasses.push(InspectorCSS['hovered-element-box']);
+      }
+      if (selectedCentroid === keyCode && !element) {
+        centroidClasses.push(InspectorCSS['inspected-element-box']);
+      }
+    }
+
+    const divStyle = {
```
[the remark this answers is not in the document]
emaan-c: I removed the base options and added it into the css itself. However, I can't add this object to the top of the file just because I have to access certain properties like `keyCode`, `selectedCentroid`, `element`, `centerX`, `centerY`, and etc.

app/renderer/components/Inspector/HighlighterRects.js
emaan-c: Yup, added those changes in the new commit. When I was trying to fix the feature for the android version, the scale ratio was off by a lot or it wasn't accurate. So I used the screenshot width as the scaleRatio and made the radius distance = 6% of the screenshot width

app/renderer/components/Inspector/HighlighterRects.js
emaan-c: The reason why I used a loop with counters is just because I needed to use those counters when determining the position of each element overlap. I couldn't find any other way to accomplish that without using for loop counters

app/renderer/components/Inspector/HighlighterRects.js
emaan-c: Not necessarily, we are checking to see if an arbitrary element in the source (which is not equal to the current element) is an ancestor of the current element

app/renderer/components/Inspector/HighlighterCentroid.js
```diff
@@ -0,0 +1,86 @@
...
+      if (hoveredCentroid === keyCode) {
+        centroidClasses.push(InspectorCSS['hovered-element-box']);
+      }
+      if (selectedCentroid === keyCode && !element) {
+        centroidClasses.push(InspectorCSS['inspected-element-box']);
+      }
+    }
+
+    const overlapDivStyle = {
+      borderStyle: 'dashed',
```
mykola-mokhnach, line 41: shall we hardcode styles?

app/renderer/components/Inspector/HighlighterCentroid.js
```diff
@@ -0,0 +1,86 @@
...
+      borderColor: '#178b00',
+      visibility: keyCode === selectedCentroid ? 'visible' : 'hidden'
+    };
+
+    const centroidDivStyle = {
+      left: centerX,
+      top: centerY,
+      borderRadius: element && !container && '0%',
+      ...(currentRender === OVERLAP && overlapDivStyle)
```
mykola-mokhnach, line 50: I'm not quote sure this is going to work as expected if `currentRender === OVERLAP` results to false.
how about `...(currentRender === OVERLAP ? overlapDivStyle : {})`?

app/renderer/components/Inspector/HighlighterCentroid.js
```diff
@@ -0,0 +1,86 @@
...
+    const overlapDivStyle = {
+      borderStyle: 'dashed',
+      borderColor: '#178b00',
+      visibility: keyCode === selectedCentroid ? 'visible' : 'hidden'
+    };
+
+    const centroidDivStyle = {
+      left: centerX,
+      top: centerY,
+      borderRadius: element && !container && '0%',
```
mykola-mokhnach, line 49: it's not quite obvious to what this expression could be calculated

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,165 @@
...
+            path: key,
+            keyCode: key,
+            container: null,
+            accessible: null
+          }
+        };
+        elements = elements.concat(element,
+          this.updateElementsCoords(elementsByOverlap[key], key));
+      } else {
+        elements.push(elementsByOverlap[key][0]);
```
mykola-mokhnach, line 50: is the first element guaranteed to exist?

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -111,9 +251,12 @@ export default class HighlighterRects extends Component {
 
     // Don't show highlighter rects when Search Elements modal is open
     if (!isLocatorTestModalVisible) {
-      recursive(source);
+      renderElements(elements);
+      if (centroidMode) {
+        renderCentroids(elements);
```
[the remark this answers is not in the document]
jlipps: I guess I just want to make sure we're not drawing the element rectangles twice: once in renderElements and once in renderCentroids. That's not the case, is it?

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,165 @@
...
+import HighlighterCentroid from './HighlighterCentroid';
 import B from 'bluebird';
-import { SCREENSHOT_INTERACTION_MODE } from './shared';
+import { SCREENSHOT_INTERACTION_MODE, parseCoordinates, RENDER_CENTROID_AS } from './shared';
 
 const {TAP, SWIPE, SELECT} = SCREENSHOT_INTERACTION_MODE;
+const {CENTROID, OVERLAP, EXPAND} = RENDER_CENTROID_AS;
+
+// Centroid positions are shifted based off screenshot width % for perfect alignment
+const POS_OFFSET = 0.03;
```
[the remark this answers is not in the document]
emaan-c: Yup, thats a great point and I made changes to now ensure that we are not using any magic numbers and instead we are just relying on CSS in terms of making sure the distances between the overlapping centroids and the positioning of the centroids are more accurate, consistent and I don't think it will produce any bugs in the future as well now!

app/renderer/components/Inspector/Inspector.js
```diff
@@ -186,14 +188,47 @@ export default class Inspector extends Component {
...
-           mjpegScreenshotUrl, isAwaitingMjpegStream} = this.props;
+           mjpegScreenshotUrl, isAwaitingMjpegStream, toggleCentroidMode, showCentroids} = this.props;
     const {path} = selectedElement;
 
     const showScreenshot = ((screenshot && !screenshotError) ||
                             (mjpegScreenshotUrl && !isAwaitingMjpegStream));
 
+    let screenShotControls = <div className={InspectorStyles['screenshot-controls']}>
+      <div className={InspectorStyles['action-controls']}>
+        <Tooltip title={t(showCentroids ? 'Hide Element Handles' : 'Show Element Handles')} placement="topRight">
```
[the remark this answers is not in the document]
emaan-c: Hey @mykola-mokhnach , thanks for the feedback, much appreciated! I was just wondering if I would also need to send a PR to appium desktop to accomplish this?

image

app/renderer/components/Inspector/HighlighterRects.js
emaan-c: Yup its guaranteed to exist because all the elements in the xml file are sorted based off of their positions, and so any element that has a unique position is added to an array and assigned a new object so an object cannot have an empty array.

app/renderer/components/Inspector/HighlighterRects.js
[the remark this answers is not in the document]
emaan-c: Nope, we have a list that contains all elements and their properties and so `renderElements` traverses that list and draws the element rectangles only and `renderCentroids` also traverses that same list and draws the element centroids and the +/- centroids as well

app/renderer/components/Inspector/HighlighterCentroid.js
```diff
@@ -0,0 +1,126 @@
...
+const CENTROID_STYLES = {
+  VISIBLE: 'visible',
+  HIDDEN: 'hidden',
+  CONTAINER: '50%',
+  NON_CONTAINER: '0%',
+};
+
+// Generate new coordinates along a circlular trajectory
+// for overlapping elements only
+const CENTROID_POS = (type, angle, coord) =>
```
jlipps, line 15: this can just be a function, not a constant anonymous function, e.g.,

```js
function getCentroidPos (type, angle, coord) {
  // ...
}
```

app/renderer/components/Inspector/HighlighterRects.js
```diff
@@ -1,20 +1,159 @@
...
+    if (source.path) {
+      if (overlaps[coordinates]) {
+        overlaps[coordinates].push(obj);
+      } else {
+        overlaps[coordinates] = [obj];
+      }
+    }
+
+    if (source.children) {
+      for (let childEl of source.children) {
```
jlipps, line 100: `const`

app/renderer/components/Inspector/Inspector.js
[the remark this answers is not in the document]
jlipps: unfortunately, yes, they share the same localization project. so yeah add them to the english json in that project, we'll merge it, then you can run the script as mentioned in the instructions you found so that the changes become part of this PR

jlipps (changes requested):
a few tiny tiny things, plus the appium desktop PR for localization which you are right about, then we can merge this!

jlipps:
OK, the other commit has been merged

emaan-c:
>

emaan-c:
> OK, the other commit has been merged

Sounds perfect @jlipps ! All the changes have just been committed

jlipps (approved):

PR #505, opened 2022-07-05
rename robotic.mobi to robotqa

donmezburak:
Robotic.mobi has changed name as RobotQA.


jlipps (approved):
There is some risk that by changing the internal string names used for preference keys that a user who had the 'roboticmobi' panel selected might open the new version of the app and be greeted with an error. Did you try actually building this and running it in that scenario to test?

donmezburak:
Hello, thank you for pointing this issue.

We have already informed our clients for renaming. So that, they will use the new one, probably delete the old. Thanks.

When will your team publish the new version with this pr ?

jlipps:
I don't think informing them is enough. We might need some error handling if we don't want the app to be unusable. In the meantime I'm not sure whether it will be a problem. Is there a way for you to test it?

jlipps:
closing due to conflicts and no response

PR #514, opened 2022-07-12
feat: Gesture editor for more complex gestures

### LINKED ISSUE #400: feature request: gesture editor for more complex gestures
It would be nice to find a way to allow for the flexibility of the Actions API within the inspector. I'm sure there are a lot of challenges in making an intuitive UI for this but it seems valuable.

emaan-c:
This PR closes #400 

This is the gesture editor feature which allows users to execute complex gestures!

image


jlipps:
👍  looking forward to seeing it develop!

app/renderer/components/Inspector/GestureEditor.js
```diff
@@ -0,0 +1,565 @@
...
+                      {y !== undefined && <p>Y: {y}{coordType === PIXELS ? 'px' : '%'}</p>}
+                    </div>
+                  }>
+                  {type === POINTER_MOVE && <RightCircleOutlined key={tick.id} className={InspectorCSS['gesture-header-icon']} style={iconStyle}/>}
+                  {type === POINTER_DOWN && <DownCircleOutlined key={tick.id} className={InspectorCSS['gesture-header-icon']} style={iconStyle}/>}
+                  {type === POINTER_UP && <UpCircleOutlined key={tick.id} className={InspectorCSS['gesture-header-icon']} style={iconStyle}/>}
+                  {type === PAUSE && <PauseCircleOutlined key={tick.id} className={InspectorCSS['gesture-header-icon']} style={iconStyle}/>}
+                </Popover>}/>;
+            } else {
+              return <Step key='step2' status={STATUS.WAIT}
```
jlipps, line 504: what does `step2` mean?

app/renderer/components/Inspector/SavedGestures.js
```diff
@@ -0,0 +1,173 @@
...
+    const {applyClientMethod} = this.props;
+    const pointers = this.convertCoordinates(gesture.actions);
+    const actions = this.formatGesture(pointers);
+    applyClientMethod({methodName: SCREENSHOT_INTERACTION_MODE.GESTURE, args: [actions]});
+  }
+
+  formatGesture (pointers) {
+    const actions = {};
+    for (const pointer of pointers) {
+      actions[pointer.name] = (pointer.ticks).map((tick) => {
```
jlipps, line 83:
```js
actions[pointer.name] = pointer.ticks.map((tick) => _.omit(tick, 'id'))
```

app/renderer/components/Inspector/SavedGestures.js
```diff
@@ -0,0 +1,173 @@
...
+        delete tick.id;
+        return tick;
+      });
+    }
+    return actions;
+  }
+
+  convertCoordinates (pointers) {
+    const newPointers = JSON.parse(JSON.stringify(pointers));
```
jlipps, line 92: cloneDeep

app/renderer/components/Inspector/SavedGestures.js
```diff
@@ -0,0 +1,173 @@
...
+    newPointers.map((pointer) => {
+      (pointer.ticks).map((tick) => {
+        if (tick.type === POINTER_TYPES.POINTER_MOVE) {
+          tick.x = this.percentageToPixels(tick.x, true);
+          tick.y = this.percentageToPixels(tick.y, false);
+        }
+        return tick;
```
jlipps, line 99: these return values aren't doing anything. these should be for loops.

app/renderer/components/Inspector/SavedGestures.js
```diff
@@ -0,0 +1,173 @@
...
+      });
+      return pointer;
+    });
+    return newPointers;
+  }
+
+  percentageToPixels (percentage, isX) {
```
jlipps, line 106: are we duplicating this function somewhere else? should probably be a utility method if so, that we can pass the width/height into

app/renderer/components/Inspector/GestureEditor.js
```diff
@@ -0,0 +1,565 @@
...
+/**
+ * Shows the gesture editor interface
+ */
+class GestureEditor extends Component {
+
+  constructor (props) {
+    super(props);
+    this.interval = null;
+    this.state = {
+      name: this.props.loadedGesture ? this.props.loadedGesture.name : 'Untitled',
```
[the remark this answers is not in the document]
emaan-c: We can't change it to that because its not guaranteed that there will be a loaded gesture because when we create a new gesture, `loadedGesture` is null and `loadedGesture.name` would throw an error

app/renderer/components/Inspector/GestureEditor.js
```diff
@@ -0,0 +1,565 @@
...
+        return tick;
+      });
+    }
+    return actions;
+  }
+
+  // This converts all the coordinates in the gesture to px/%
+  convertCoordinates (type) {
+    const {coordType, pointers} = this.state;
+    const newPointers = JSON.parse(JSON.stringify(pointers));
```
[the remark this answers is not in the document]
emaan-c: I tried using `_.cloneDeep` but it still doesnt give a deep copy of the datastructure which causes bugs when saving/saving as

app/renderer/components/Inspector/GestureEditor.js
```diff
@@ -0,0 +1,565 @@
...
+    this.setState({pointers});
+  }
+
+  // Updates the current tick within local state
+  updateTick (tick, msg, value) {
+    const {POINTER_TYPE} = TICK_PROPS;
+    const {selectTick} = this.props;
+    const {pointers} = this.state;
+    const currentPointer = pointers.find((pointer) => pointer.id === tick.id[0]);
+    currentPointer.ticks = (currentPointer.ticks).map((prevTick) => {
```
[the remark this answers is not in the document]
emaan-c: The reason why we can't mutate the object as it is because React doesnt allow you to mutate state objects. So if were mutating a complex object like this gesture within the local state, then we have to send a whole new copy of the previous object + any changes to the local set state so it can be saved within the state. Its really weird! However, I can apply your change above but it leads to a whole bunch of `if` statements, I thought this way would be a lot more cleaner and it doesnt require us to to type check on a bunch of values, and also allows us to restructure each tick if we are changing a pointer type. Else we would have to use a delete operator and write a lot of unnecessary code, for some reason some of the lodash functions dont operate properly when were set state and etc.

app/renderer/components/Inspector/GestureEditor.js
emaan-c: `step2` are just place holder values within the timeline to keep it the same relative lengths as the other pointer timelines. I made the name more descriptive

app/renderer/components/Inspector/GestureEditor.js
```diff
@@ -0,0 +1,565 @@
...
+      id: this.props.loadedGesture ? this.props.loadedGesture.id : null,
+      pointers: this.props.loadedGesture ? this.props.loadedGesture.actions : DEFAULT_POINTERS(),
+      coordType: COORD_TYPE.PERCENTAGES,
+      activeKey: '1',
+    };
+  }
+
+  // Draw gesture every 1000 ms
+  componentDidMount () {
+    this.interval = setInterval(() => {
```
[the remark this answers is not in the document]
emaan-c: I understand im in the process of changing this. the change isnt in the new PR atm

app/renderer/components/Inspector/GestureEditor.js
```diff
@@ -153,137 +159,142 @@ class GestureEditor extends Component {
...
-
-  percentageToPixels (percentage, isX) {
-    const {width, height} = this.props.windowSize;
-    if (!isNaN(percentage)) {
-      if (isX) {
-        return Math.round(width * (percentage / 100));
-      } else {
-        return Math.round(height * (percentage / 100));
+    const len = prevPointerMoves.length;
+    if (len !== 0) {
```
jlipps, line 195: if you reverse this conditional it will be easier to read and save a level of indent.

```js
if (len === 0) {
  return 0;
}

// other stuff here
```

app/renderer/components/Inspector/GestureEditor.js
```diff
@@ -153,137 +159,142 @@ class GestureEditor extends Component {
...
+      const obj = {x1: prevPointerMoves[len - 1].x, y1: prevPointerMoves[len - 1].y, x2, y2};
+      if (coordType === PERCENTAGES) {
+        obj.x1 = percentageToPixels(obj.x1, true, width, height);
+        obj.y1 = percentageToPixels(obj.y1, false, width, height);
+        obj.x2 = percentageToPixels(obj.x2, true, width, height);
+        obj.y2 = percentageToPixels(obj.y2, false, width, height);
       }
+      const eq = (val1, val2) => Math.sqrt((val1 ^ 2) + (val2 ^ 2));
```
jlipps, line 203: what is `eq`? seems like a name like `distance` or `length` might be easier to understand?

app/renderer/components/Inspector/GestureEditor.js
```diff
@@ -153,137 +159,142 @@ class GestureEditor extends Component {
...
+      const xDiff = obj.x2 - obj.x1;
```
jlipps, line 204: should these be absolute values? otherwise you're adding potentially negative numbers in the distance equation which could lead to strange outcomes.

app/renderer/components/Inspector/GestureEditor.js
```diff
@@ -153,137 +159,142 @@ class GestureEditor extends Component {
...
-        currentTick.x = this.pixelsToPercentage(x, true);
-        currentTick.y = this.pixelsToPercentage(y, false);
+        currentTick.x = pixelsToPercentage(x, true, width, height);
+        currentTick.y = pixelsToPercentage(y, false, width, height);
       } else {
         currentTick.x = x;
         currentTick.y = y;
       }
+
+      if (!currentTick.duration) {
```
jlipps, line 232: this will probably update the duration if the user has entered "0" manually, right? ideally we would not update the duration if the user has interacted with the field at all.

app/renderer/components/Inspector/GestureEditor.js
```diff
@@ -153,137 +159,142 @@ class GestureEditor extends Component {
...
+      color: COLORS[key - 1]
+    });
+    this.setState({activeKey: newKey, pointers});
+  }
+
+  deletePointer (targetKey) {
     const {unselectTick} = this.props;
     const {pointers} = this.state;
+    // newKey variable keeps track of the previous pointer before deleting the current one
+    let newKey = '1';
```
jlipps, line 256: a comment explaining why it's defaulting to `1` would be nice!

app/renderer/components/Inspector/shared.js
```diff
@@ -1,6 +1,26 @@
...
+  if (!isNaN(pixel)) {
+    if (isX) {
+      return parseFloat(((pixel / width) * 100).toFixed(1), 10);
+    } else {
+      return parseFloat(((pixel / height) * 100).toFixed(1), 10);
+    }
+  }
+}
+
+export function percentageToPixels (percentage, isX, width, height) {
```
jlipps, line 14: given that we are now passing in width and height, i suggest the function would be more simply implemented as follows:

```js
function percentageToPixels (pct, maxPixels) {
  if (!isNaN(pct)) {
    return Math.round(maxPixels * (pct / 100));
  }
}
```

and then you just pass in `width` or `height` based on whether the pct is an x or y value. Same comment goes for `pixelsToPercentage`. also, are the percentages being passed in as 0-100? or 0-1? typically percentages would be represented in code as 0-1, in which case you don't need to do the `/ 100` step here

app/renderer/components/Inspector/GestureEditor.js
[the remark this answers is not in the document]
jlipps: in that case, `this.props.loadedGesture?.name || 'Untitled'`

app/renderer/components/Inspector/GestureEditor.js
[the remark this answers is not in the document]
jlipps: woah, that's weird. there's probably something we can investigate more there, but i guess let's just keep the reparsing solution for now.

app/renderer/components/Inspector/GestureEditor.js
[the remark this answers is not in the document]
jlipps: ok, I get it about using setState. just mutating the object doesn't trigger a state update.

i'm not getting your point about how it would look more complicated using my suggestion, however. it would actually involve fewer if statements than you currently have, the way i see it? (it would add one if statement that you don't currently have, in the case where we don't find a tick that matches the requested tick; but this is a case which isn't  handled by your current code and might be a potential bug)

app/renderer/components/Inspector/shared.js
emaan-c: The numbers being passed into the function are from 0-100, so we do need to `/ 100`! I decided it to do it this way just so its easier to display the values instead of having to multiply by 100 each time if we want to display the percentages

app/renderer/components/Inspector/GestureEditor.js
emaan-c: There wouldn't be negative values since the positions are always positive, but your definitely right in terms of causing potential problems in the future if there happen to be negative values some how. I added `abs()` to ensure we don't run into that issue

jlipps (approved):

Issue #525, opened 2022-07-17
Initial screen layout seems to be broken with V2022.7.1


wswebcreation:
## Appium or Appium Desktop?
- [x] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## The problem
When you load the latest AI the layout is broken, see below
<img width="1392" alt="image" src="https://user-images.githubusercontent.com/11979740/179417852-d9cf9927-2751-4f71-9ba8-50b729eee379.png">
The bottom bar is gone, the top bar doesn't seem to be fixed at the top.

This worked with the previous version

## Environment

- I am running Appium Desktop version 2022.7.1.
- I am on (pick one):
    - [x] Mac
    - [ ] Windows
    - [ ] Linux

wswebcreation:
I'm not sure what changed but when I add `overflow: auto;` to the [`.scrollingTabCont`](https://github.com/appium/appium-inspector/blob/main/app/renderer/components/Session/Session.css#L82)-class it works

emaan-c:
Hey @wswebcreation, thanks for the issue report. Feel free to open a pull request to push these changes!

wswebcreation:
Hi @emaan-c 

Will do, but wondering if you also see this issue. It seems that this started happening after the antd upgrade in the deps. I couldn't find any other change you, or someone else made recently that would have impacted this.

I can create a PR for it, will do that in a few min

wswebcreation:
see 👆

emaan-c:
Im not sure exactly what may have caused this issue, it could have been the antd upgrade in the deps. I know some other users along with myself have experienced this UI bug, I just hadn’t got around to fixing it yet. However, I did pull your PR down and it did seem to fix the issue!

anogo:
The same problem with App Version 2023.3.1

jlipps:
can you share some screenshots @anogo ? and maybe @eglitise might have some ideas?

eglitise:
I did encounter something similar while working on #752, but not to the extent of this issue, and I fixed it as part of the same PR. Otherwise I haven't observed this since #528 was merged in version 2022.8.1.

PR #528, opened 2022-07-19
Fix issue 525

wswebcreation:
This fixes https://github.com/appium/appium-inspector/issues/525

This might be pretty important for people who downloaded the latest version. They are stuck now if the list is bigger than 10 stored sessions

### DIFF
--- a/app/renderer/components/Session/Session.css
+++ b/app/renderer/components/Session/Session.css
@@ -86,6 +86,7 @@
   flex-grow: 1;
   display: flex;
   flex-direction: column;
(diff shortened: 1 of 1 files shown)


jlipps (approved):
looks good, do you happen to have a screenshot for quick review?

wswebcreation:
This is before, here you don't see the footer and the session container can't be scrolled

image

This is with the fix, simple but effective 😄 

image

KazuCocoa (approved):

Issue #560, opened 2022-08-15
scripts in projects.json do not work on Windows environment


Mehdi-HAFID:
## Appium or Appium Desktop?

Appium 

## The problem

The current version has an import in the file app/renderer/actions/Session.js of
`import ky from 'ky/umd';`
it fails for me. I have to change it to:` import ky from 'ky';`

Also the script `watch:browser` does not work on Windows. I had to add cross-env to the start:
`"watch:browser": "cross-env BUILD_BROWSER=1 NODE_ENV=development parcel serve --target browser ./app/renderer/index.html",`

## Environment
- I am on:
    - [x] Windows

KazuCocoa:
https://github.com/appium/appium-inspector/pull/562 may help. You can apply the change on your local to check if it works.

Mehdi-HAFID:
you only fixed the import problem, the watch:browser does not have `cross-env` at the start.

KazuCocoa:
The close was an automatic one by github. I'll try to check with Win machine when I have time. (regularly I don't use it, so I need to prepare)

KazuCocoa:
Not correct yet, but the package.json should use `SET` to set env ver for Win.
So, we should define like the below command:
```
"watch:browser:win32": "SET BUILD_BROWSER=1 | SET NODE_ENV=development | parcel serve --target browser ./app/renderer/index.html",
```
https://stackoverflow.com/questions/45082648/npm-package-json-os-specific-script

The above one is not correct to use `|`.

Mehdi-HAFID:
I don't understand, I added `cross-env` at the start and it's working. plus cross-env is used in a lot of places in `package.json`. Why would introduce a new command just for Windows, it defeat the purpose of using cross-env package in the first place!!

KazuCocoa:
I think the project has been archived over 2 years ago (https://github.com/kentcdodds/cross-env is the project is my research was correct). Do you think the project will keep working well? I thought adding minimal necessary command for Win was better than adding a new not-maintained dependency (since we need to maintain this project)

PR #562, opened 2022-08-16
fix: ky import

KazuCocoa:
fixes https://github.com/appium/appium-inspector/issues/560

According to https://github.com/appium/appium-inspector/pull/485, the `umd` was dropped in 0.26.0. So we can simply import ky from ky.

I have tested a few commands manually.


jlipps:
thanks kazu!

PR #591, opened 2022-09-01
Adjustments for Italian language

giacomofuria:
Hi! 

As I did [here](https://github.com/appium/appium-gui-libs/pull/7) I am making adjustments to show appium-inspector in Italian language. In this PR I've added the key 'it' in each translations.json file and added a new Key 'JSON Representation' in italian translations.json file.


KazuCocoa:
Thank you.

Could you add the diff in assets/locales/en/translation.json in appium-desktop repo?
https://github.com/appium/appium-desktop/blob/5dfb57f5ccbcc6f479fca0ab2c633e66d37238a3/CONTRIBUTING.md#localization
Other languages will be managed by crowdin, so you do not need to update them.

app/main/menus.js
```diff
@@ -17,7 +17,7 @@ const extraMacFileMenus = [{
 const extraFileMenus = [{
   index: 0,
   menu: {
-    label: i18n.t('&New Session Window…'),
+    label: i18n.t('New Session Window…'),
```
KazuCocoa, line 20: Maybe does `&` need to be the same as https://github.com/appium/appium-gui-libs/pull/7 ?

assets/locales/en/translation.json
```diff
@@ -11,6 +11,7 @@
   "pa-IN": "Punjabi",
   "te": "Telugu",
   "pt-BR": "Portuguese (Brazil)",
+  "it":"Italiano",
```
jlipps, line 14: changing things here will have no effect. these values are copied over from appium desktop (it's weird, i know, and we should probably change the way this works). so you'll need to make a PR to appium desktop first with these changes.

jlipps (changes requested):

KazuCocoa:
Added `it` in https://github.com/appium/appium-desktop/pull/2117 ad a minimal update

PR #625, opened 2022-09-29
I added an implicit wait in the recorder to match more

SuperRoach:
real life usage of the recorder application,

Where the user isn't instantly having buttons be visible.

### DIFF
--- a/app/renderer/lib/client-frameworks/js-wd.js
+++ b/app/renderer/lib/client-frameworks/js-wd.js
@@ -18,6 +18,7 @@ const caps = ${caps};
 
 async function main () {
   await driver.init(caps);
+  await driver.setImplicitWaitTimeout(5000);
 ${this.indent(code, 2)}
   await driver.quit();
 }


jlipps (approved):
sounds good. i will merge this, however the WD framework is going to be removed soon anyway since it is no longer functional with appium 2

brett-james-rocketlab:
If I get time today, I'll update the WDIO version to have an implicit wait as well, But I'll have to test if it's a problem with it anyway :)

It's a shame about WD! it's a less verbose way to write tests.

Will need to find the equivlant of this command in the settings.
http://v4.webdriver.io/api/protocol/timeoutsImplicitWait.html

I have found that the WDIO generated code from the recorder is unusable in general in it's current state - but I'll make another PR when I get around to it to not clog this PR up :)

jlipps:
that would be awesome @brett-james-rocketlab ! yeah, sometimes the recorder template code is not updated when, for example, we upgrade wdio's major version or whatever. so it would be nice for someone to double check its usability and make the appropriate updates!

brett-james-rocketlab:
Do you have a roadmap for the next release after the 8.1 in august by the way?

jlipps:
hmm, there was a publish i forgot to finish, which i just released: https://github.com/appium/appium-inspector/releases/tag/v2022.9.1

beyond that, what do you mean by 'roadmap'?

brett-james-rocketlab:
No, that was exactly what I meant :) Roadmap was just another way of saying when there was a plan of releasing 9.1. Thank you.

PR #633, opened 2022-10-11
fix: save button not enabled after editing raw capability json

### LINKED ISSUE #632: Save button not enabled after editing saved capability using json editor
## Appium or Appium Desktop?

- [X] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## The problem
After editing any saved desired capability from the JSON editor, the save button is not enabled after clicking on save icon. Refer to the below screen shot
appium_inspector_save_button

## Environment

- I am on (pick one):
    - [X] Mac
    - [ ] Windows
    - [ ] Linux

sudharsan-selvaraj:
Fixes #632

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ Certificates.p12
 .idea/
 .settings/
 .project
+.history
 junit-test-results.xml
 junit-testresults.xml
 appium-desktop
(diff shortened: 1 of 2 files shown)


jlipps (approved):
thanks, makes sense!

PR #634, opened 2022-10-11
Fixes resetting of bstack:options

innovater21:
Fixes resetting of user passed bstack:options

Add check before adding source in bstack:options

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -306,7 +306,7 @@ export function newSession (caps, attachSessId = null) {
         port = session.server.browserstack.port = process.env.BROWSERSTACK_PORT || 443;
         path = session.server.browserstack.path = '/wd/hub';
         username = session.server.browserstack.username || process.env.BROWSERSTACK_USERNAME;
-        desiredCapabilities['bstack:options'] = {};
+        desiredCapabilities['bstack:options'] = desiredCapabilities['bstack:options'] ? desiredCapabilities['bstack:options'] : {};
         desiredCapabilities['bstack:options'].source = 'appiumdesktop';
         accessKey = session.server.browserstack.accessKey || process.env.BROWSERSTACK_ACCESS_KEY;
         if (!username || !accessKey) {


KazuCocoa:
Could you sign the CLA?

innovater21:
@KazuCocoa done

app/renderer/actions/Session.js
```diff
@@ -306,8 +306,11 @@ export function newSession (caps, attachSessId = null) {
         port = session.server.browserstack.port = process.env.BROWSERSTACK_PORT || 443;
         path = session.server.browserstack.path = '/wd/hub';
         username = session.server.browserstack.username || process.env.BROWSERSTACK_USERNAME;
-        desiredCapabilities['bstack:options'] = {};
-        desiredCapabilities['bstack:options'].source = 'appiumdesktop';
+        if (desiredCapabilities['bstack:options']) {
+          desiredCapabilities['bstack:options'].source = 'appiumdesktop';
+        } else {
+          desiredCapabilities.source = 'appiumdesktop';
```
jlipps, line 312: so you are intentionally setting a non-w3c style capability to the main level of the capabilities object? why not put it on `bstack:options` to be safe? (and simply create `bstack:options` if it doesn't exist, like the previous code did)

innovater21: The current behaviour was resetting the user's passed bstack:options to a empty object. Due to this all the properties passed in bstack:options during w3c session never reached browserstack.
And at browserstack end we use source to detect any session ( mjsonp or w3c) run using appium desktop 
1. `bstack:options` are passed when it is a w3c session, hence during w3c session , adding source into `bstack:options`
2. In mjsonp sessions, all the w3c, non w3c capabilities are passed directly inside desiredCapabilities, hence added in desiredCapabilities
I hope this answers your question @jlipps

innovater21:
please check the above comment @jlipps

innovater21:
Please review @jlipps

innovater21:
cc : @KazuCocoa

KazuCocoa:
He will get back to here maybe next week, fyi

app/renderer/actions/Session.js
jlipps: i guess i still don't see why you wouldn't just do this:

```js
if (!desiredCapabilities['bstack:options']) {
  desiredCapabilities['bstack:options'] = {};
}
desiredCapabilities['bstack:options'].source = 'appiumdesktop';
```

this also presumably fixes the bug but doesn't fall back to using mjsonwp capabilities. we should really only use w3c caps nowadays.

innovater21:
can you please check @jlipps 
cc : @KazuCocoa

KazuCocoa:
Did you check jlipps's reply? https://github.com/appium/appium-inspector/pull/634#discussion_r1003621048

app/renderer/actions/Session.js
innovater21: `bstack:options` is only used for w3c sessions
For sessions being run from older appium desktop client, the source property is not being added to MJSONWP session.
With the above suggestion source property gets skipped for MJSONWP session hence can't do it using that way

Let me know if there are any other concerns

jlipps: I don't understand your point. This change is only for new appium inspector clients, which use the W3C strategy to create sessions.

innovater21: yeah understood. We just need the change for W3C and handling for MJSONWP session is not required

innovater21: @jlipps please review

jlipps (approved):
Yes, this looks great now! Will wait for CI to pass and then we can merge.

innovater21:
Thanks @jlipps for the quick review. All the checks looks good

PR #635, opened 2022-10-15
feat: show spinner on source tree when mjpegScreenshotUrl is provided

### LINKED ISSUE #504: feat request: show spinner over source instead of screenshot when viewing an MJPEG stream
## The problem

Mjpeg video mode is still showing loader when tap is called though it is video/screen cast
image

## Environment

- I am running Appium Desktop version _2022.5.4_.
- I am on (pick one):
    - [ ] Mac

sudharsan-selvaraj:
Fixes #504 

I have tested it for both scenarios with and without `mjpegScreenshotUrl`

### DIFF
--- a/app/renderer/components/Inspector/Source.js
+++ b/app/renderer/components/Inspector/Source.js
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { Tree } from 'antd';
+import { Tree, Spin } from 'antd';
 import LocatorTestModal from './LocatorTestModal';
 import InspectorStyles from './Inspector.css';
 import { withTranslation } from '../../util';
@@ -60,6 +60,8 @@ class Source extends Component {
       selectedElement = {},
       showSourceAttrs,
       t,
+      mjpegScreenshotUrl,
+      methodCallInProgress,
     } = this.props;
     const {path} = selectedElement;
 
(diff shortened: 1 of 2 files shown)


app/renderer/components/Inspector/Source.js
```diff
@@ -78,18 +80,20 @@ class Source extends Component {
...
-          onExpand={setExpandedPaths}
-          expandedKeys={expandedPaths}
-          onSelect={(selectedPaths) => this.handleSelectElement(selectedPaths[0])}
-          selectedKeys={[path]}
-          treeData={treeData} />
-        :
-        <Tree
-          treeData={[]} />
-      }
+      <Spin size='large' spinning={!!methodCallInProgress && !!mjpegScreenshotUrl}>
```
jlipps, line 83: are you sure `!!mjpegScreenshotUrl` is correct? you only want to show the spinner when `mjpegScreenshotUrl` is set?

jlipps (commented):
i'm not sure how i feel about this change. i kind of like there to be a spinner during a method call so the user knows something is happening. i can understand that we may not want it on the screenshot when we're in mjpegscreenshot mode. but maybe there's another place we can put it instead in that case?

eglitise:
Changes have been merged as part of #884. FYI @sudharsan-selvaraj

Issue #615, opened 2022-09-20
Recorder generated scripts do not fetch coordinates


Maydrian:
## Recorder generated scripts do not fetch coordinates
Setup: Connecting with inspector on windows machine to a remote appium server machine (Ubuntu).
All coordinate based recordings like swipe or tap at specific location do not record the actual coordinates.
Instead the generated code refers to them as ([object Object],undefined) which might be some sort of null reference.

Python Code snipped which was recorded by tapping on Coordinates X:500; Y :500:
```
actions = ActionChains(driver)
actions.w3c_actions = ActionBuilder(driver, mouse=PointerInput(interaction.POINTER_TOUCH, "touch"))
actions.w3c_actions.pointer_action.move_to_location([object Object], undefined)
actions.w3c_actions.pointer_action.pointer_down()
actions.w3c_actions.pointer_action.pause(0.1)
actions.w3c_actions.pointer_action.release()
actions.perform()
```
This issue persists over all available output languages.

## Environment

- I am running Appium Inspector version _2022.8.1_ Windows.
- I am running Appium server (remote) version _2.0.0-beta.44_ running on Ubuntu 20.04.
- Unit under test: Google Pixel 2; Android 11
- I am on:
    - [ ] Mac
    - [X] Windows
    - [ ] Linux

rushian:
I will check if the problem is in GestureEditor.js

jlipps:
this is not an issue with gestureeditor, but with the code generation

secretrobotron:
Confirming same problem on inspector with both Mac and Windows on two different Android devices on different OS versions.

PR #686, opened 2022-12-16
fix: added utils for encoding tap/swip coords properly

secretrobotron:
The method for calling the various `codeFor_*` functions doesn't seem to be aligned with the way that the corresponding actions are constructed to execute a task. E.g., a swipe consists of _move, down, pause, release_, under a specific pointer, _finger1_. These functions are, instead, expecting x and y coordinates.

This patch add some util functions to the `Framework` class so that its subclasses can write the appropriate code for _tap_ and _swipe_ actions by unpacking the more complex pointer action object that is constructed elsewhere within the inspector.

fixes: #615

### DIFF
--- a/app/renderer/lib/client-frameworks/framework.js
+++ b/app/renderer/lib/client-frameworks/framework.js
@@ -1,3 +1,5 @@
+import { DEFAULT_TAP, DEFAULT_SWIPE } from '../../components/Inspector/shared';
+
 export default class Framework {
 
   constructor (host, port, path, https, caps) {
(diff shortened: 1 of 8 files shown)


secretrobotron:
Btw, I'm a bit of a webpack / react debutant. If there is a more elegant way to do this, lmk and I'm happy to fix.

rushian:
Thank you for help!!!

app/renderer/lib/client-frameworks/framework.js
```diff
@@ -13,6 +15,23 @@ export default class Framework {
     this.lastAssignedVar = null;
   }
 
+  static getTapCoordinatesFromPointerActions (pointerActions) {
```
KazuCocoa, line 18: `static` can be removed. Then, the caller like in java.js can call this method as `this.getTapCoordinatesFromPointerActions` instead of `Framework.getTapCoordinatesFromPointerActions`
```suggestion
  getTapCoordinatesFromPointerActions (pointerActions) {
```

(The below is also the same)

secretrobotron: Ah right. Done.

KazuCocoa (approved):
thank you!

jlipps (commented):
thanks for the contribution!

PR #730, opened 2023-02-28
Fix App Source/Selected Element sections changing size

TDL-EdgarsEglitis:
This PR fixes issue #664.

PR #744, opened 2023-03-23
style: improve CSS on inspector screen

eglitise:
Following up from #730, I noticed a few other quick CSS fixes on the inspector screen:

* Ensure vertical centering for the source tree container header (now matches the selected element container header)
* Align the bottom coordinates of the screenshot and selected element containers with the source tree container
  * This also fixes the issue where, if an element was selected or the source tree expanded, the panels vertically expanded by 14px. Still not sure of the root cause, but the symptoms are fixed.


jlipps:
thanks!

PR #746, opened 2023-03-23
feat: add device action buttons and allow pausing of source refreshing

JoelWhitney:
#### Overview
I'm currently utilizing physical devices that are plugged into a Mac mini across the country. Periodically I will need to manually interact with these devices. Currently, this is very cumbersome using Appium inspector because there are no hardware buttons in the GUI and the page source/screenshot is refreshed every time a command is executed. To make device interactions more seamless, I would like to add some hardware buttons to the GUI and provide a way to pause page source refreshing. When coupled with a mjpeg stream, this makes interacting with devices more streamlined.

#### Changes made
- Provides session capabilities to webdriver when attaching to session
- Adds device actions for iOS: Home and Siri buttons
- Adds device actions for Android: Home, Back, and Overview buttons
- Adds button to pause page source refreshing
- Gets mjpegScreenshotUrl from session capabilities if not explicitly provided in desired capabilities

#### iOS
ios

#### Android
android


JoelWhitney:
I just saw this PR, which is somewhat related
https://github.com/appium/appium-inspector/pull/635

jlipps:
thanks for this contribution @JoelWhitney! will take a look soon

jlipps:
@JoelWhitney are you able to merge main so I can review knowing there are no conflicts?

JoelWhitney:
@jlipps looks like it's up to date, I merged main this morning

app/renderer/actions/Inspector.js
```diff
@@ -692,7 +721,7 @@ export function keepSessionAlive () {
 
 export function callClientMethod (params) {
   return async (dispatch, getState) => {
-    const {driver, appMode, mjpegScreenshotUrl} = getState().inspector;
+    const {driver, appMode, mjpegScreenshotUrl, isRefreshingSource} = getState().inspector;
```
jlipps, line 724: can we change this to `shouldRefreshSource` or `isSourceRefreshOn`? `isRefreshingSource` sounds like a state variable tracking whether the source is in the process of being refreshed, which I don't think is what you intend.

app/renderer/components/Inspector/DeviceActions.js
```diff
@@ -0,0 +1,43 @@
+import React, { Component } from 'react';
+import { Button, Tooltip } from 'antd';
+import InspectorStyles from './Inspector.css';
+import { withTranslation } from '../../util';
+import { HiOutlineMicrophone, HiOutlineHome } from 'react-icons/hi';
+import { BiSquare, BiCircle } from 'react-icons/bi';
```
jlipps, line 6: can we not use icons from ant design, which are already included?

app/renderer/util.js
```diff
@@ -283,3 +283,36 @@ export function addVendorPrefixes (caps) {
...
+ * @param  {Object}  caps  caps of session response
+ * @return {Boolean}       true if run on iOS device
+ */
+export function isIOS (caps) {
+  if (!caps) {
+    return false;
+  }
+
+  return Boolean(
+      (caps.platformName && caps.platformName.match(/iOS/i)) ||
```
jlipps, line 299: since the `platformName` cap is required for every session, we should theoretically be able to look just at that and not worry about deviceName, for both iOS and Android.

jlipps (changes requested):
this looks great! made a few small comments

app/renderer/actions/Inspector.js
JoelWhitney: Yup, I see what you mean. Updated to `isSourceRefreshOn`

app/renderer/util.js
JoelWhitney: Okay, since `platformName` cap is required, I can probably remove these util functions and move the match check out to `Session.js` since that was the only place these functions were used.

Moved here: https://github.com/appium/appium-inspector/pull/746/commits/a2618b0e892d7ea8347dcea2e3e3f904969cca42#diff-cc0a8b1e7efd7e5b9ee01a00c256283251315b2c0c18cca08911fef1c01fa0b9R530-R531

JoelWhitney:
@jlipps thanks for reviewing. I added some comments above and pushed a commit to address some of the feedback. Let me know if you have suggestions for addressing the icon feedback

jlipps:
@JoelWhitney in that case I'm fine to keep the new icons. thanks! also I tried to merge in main but it looks like there are now more conflicts again, due to merging other PRs. do you mind fixing conflicts and pushing once more? then i'll be able to get this merged and published.

JoelWhitney:
@jlipps everything appears up to date. I fetched upstream and merged, which returned `Already up to date.`. I wasn't able to find any conflicts either, what conflicts did you see?

jlipps:
sorry @JoelWhitney i had the merge button set to 'rebase' instead of 'squash' so it was showing as unrebasable due to conflicts. my fault!

PR #751, opened 2023-03-31
fix: do not override`tb:options` when specified

testingbot:
This changes makes sure we do not override any `tb:options` specified by the user.
If a user entered their own `tb:options`, it should not be overridden.


jlipps (approved):

jlipps:
thanks!

PR #752, opened 2023-03-31
style: UX improvements for session screen

eglitise:
This PR focuses on UX improvements for the session screen.
* Compacted listing of server parameters
  * Parameters for all server types now take up only 1 row (exceptions being SauceLabs and Headspin). I find all the elements still perfectly usable even at minimum window width.
* Reduced minimum window height from 800px to 700px
  * No scaling issues were encountered at this new minimum height.
* Use antd Badge to render the saved capability set count
  * Enhanced visual clarity (and animations!)
  * Changes in translation files were required here. As per the guidelines, I have created a PR at appium-desktop with these same changes: [appium-desktop#2155](https://github.com/appium/appium-desktop/pull/2155)
* Improved editing experience in the formatted capabilities panel
  * Text font now stays the same in editing mode
  * The editable text area does not default to a 9-row height, but instead scales to the maximum available area, along with the inspector window. Due to this, manual rescaling for the text area itself has been disabled.
  * JSON warnings cannot be potentially hidden by the inspector footer anymore; instead, they reduce the height of the editable text area
* Improved usability for the saved capability sets table
  * Header is now sticky
  * Reduced padding between rows (I think this helps if you have a lot of saved sets - I personally have 40)
* Reduced margins/padding in a few other places

Here is a short video showcasing most of these changes:

https://user-images.githubusercontent.com/37242620/229135316-effbb270-9d75-4559-a154-8f2b698b58d4.mov


jlipps (approved):
this is great, thanks so much!

PR #757, opened 2023-04-05
chore(deps): update @appium/fake-driver to 5.2.3

eglitise:
This PR updates @appium/fake-driver to ~5.2.2~ 5.2.3.

A few notes:

1. It seems that `@appium/fake-driver` has dropped support for Appium 1.x. Since Appium Inspector only uses this package as a dev dependency, this will not affect regular users. However, **integration tests will now only run on Appium 2**.
2.  While investigating the integration tests, I noticed that `web2driver` is still setting the default path to `/wd/hub`. I've changed the path here, but perhaps this is something worth updating [at the source](https://github.com/headspinio/web2driver/blob/master/src/driver.js)?
3. ~Imports for `appium-support` were changed to `@appium/support`, except for one file in [appium-gui-libs](https://github.com/appium/appium-gui-libs/blob/5239eebaf0eab9078d9a0c50787bde30ecaf2c81/crowdin-sync-translations.js). Would it be correct to change that file as well, at risk of affecting Appium Desktop?~ Resolved now that appium-gui-libs has been pulled into this repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -178,7 +178,8 @@
     "postcss-modules": "Can't upgrade any further until we upgrade to Parcel v2"
   },
   "devDependencies": {
-    "@appium/fake-driver": "2.2.0",
+    "@appium/fake-driver": "5.2.2",
+    "@appium/support": "3.1.9",
     "@babel/core": "7.21.4",
     "@babel/eslint-parser": "7.21.3",
     "@babel/eslint-plugin": "7.19.1",
--- a/package.json
+++ b/package.json
@@ -178,8 +178,8 @@
     "postcss-modules": "Can't upgrade any further until we upgrade to Parcel v2"
   },
   "devDependencies": {
-    "@appium/fake-driver": "5.2.2",
-    "@appium/support": "3.1.9",
+    "@appium/fake-driver": "5.2.3",
+    "@appium/support": "3.1.11",
     "@babel/core": "7.21.4",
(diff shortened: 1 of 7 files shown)


jlipps:
@eglitise yes, Appium Desktop is no longer supported. We should probably just bring gui-libs into this repo and stop maintaining a separate repo/submodule.

package.json
jlipps, line 1: we should also be able to remove the `appium-support` dependency now from this file as well, if it exists?

eglitise: This file did not include the `appium-support` dependency, likely relying on the import from `@appium/fake-driver`. But since it is being used as an import, it's probably safer to explicitly add `@appium/support` as a dependency.

eglitise:
I suppose I will leave this PR on pause until appium-gui-libs is migrated to this repo.

jlipps:
yeah, makes sense.

eglitise:
Rebased this after the appium-gui-libs PR, but now I'm not sure why CI fails integration tests 🤔 They run fine on my Mac...

jlipps:
hmm I wonder if it's because appium fake driver requires appium as a peer dependency now, and for the npm that comes with node 14, peer deps aren't automatically installed if missing? we should maybe adjust our CI scripts to install the latest npm even with the older versions of node.

eglitise:
That does seem to be the likely reason. I've created #791 to try and remedy this.

PR #768, opened 2023-04-11
chore: update readme and fix Android system buttons

eglitise:
This PR fixes a few minor things in #746: 
* pause source refresh button
  * turns out that if MJPEG stream is not used, toggling this button also stops refreshing screenshot - this has been added to the readme
* system buttons
  * rename Android Overview button to App Switch, according to [Android Developer reference](https://developer.android.com/reference/android/view/KeyEvent#KEYCODE_APP_SWITCH)
  * swap keycodes for Android Back and Android Home buttons, also according to Android Developer reference
  * add more details in readme


jlipps (approved):

jlipps:
thanks for this @eglitise ! for future reference i would probably call the commits that you call `chore` here also `fix` since they're actually user-facing changes

PR #772, opened 2023-04-13
fix: remove leftover 'savedSessionsCount'

eglitise:
This is a minor fix for [commit e87bf9b](https://github.com/appium/appium-inspector/commit/e87bf9b393d8972725f23df5e59aa1f69073bd75).

Also, it seems that CI is failing (as it cannot build Inspector v2023.4.1). This is likely related to the package-lock changes in the same commit.


jlipps (approved):

jlipps:
that makes sense. We upgraded the version of NPM used in CI and that seems to have helped!

PR #773, opened 2023-04-13
ci: crowdin setup

KazuCocoa:
Trying to move crowdin update task from https://github.com/appium/appium-desktop/blob/master/ci-jobs/master-build.yml.

- set CROWDIN_PROJECT_KEY
- Check the behavior with this appium-inspector resource
- Move translation documentation from https://github.com/appium/appium-desktop


KazuCocoa:
let me test this script (copied from the appium-desktop as-is for now) if this works expectedly

PR #780, opened 2023-04-18
chore: remove copy-en-i18n.js and related code

eglitise:
I believe Crowdin is now configured to use this repo instead of Appium Desktop. As a result, the `copy-en-i18n.js` script to copy translations is obsolete. This PR removes it, along with its related code (calls in CI configuration files + unique package dependencies).
The README is also updated. I hope my description of the process is correct, please let me know if something is inaccurate.


jlipps (approved):
LGTM, we should probably also have @KazuCocoa review. Thanks for all your help @eglitise !

KazuCocoa (approved):
Looks good to me

PR #781, opened 2023-04-18
feat: support renaming saved capability sets

eglitise:
This PR adds the ability to rename any saved capability set, through a new edit button next to the title.

Why is this useful? I have had situations where I need to replace one of my test phones, for which I had saved multiple capability sets. In order to update the caps and the set names, previously I would need to create entirely new sets, and delete the old ones. With this change, I only need to edit caps for the device udid/name, and rename the set title.

The implementation is fairly similar to that of editing the saved capabilities themselves. One key difference is that the new title is saved immediately after pressing the small save button - the Save button in the footer is not used.

Here is a brief video to show how this works:

https://user-images.githubusercontent.com/37242620/232803329-94b9c04e-57bf-4992-9157-e4b012bf7e74.mp4

Also, the changes in `SavedSessions.js` are not directly related, but they greatly reduce the number of calls to `sessionFromUUID` when switching between saved sets.

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -48,8 +48,8 @@ export const GET_SESSIONS_DONE = 'GET_SESSIONS_DONE';
 
 export const ENABLE_DESIRED_CAPS_NAME_EDITOR = 'ENABLE_DESIRED_CAPS_NAME_EDITOR';
 export const ABORT_DESIRED_CAPS_NAME_EDITOR = 'ABORT_DESIRED_CAPS_NAME_EDITOR';
-export const SAVE_RAW_DESIRED_CAPS_NAME = 'SAVE_RAW_DESIRED_CAPS_NAME';
-export const SET_RAW_DESIRED_CAPS_NAME = 'SET_RAW_DESIRED_CAPS_NAME';
+export const SAVE_DESIRED_CAPS_NAME = 'SAVE_DESIRED_CAPS_NAME';
+export const SET_DESIRED_CAPS_NAME = 'SET_DESIRED_CAPS_NAME';
 
 export const ENABLE_DESIRED_CAPS_EDITOR = 'ENABLE_DESIRED_CAPS_EDITOR';
 export const ABORT_DESIRED_CAPS_EDITOR = 'ABORT_DESIRED_CAPS_EDITOR';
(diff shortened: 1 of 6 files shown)


app/renderer/actions/Session.js
```diff
@@ -869,6 +874,32 @@ export function getRunningSessions () {
...
+  };
+}
+
+export function abortDesiredCapsNameEditor () {
+  return (dispatch) => {
+    dispatch({type: ABORT_DESIRED_CAPS_NAME_EDITOR});
+  };
+}
+
+export function saveRawDesiredCapsName () {
```
jlipps, line 889: i think the word `raw` here is a bit unnecessary. we use `raw` for the other aspect of the caps editor because the user is editing raw JSON instead of using the UI. but in the case of the title, there is no distinction between raw or non-raw. so i think it's not necessary to make that distinction in code either. i'd just use `saveDesiredCapsName` here instead (and as well in other places where 'raw' is used for the title)

app/renderer/components/Session/FormattedCaps.js
```diff
@@ -17,9 +17,36 @@ export default class FormattedCaps extends Component {
   }
 
   render () {
-    const {caps, title, isEditingDesiredCaps, startDesiredCapsEditor, abortDesiredCapsEditor, saveRawDesiredCaps, setRawDesiredCaps, rawDesiredCaps,
+    const {caps, title, isEditingDesiredCapsName, startDesiredCapsNameEditor, abortDesiredCapsNameEditor, saveRawDesiredCapsName, setRawDesiredCapsName, rawDesiredCapsName,
+           isEditingDesiredCaps, startDesiredCapsEditor, abortDesiredCapsEditor, saveRawDesiredCaps, setRawDesiredCaps, rawDesiredCaps,
            isValidCapsJson, invalidCapsJsonReason, t} = this.props;
-    return caps && <Card title={title || 'JSON Representation'} className={SessionStyles.formattedCaps}>
+    return caps && <Card className={SessionStyles.formattedCaps}
+      title={!title ? 'JSON Representation' : (!isEditingDesiredCapsName ? title :
```
jlipps, line 24: for readability, rather than using a nested ternary i would break all this out into a set of if/then statements above culminating in one variable that can be assigned to `title` here. the same strategy might also improve the readability of the sections below.

jlipps (changes requested):
this looks great! just have one or two small changes in mind

eglitise:
In addition to review changes, I've added a fix that disables any open editors when the user changes the selected capability set. Before this, it was possible to accidentally overwrite the caps/title of a different set.

jlipps (approved):

PR #790, opened 2023-04-24
fix: rename Appium Desktop -> Appium Inspector everywhere

eglitise:
This PR replaces all (replaceable) instances of Appium Desktop with Appium Inspector. Exceptions include strings for API integration with third parties, like Crowdin and cloud providers, which were left as-is.

Translation file changes also include renaming of 2 keys and removal of 1 unused key.


jlipps (approved):

jlipps:
👏

PR #791, opened 2023-04-25
ci: cleanup validation workflow

eglitise:
This PR ensures that CI runs for Node 14 use NPM 8 instead of the default NPM 6. This is because NPM 6 does not auto-install peer dependencies, which are used in newer `@appium/fake-driver` versions.

I used `npm@8` instead of `npm@latest` to prevent any breaking changes in the future.

Worth noting that [Node 14 support ends on 2023-04-30](https://github.com/nodejs/release#release-schedule), so maybe it's simpler to just wait a few days and then remove it from CI entirely? 🤔 

Also, this is a good place to ask, perhaps it's worth removing the `on: pull_request` trigger in `node.build.yml`? That workflow is already fully covered by the PR validation workflow.

### DIFF
--- a/.github/workflows/node.build.yml
+++ b/.github/workflows/node.build.yml
@@ -6,8 +6,6 @@ name: Node.js CI
 on:
   push:
     branches: [ main ]
-  pull_request:
-    branches: [ main ]
 
 jobs:
   build:
--- a/.github/workflows/node.build.yml
+++ b/.github/workflows/node.build.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [16.x, 18.x]
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
 
     steps:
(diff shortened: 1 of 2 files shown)


jlipps:
yeah, I'm fine if we want to just wait a few days and then remove node 14 support from ci entirely. i mean theoretically i am not sure why we even test on multiple versions of node to begin with. we're building a GUI app and shipping it to end users. it's not a library. @KazuCocoa do you know?

KazuCocoa:
i don't remember, but I guess that was for dev environment?

jlipps:
Yeah probably. In that case I think we can just remove Node 14 now, since I don't think we need to support all node LTSs just for the inspector's dev environment.

eglitise:
I've now changed the PR to instead remove Node 14.
What about the `on: pull_request` trigger in `node.build.yml`, is it worth keeping that?

KazuCocoa:
yea, it sounds reasonable

KazuCocoa (approved):

eglitise:
The on PR trigger for `node.build.yml` has now also been removed 👍

jlipps:
👍

PR #795, opened 2023-04-28
fix: improve translation use

eglitise:
This PR fixes various translation-related things:
* fix missing links between some keys and their uses
* rename a few keys
* remove many unused keys (mainly leftovers from Appium Desktop)
* add many keys which already had uses

Note that there are still several remaining locations with hardcoded English strings and no translation key. These have not been changed.


KazuCocoa:
btw, translation change can be only in `en`. https://github.com/appium/appium-inspector/actions/workflows/translation.yml will create a pr to update other languages

KazuCocoa:
Can you remove keys modification such as `Execute Mobile` in this pr for now?

KazuCocoa (approved):
👍 thank you!

PR #803, opened 2023-05-08
feat: add class name locator strategy where applicable

eglitise:
This PR adds the class name locator strategy in the 'Find By' table when selecting elements with unique class names.
Considering that this strategy is listed as the fastest for both [iOS](https://github.com/facebookarchive/WebDriverAgent/wiki/How-To-Achieve-The-Best-Lookup-Performance#select-the-most-effective-lookup-strategy) and [Android](https://github.com/appium/appium-uiautomator2-driver#element-location), I think it is worth adding.
Android uses the `class` element property, whereas iOS uses `type`.

Example:
image

### DIFF
--- a/app/renderer/components/Inspector/shared.js
+++ b/app/renderer/components/Inspector/shared.js
@@ -50,6 +50,8 @@ const STRATEGY_MAPPINGS = [
   ['id', 'id'],
   ['rntestid', 'id'],
   ['resource-id', 'id'],
+  ['class', 'class name'],
+  ['type', 'class name'],
 ];
 
 export function getLocators (attributes, sourceXML) {


jlipps:
Before I review the code, probably worth having a conceptual discussion around this. The id/xpath/ax id selectors are presented as ways to identify _this_ element. class name is not unique and therefore is not typically going to actually successfully find the element under consideration except by accident. I'm not sure if this PR already includes this, but I think we would only want to show the class name as a selector if we can determine that it's uniquely targeting the right element (i.e., there is only one element of that class on the page). Or that we should have a warning icon or something.

eglitise:
@jlipps I fully agree. Conveniently, the existing code [already had a uniqueness check](https://github.com/appium/appium-inspector/pull/803/files#diff-a6642aaeb8cb65d91f2271de4f964373dc06aa539befadde5862c4ad9251fc6bR61) for all the optimal locator strategies, so I did not need to change anything there :) The class name row is therefore only shown for the elements where it can be used as a unique identifier.

PR #806, opened 2023-05-09
fix: revert electron-settings IPC refactor

eglitise:
This PR reverts most of the changes in #801. This is because I did not validate them in the browser build, and as it turns out, it completely breaks.
The reason behind this problem is the two electron imports in `helpers.js`. Ideally they should be imported conditionally, like in `polyfills/index.js`, but it seems that `PREFERRED_LANGUAGE` is requested before they finish loading, so it may not be assigned correctly.
I was researching many solutions but none seemed to work (at least with my current JS/React knowledge). It is possible that upgrading to parcel v2 could help here, but that is not a simple fix either. So for now I think it is best to revert this.
Still, not all of the commits in the original branch were breaking, so I have kept a few of them.

### DIFF
--- a/app/configs/app.config.js
+++ b/app/configs/app.config.js
@@ -1,3 +1,5 @@
+import settings from '../shared/settings';
+
 const config = {
   platform: process.platform,
   languages: [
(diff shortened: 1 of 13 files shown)


jlipps (approved):
oh that's too bad! it's possible i knew this before, and then forgot :-) but yeah the browser version is super important. maybe there's a way we can test the browser build in CI so we get quicker feedback

PR #807, opened 2023-05-09
fix: update link to docs (fix #603)

### LINKED ISSUE #603: In Session window the link to Capabilities Documentation is broken
## Appium or Appium Desktop?

- [ X ] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## The problem

When we click at Desired Capabilities Documentation link in the Session window it open the broked link:
https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/caps.md

instead 
http://appium.io/docs/en/writing-running-appium/caps/index.html

## Environment

- I am on (pick one):
    - [ ] Mac
    - [ x ] Windows
    - [ ] Linux

jlipps:


KazuCocoa (approved):

eglitise:
Ok, after approving I realised that the app already links to the main documentation through its Help menu. Maybe it's better to instead keep the current text, and just change the link to the [Capabilities page of the new docs](https://appium.io/docs/en/latest/guides/caps/)?
The Help menu links are also a bit outdated though...

jlipps:
good point @eglitise i updated them all

jlipps:
looks like that was unnecessarily duplicated. what do you think now @eglitise

eglitise (approved):
Looks good now 👍

PR #825, opened 2023-05-15
chore(deps): remove npm-run-all

eglitise:
This PR removes the dev dependency `npm-run-all`.
It has not received a new version in more than 3 years, and its only use is shortening some npm scripts, which are called through a shorthand anyway.


jlipps (approved):
yeah makes sense

PR #828, opened 2023-05-16
chore: add comments to keep track of dependencies with issues

eglitise:
I did some investigation of all the dependencies in the project, and compiled a list of the ones that are likely to be outdated (for one reason or another). Since it is non-trivial to resolve most of these, I thought it best to expand the existing comments sections, in order to keep track.


jlipps (commented):
this is great work, thanks @eglitise! one looming issue for the project is whether we want to upgrade react to function style components / hooks.

would be a lot of work to totally modernize this codebase so not sure it's worth it, but maybe we could do it incrementally.

PR #842, opened 2023-05-23
style: enhance find element modal

eglitise:
This PR provides various visual improvements to the find element modal:
* Replace locator strategy dropdown with radio buttons (one less click and no scroll)
* Hide locator strategies that are unavailable for current `automationType`
* Render results count using antd Badge
* Render results list with responsive height (1-6 rows)
* Replace text for found element action buttons with icons + tooltip
* Add a clear button to selector and send keys input fields if text is entered

Here is a demonstration of the new behavior:

https://github.com/appium/appium-inspector/assets/37242620/3c553e19-bb62-4535-a95a-d4a63377f70f

### DIFF
--- a/app/renderer/components/Inspector/ElementLocator.js
+++ b/app/renderer/components/Inspector/ElementLocator.js
@@ -3,35 +3,35 @@ import { Input, Radio, Row } from 'antd';
 import InspectorStyles from './Inspector.css';
 import { withTranslation } from '../../util';
 
-class ElementLocator extends Component {
-
-  getLocatorStrategies (driver) {
-    let baseLocatorStrategies = [
-      ['id', 'Id'],
-      ['xpath', 'XPath'],
-      ['name', 'Name'],
-      ['class name', 'Class Name'],
-      ['accessibility id', 'Accessibility ID']
-    ];
-    if (driver.client.isIOS) {
+const locatorStrategies = (driver) => {
+  let baseLocatorStrategies = [
+    ['id', 'Id'],
+    ['xpath', 'XPath'],
+    ['name', 'Name'],
+    ['class name', 'Class Name'],
+    ['accessibility id', 'Accessibility ID']
+  ];
+  if (driver.client.isIOS) {
+    baseLocatorStrategies.push(
+      ['-ios predicate string', 'Predicate String'],
+      ['-ios class chain', 'Class Chain']
+    );
+  } else if (driver.client.isAndroid) {
+    if (driver.client.capabilities.automationName.toLowerCase() === 'espresso') {
       baseLocatorStrategies.push(
-        ['-ios predicate string', 'Predicate String'],
-        ['-ios class chain', 'Class Chain']
+        ['-android datamatcher', 'DataMatcher'],
+        ['-android viewtag', 'View Tag']
       );
-    } else if (driver.client.isAndroid) {
-      if (driver.client.capabilities.automationName.toLowerCase() === 'espresso') {
-        baseLocatorStrategies.push(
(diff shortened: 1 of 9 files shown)


mykola-mokhnach:
Would it maker sense to sort location strategies by performance (e.g. most performant go first)?

eglitise:
It probably could be done, but I don't see much benefit to it. As a design, the proposed 2-row radio button style (in my opinion) doesn't really imply any sorting, unlike the current dropdown design. And different automation engines may then require different list orders, which could be annoying when working with more than one engine.

mykola-mokhnach (approved):

KazuCocoa (approved):
lgtm 👍

app/renderer/components/Inspector/ElementLocator.js
```diff
@@ -29,38 +56,37 @@ class ElementLocator extends Component {
...
-          value={locatorTestStrategy}>
-          {locatorStrategies.map(([strategyValue, strategyName]) => (
-            <Option key={strategyValue} value={strategyValue}>{strategyName}</Option>
+    return <Row justify="center">
+      <Radio.Group buttonStyle="solid"
+        className={InspectorStyles.locatorStrategyGroup}
+        onChange={(e) => setLocatorTestStrategy(e.target.value)}
+        defaultValue={locatorTestStrategy}
+      >
+        <Row justify="center">
```
jlipps, line 69: I think this row should have some kind of label like 'Strategy' in case someone isn't super familiar with what these are yet

app/renderer/components/Inspector/ElementLocator.js
```diff
@@ -29,38 +56,37 @@ class ElementLocator extends Component {
...
-      </Row>
-      <Row>
-        {t('selector')}
-        <Input.TextArea className={InspectorStyles['locator-strategy-selector']} onChange={(e) => setLocatorTestValue(e.target.value)} value={locatorTestValue} />
-      </Row>
-    </div>;
+        </Row>
+      </Radio.Group>
+      <Input.TextArea
+        className={InspectorStyles.locatorSelectorTextArea}
```
jlipps, line 82: i think this field should have a label like 'Selector' or similar

app/renderer/components/Inspector/LocatedElements.js
```diff
@@ -31,60 +34,65 @@ class LocatedElements extends Component {
...
+              <Button
+                disabled={!locatorTestElement}
+                icon={<AimOutlined/>}
+                onClick={() => applyClientMethod({methodName: 'click', elementId: locatorTestElement})}
+              />
+            </Tooltip>
+            <ButtonGroup className={InspectorStyles.searchResultsActions}>
+              <Input className={InspectorStyles.searchResultsKeyInput}
+                disabled={!locatorTestElement}
+                placeholder={t('Enter keys')}
```
jlipps, line 73: maybe something like 'Keys to send' would be slightly more understandable?

jlipps (commented):
great idea! made a few suggestions

app/renderer/components/Inspector/ElementLocator.js
eglitise: Added back the original label

app/renderer/components/Inspector/ElementLocator.js
eglitise: I had moved the label to the placeholder text instead, but to maintain consistency with the strategy label, I've now moved it back.

app/renderer/components/Inspector/LocatedElements.js
eglitise: Changed to 'Enter Keys to Send'

PR #845, opened 2023-05-26
feat: add button to select found element in source

eglitise:
This PR adds a new button in the found elements modal, that allows to reveal the selected element in the current application source. I've personally found myself in need of such a feature multiple times.

This was rather tricky to implement, since elements found through the find element option do not have an assigned absolute path (unlike the elements that are selected in source, or using screenshot highlighter rectangles). The only useful parameter here is the element bounds, so that is what I used. It works as follows:

1) Iterate through the cached JSON source tree and identify all nodes whose bounds match the expected element bounds
2) If multiple nodes are found, run Appium `findElement` on each node, and compare their `elementId`s to the expected element
3) With the correct node found, call `selectElement` using its path

I've also added a new error message in cases where the app source is outdated, or if there are too many nodes with the same bounds (currently this limit is set to 10).

Here is a brief demonstration. Note that at 1:05 I am manually returning the previous screen, in order to intentionally invalidate the found element.

https://github.com/appium/appium-inspector/assets/37242620/3da3bf4c-7fac-4cb3-9739-2b53333f97cd


app/renderer/actions/Inspector.js
```diff
@@ -510,6 +512,107 @@ export function setLocatorTestElement (elementId) {
...
+ * Given an element ID found through search, and its bounds,
+ * attempt to find and select this element in the source tree
+ */
+export function selectLocatedElement (source, bounds, id) {
+  const UPPER_FILTER_LIMIT = 10;
+
+  // Parse the source tree and find all nodes whose bounds match the expected bounds
+  // Return the path + xpath of each node
+  function findPathsMatchingBounds () {
+    if (bounds && source.children && source.children[0].attributes) {
```
mykola-mokhnach, line 525: I would rather flip this condition and make it to a guard.
Guards are easier to read than deeply nested ifs

app/renderer/actions/Inspector.js
```diff
@@ -510,6 +512,107 @@ export function setLocatorTestElement (elementId) {
...
+        collectedPaths.push(...findPathsFromBounds(tree.children, bounds));
+      }
+    }
+    return collectedPaths;
+  }
+
+  // If findPathsMatchingBounds found multiple items,
+  // use Appium findElement to filter further by element ID
+  async function filterFoundPaths (foundPaths, dispatch, getState) {
+    if (foundPaths) {
```
mykola-mokhnach, line 575: same as above. turn it to a guard

app/renderer/actions/Inspector.js
```diff
@@ -510,6 +512,107 @@ export function setLocatorTestElement (elementId) {
...
+    return null;
+  }
+
+  // Calls Appium findElement for each provided xpath, and returns the path
+  // of the element whose ID matches the expected ID
+  async function findElementWithMatchingId (foundPaths, dispatch, getState) {
+    for (const path of foundPaths) {
+      const action = callClientMethod({strategy: 'xpath', selector: path[1]});
+      const { el } = await action(dispatch, getState);
+      if (el && el.elementId === id) {
```
mykola-mokhnach, line 591: could it be `el?.elementId` ?

jlipps: i don't think we have any guarantee that the same element found by multiple invocations of `findElement` will have the same ID. the drivers do not make any such guarantee, anyway, i'm pretty sure.

jlipps (commented):
this is a great feature and code looks good (I agree with @mykola-mokhnach's suggestions). my main concern is about the logic used to detect equivalence between elements. i'm not sure it's bulletproof (see comment on the line).

app/renderer/actions/Inspector.js
eglitise: @mykola-mokhnach unfortunately we can't use optional chaining yet. The app is built with `parcel` v1, which depends on `terser` 3.x, but optional chaining support was only added in `terser` 5.2.0.

eglitise: @jlipps I agree, I've definitely seen common ID changes with iOS. But if this happens, the user will simply receive the error message. I'd say this is better than trying to match using other element properties that are less unique, which may result in selecting the incorrect node.

app/renderer/actions/Inspector.js
eglitise: Fixed

app/renderer/actions/Inspector.js
eglitise: Fixed

jlipps (approved):
OK, seems reasonable to me!

Issue #698, opened 2023-01-29
'Get timing' feature for User's own created locators.


swap12875:
## 'Get timing' feature should extended to Users locators beyond the appium suggested locators only (must have feature)?

While using appium inspector, the locators which appium is suggesting are displayed on the right 'Selected Element' panel, On the same panel it have Icon of 'Hour glass'(Get timing) by which the appium gives the time taken to find that element with particular locators strategy and selectors.
But when user search the element with their own selector and strategy, he is not able to check the time taken on the UI to find that element. So It is highly Good to have rather I say must to have feature on the inspector that **'User can check the time for their own designed/created selectors and strategies on the Inspector'**

## The problem

User is not able to check the time taken by their own preferred strategies and locators to find the elements.

## Environment

- I am running Appium Desktop version _<App Version: 2022.11.1>_.;  Electron: 13.6.9;  Node.js: 14.16.0

- I am on (pick one):
    - [ ] Mac
    
## Snapshot

![Monosnap Appium 2023-01-29 11-40-10](https://user-images.githubusercontent.com/67191980/215308511-931cfd2b-4304-4d7c-b1ab-e3a1bd4ca053.jpg)

jlipps:
Yeah, good idea!

swap12875:
Hope the feature will be included in next release!

eglitise:
This feature is included in version 2023.5.2.

PR #848, opened 2023-05-29
feat: show search time in locator search results

eglitise:
This PR implements #698.
When searching for an element, the results screen will now also show the elapsed time (arrow not included):

locator-search-timing

### DIFF
--- a/app/renderer/actions/Inspector.js
+++ b/app/renderer/actions/Inspector.js
@@ -436,11 +436,11 @@ export function searchForElement (strategy, selector) {
     dispatch({type: SEARCHING_FOR_ELEMENTS});
     try {
       const callAction = callClientMethod({strategy, selector, fetchArray: true});
-      let {elements, variableName} = await callAction(dispatch, getState);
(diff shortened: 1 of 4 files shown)


mykola-mokhnach:
The PR itself is ok, although the value itself might be confusing, as it does not measure the actual lookup performance, but also includes the time needed to transfer the data over the wire to the server and back.
Perhaps, it makes sense to add an explanation hint there?

eglitise:
Hmm, I'm not sure if it is relevant for the user. From my point of view, the main goal in showing the search time (both here and in the selected element panel) is to help the user figure out the fastest strategy, rather than accurately calculate absolute values (which will change with each search anyway). Using this locator in a test run will still add the data transfer overhead time.

But actually, I simply copied the current approach used in the selected element panel :)

mykola-mokhnach (approved):

jlipps:
Yeah I think round trip time is OK since that is often what users are interested in anyway. LGTM!

PR #849, opened 2023-05-29
fix: do not resend highlight for selected element

eglitise:
This fixes the issue where, when searching for elements and selecting any result from the list, if the same result was then selected again, the highlight calculation would be requested again.

### DIFF
--- a/app/renderer/components/Inspector/LocatedElements.js
+++ b/app/renderer/components/Inspector/LocatedElements.js
@@ -44,8 +44,8 @@ class LocatedElements extends Component {
             dataSource={locatedElements}
             renderItem={(elementId) =>
               <List.Item type='text'
-                className={locatorTestElement === elementId ? InspectorStyles.searchResultsSelectedItem : ''}
-                onClick={() => setLocatorTestElement(elementId)}
+                {...locatorTestElement === elementId && { className: InspectorStyles.searchResultsSelectedItem }}
+                {...locatorTestElement !== elementId && { onClick: () => {setLocatorTestElement(elementId);} }}
               >
                 {elementId}
               </List.Item>


app/renderer/components/Inspector/LocatedElements.js
```diff
@@ -44,8 +44,8 @@ class LocatedElements extends Component {
             dataSource={locatedElements}
             renderItem={(elementId) =>
               <List.Item type='text'
-                className={locatorTestElement === elementId ? InspectorStyles.searchResultsSelectedItem : ''}
-                onClick={() => setLocatorTestElement(elementId)}
+                {...locatorTestElement === elementId && { className: InspectorStyles.searchResultsSelectedItem }}
```
mykola-mokhnach, line 47: I am not quite sure this is the result we want, let say if `locatorTestElement !== elementId` then the result of the first expression would be `...false`, which does not make sense. Same applied to the second expression

mykola-mokhnach: I also prefer to use explicit parentheses to such expression as it is not obvious to which element the spread operator is applied, e.g. `...(expr)` vs `...expr`

eglitise: I'll admit, I'm not too familiar with the spread operator, so I was basing the code off [this SO answer](https://stackoverflow.com/a/54039314). Testing showed it was working fine, but if there's a better way, I'd be more than happy to change it.
I definitely agree on the parentheses though, I somehow missed those.

mykola-mokhnach: it probably works because the spread operator implementation in js is more forgiving in comparison to other languages.

I would try to rewrite this like `...(locatorTestElement === elementId ? { className: InspectorStyles.searchResultsSelectedItem } : {})`

PR #857, opened 2023-05-30
style: align selected element actions design to search results

eglitise:
This PR aligns the design of selected element actions to that of the locator search results:
* Replace send keys modal with inline input field
  * The field is flexible and will adjust to the width of the whole inspector window
* Adjust button grouping
* Change icon of clear button

selected-elem-actions-min-width
selected-elem-actions-max-width


mykola-mokhnach (approved):

jlipps (commented):
👍

PR #864, opened 2023-06-01
chore: refactor redux store

eglitise:
The Redux store was previously split into 2 files, with most of their functionality shared, plus a combiner file. I've now combined these in one file and added some additional refactor. `redux-logger` has also been moved to dev dependencies.
Note that hot module replacement isn't wrapped in a dev environment check, because Parcel automatically disables it in production (and there's already an unconditional instance of HMR in `app/reducers/index.js`)


jlipps (commented):
OK makes sense

PR #866, opened 2023-06-02
chore(deps): replace connected-react-router with redux-first-history

eglitise:
This PR replaces the abandoned `connected-react-router` with `redux-first-history`. While `redux-first-history` also doesn't appear to be maintained, its main benefit is removing blockers for updating dependencies like `history`, `react`, `react-redux` and `react-router`.

I tested both desktop app and browser versions, and both function as expected. Admittedly I'm not really sure how to test the history functionality, but I feel like it does not matter here that much - the routes change correctly, so I assume the package is working properly.


jlipps (approved):
yeah I don't think we use history so the routing stuff working properly is all we really need

PR #874, opened 2023-06-06
chore: switch to React 18 root API

eglitise:
This enables the new [React 18 root API](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-client-rendering-apis), removing the warning that always appears on launch.
However, this does add a new warning when hot module replacement is activated. It can be removed by migrating to React Fast Refresh, but that in turn requires rewriting the app to use React function components, as well as Parcel 2.

### DIFF
--- a/app/renderer/index.js
+++ b/app/renderer/index.js
@@ -1,26 +1,26 @@
 import React from 'react';
-import { render } from 'react-dom';
+import { createRoot } from 'react-dom/client';
 import Root from './containers/Root';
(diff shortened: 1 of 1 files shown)


mykola-mokhnach (approved):

jlipps (commented):
yeah sounds good. updating to function components and parcel 2 is a bigger project for sure. i suppose we could mix/match with function components, right? like upgrade them one PR at a time while keeping the others as classes?

eglitise:
I hope so 😅 will try to investigate that in the near future.

PR #885, opened 2023-06-08
feat: add infobox for disabling package name autocompletion of id locators

eglitise:
This PR is a highly specific change, but bear with me.

Background: Up until a week ago I was unaware of the `appium:disableIdLocatorAutocompletion` capability. I had been testing an Android app that had Jetpack Compose elements with `testTag`s, and I was wondering why they were listed as `resource-id`s, yet I could not find them through the element search. I didn't delve too deep into this, and just settled on using simple XPaths. Fast forward to this week, and the aforementioned capability has now unlocked the ability to find these elements by id.

To avoid others getting into the above situation, I am suggesting to add an infobox that will appear under very specific circumstances:
* `UiAutomator2` driver is used
* Element search is done using the `id` locator strategy
* The selector used in element search does not have a package name prefix
* `appium:disableIdLocatorAutocompletion` is set to `false` (the default value)
* No results are found

If all of the above points are true, the user will be greeted by this comprehensive message:

image

Still, this isn't perfect. It may be a bit misleading in the case where the user attempts to search for a visible element that does have a prefix, but is not part of the current application package (e.g. elements with the `android:id/` prefix). But, I would hope that any element with a package prefix would also have the prefix included in the search...

### DIFF
--- a/app/renderer/components/Inspector/LocatedElements.js
+++ b/app/renderer/components/Inspector/LocatedElements.js
@@ -1,10 +1,20 @@
 import React, { Component } from 'react';
-import { Input, Row, Button, Badge, List, Space, Spin, Tooltip } from 'antd';
+import { Alert, Input, Row, Button, Badge, List, Space, Spin, Tooltip } from 'antd';
 import { AimOutlined, ClearOutlined, SendOutlined, MenuUnfoldOutlined } from '@ant-design/icons';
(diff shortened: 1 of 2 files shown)


jlipps:
I've run into this confusing scenario as well, so I think it's a worthwhile change.

PR #889, opened 2023-06-09
feat: fix element search error + add infobox if automationName is missing

eglitise:
This PR should only affect Appium 1.x users.
It fixes an application error if element search is opened in a session with `appium:automationName` not specified. Only the default locator strategies will be shown in that case.

In order to help users unhide any platform-specific capabilities, I have added a new infobox that only appears if `appium:automationName` is missing:
image


eglitise:
Hmm, I feel like a repeated warning like that might be too much. I start and stop sessions more than a dozen times per day, and a warning like that would become annoying very fast. At minimum it should be permanently dismissable. But I'm open to reconsidering this once Appium 2 is out of beta.

PR #903, opened 2023-06-13
fix: use the term 'action' only in recorder

eglitise:
Currently, there are several locations that use the term 'action', while referring to completely different things:
* interactions detected when recording is turned on
* custom gestures
* Appium client commands
* OS-level controls for iOS or Android devices

This PR aims to clear up this terminology:
* The term 'action' in UI should now be mainly used in the recorder functionality
* Custom gestures use the term 'gesture'
* Appium client commands use the term 'command'
* OS-level controls use the term 'controls'


jlipps (commented):
I agree with this for the most part. My only concern is that I wanted to align with the W3C language around "Actions" (which means gestures). I want users to know that Appium Actions are the same as W3C Actions. Maybe we could call them "Touch Actions" to be more clear?

But 💯 to all the other nomenclature clarifications.

eglitise:
Touch Actions sounds too similar to the deprecated TouchActions API. I think most people would understand the term 'gestures' better than any equivalent, so I would like to keep it as is. At the same time I agree that linking it to W3C Actions is also important, so I have added a small description that references this:

image

jlipps (approved):
OK, this makes sense to me!

PR #904, opened 2023-06-13
feat: reorganise window header buttons

eglitise:
This PR mainly changes two things:
1) Removes the back button from the header buttons. It was calling `driver.back()`, which is meant to be used only in web contexts, not for general use. The command is still accessible through the Commands tab.
2) Moves the device control buttons to the header. This is primarily because the scaling was broken at low window heights, causing these buttons to go offscreen. I think this positioning also makes more sense, since these buttons are not used for interacting with the screenshot, unlike the neighboring element handle toggle and select/swipe/tap buttons.

The new header and screenshot buttons now looks like this. Note that the Android controls are of course not present on non-Android devices.
image


jlipps (approved):
yes, this is an improvement!

eglitise:
I suppose I'll change it to `feat` 😅

Issue #549, opened 2022-08-08
Appium session button is not visible in landscape mode for linux 2022


BaseerAzhar:
## The problem
Hey,
In the appium inspector the start session is not visible in landscape mode..but it always visible in portrait mode... check in different screen size as well

## Environment

* Appium version: [v2022.7.1](https://github.com/appium/appium-inspector/releases/tag/v2022.7.1)
* Desktop OS/version used to run Appium: Lixux 22, with appium inspector image..

## Details
![image](https://user-images.githubusercontent.com/32225656/183476825-b9a44beb-dbb4-4265-8a65-db095a622055.png)

jlipps:
increase your screen resolution or use the web version at https://inspector.appiumpro.com

BaseerAzhar:
hey my screen resolution is completely full and already check on different
screens as well but not working

jlipps:
what is your screen resolution?

shubham2211b:
Still same issue facing with my system not able get Start button

shubham2211b:
Any fix for the same

eglitise:
Please try with version 2023.6.2, it improve support for smaller screens.

mtahirmunir:
Hopefully, the article "[The Appium Inspector Start Session Button Not Displaying](http://blog-box.com/the-appium-inspector-start-session-button-not-displaying/)" helps you resolve the issues with the Appium Inspector's Start Session Button not displaying.

PR #910, opened 2023-06-15
fix: adjust window dimensions and appearance at min window size

eglitise:
This PR adjusts the default and minimum dimensions for the main window and several components.
The original goal was to try and reduce these dimensions in order to assist with issues like #549, but testing on Windows showed that the minimum height actually needs to be slightly _increased_ to adjust for the title bar (which is likely to affect Linux the same way). The minimum width could still be decreased, and as a result, the new minimum dimensions have been changed to `890 x 710`.

Other notable adjustments:
* Set default window height to match minimum height (may further help with the issue linked above)
* Adjust minimum sizes for a few components (app body, capabilities table, inspector) to fix their appearance at minimum window size
* Slightly increase available width of elements in 'Advanced Settings' to improve appearance at minimum window size

### DIFF
--- a/app/main/windows.js
+++ b/app/main/windows.js
@@ -20,10 +20,10 @@ function buildSplashWindow () {
 function buildSessionWindow () {
   const window = new BrowserWindow({
     show: false,
-    width: 1280,
-    height: 800,
-    minWidth: 1000,
-    minHeight: 700,
+    width: 1100,
+    height: 700,
+    minWidth: 870,
+    minHeight: 640,
     titleBarStyle: 'hiddenInset',
     webPreferences: {
       nodeIntegration: true,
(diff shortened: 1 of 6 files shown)


app/main/windows.js
```diff
@@ -20,10 +20,10 @@ function buildSplashWindow () {
 function buildSessionWindow () {
   const window = new BrowserWindow({
     show: false,
-    width: 1280,
-    height: 800,
-    minWidth: 1000,
-    minHeight: 700,
+    width: 1100,
```
mykola-mokhnach, line 23: Can we somewhere document these limitations?

mykola-mokhnach (approved):

KazuCocoa (approved):

app/main/windows.js
eglitise: Added a new subsection in the readme

jlipps:
great, this has been bugging me!

Issue #879, opened 2023-06-07
chore: migrate codebase to React function components


eglitise:
Since modern React is based on function components, it makes sense to eventually migrate to their use.
Not only that, but the latest versions of several dependencies (like `react-router`) already assume that function components are being used, which makes their upgrade procedure in the current codebase very difficult, if not outright impossible.

This task appears to be doable incrementally, though more complex class components will take longer due to hooks (`useState`, `useRef`, `useEffect`, etc.)

PR #918, opened 2023-06-16
chore: migrate Session items to React function components

eglitise:
This PR is a partial implementation of #879, covering all the components in the Session route.
Most of the components were relatively straightforward to migrate: only 2 needed lifecycle methods, and only 1 needed an instance variable. Though I have a hunch that the Inspector route will be more complicated...

I've also removed a few unnecessary/unused methods:
* `componentDidMount` in `CloudProviderSelector` - needlessly duplicates some calls in the equivalent `Session` method
* `handleCloseModal` in `CloudProviderSelector` - `<Modal onCancel=` already handles this case
* `removeCloudProvider` in `Session` - unused and already handled in `CloudProviderSelector`

### DIFF
--- a/app/renderer/components/Session/AdvancedServerParams.js
+++ b/app/renderer/components/Session/AdvancedServerParams.js
@@ -1,44 +1,40 @@
-import React, { Component } from 'react';
+import React from 'react';
 import { Collapse, Form, Row, Col, Checkbox, Input } from 'antd';
 import styles from './Session.css';
 
-const {Panel} = Collapse;
+const { Panel } = Collapse;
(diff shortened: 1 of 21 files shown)


jlipps (approved):
nice! assuming you've done some decent testing to load all these components, code looks good!

eglitise:
Yes, all of the components were confirmed to load. I also checked the behavior of the React hooks for the lifecycle methods and instance variable, and they worked too. Finally, I checked the core behavior of launching a session. Didn't go into too much in-depth testing other than that.

PR #928, opened 2023-06-20
fix: resolve issue with saving gestures

eglitise:
This fixes a problem where it was not possible to save new gestures. The fix implementation is taken from the code for saving new sessions.
Not actually sure when this issue was introduced, since I already had several saved gestures from an older version.


jlipps (commented):
👍

PR #938, opened 2023-06-25
fix: resolve issue with loading selected element

eglitise:
When selecting an element via the highlighted rectangle overlaid over the screenshot, oftentimes there may be an issue where the selected element actions load endlessly, meaning, the element search did not finish.
This was caused by a state mutation of the expanded paths array. By copying this array, the issue is resolved.

### DIFF
--- a/app/renderer/actions/Inspector.js
+++ b/app/renderer/actions/Inspector.js
@@ -140,12 +140,12 @@ export function selectElement (path) {
     const { selectedElement, sourceXML, expandedPaths } = getState().inspector;
 
     // Expand all of this element's ancestors so that it's visible in the source tree
-    const copiedExpandedPaths = expandedPaths.slice();
+    const copiedExpandedPaths = [...expandedPaths];
     let pathArr = path.split('.').slice(0, path.length - 1);
     while (pathArr.length > 1) {
       pathArr.splice(pathArr.length - 1);
       let path = pathArr.join('.');
-      if (copiedExpandedPaths.indexOf(path) < 0) {
+      if (!copiedExpandedPaths.includes(path)) {
         copiedExpandedPaths.push(path);
       }
     }
--- a/app/renderer/actions/Inspector.js
+++ b/app/renderer/actions/Inspector.js
@@ -137,26 +137,24 @@ export function selectElement (path) {
   return async (dispatch, getState) => {
     // Set the selected element in the source tree
     dispatch({type: SELECT_ELEMENT, path});
-    const state = getState().inspector;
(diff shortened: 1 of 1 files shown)


app/renderer/actions/Inspector.js
```diff
@@ -137,26 +137,24 @@ export function selectElement (path) {
...
     // Set the selected element in the source tree
     dispatch({type: SELECT_ELEMENT, path});
-    const state = getState().inspector;
-    const {attributes: selectedElementAttributes, xpath: selectedElementXPath} = state.selectedElement;
-    const {sourceXML} = state;
+    const { selectedElement, sourceXML, expandedPaths } = getState().inspector;
 
     // Expand all of this element's ancestors so that it's visible in the source tree
-    let {expandedPaths} = getState().inspector;
+    const copiedExpandedPaths = expandedPaths.slice();
```
mykola-mokhnach, line 143: I usually use [...arr] to copy an array

app/renderer/actions/Inspector.js
```diff
@@ -137,26 +137,24 @@ export function selectElement (path) {
...
     let pathArr = path.split('.').slice(0, path.length - 1);
     while (pathArr.length > 1) {
       pathArr.splice(pathArr.length - 1);
       let path = pathArr.join('.');
-      if (expandedPaths.indexOf(path) < 0) {
-        expandedPaths.push(path);
+      if (copiedExpandedPaths.indexOf(path) < 0) {
```
mykola-mokhnach, line 148: could also be (!copiedExpandedPaths.includes(path))

app/renderer/actions/Inspector.js
mykola-mokhnach: it also makes sense to add a comment about why it is necessary to copy the array

mykola-mokhnach (approved):

app/renderer/actions/Inspector.js
eglitise: I didn't really check this condition, but I agree that using not include makes it a lot clearer.

app/renderer/actions/Inspector.js
mykola-mokhnach: Perhaps eventually it would make sense to start using https://github.com/immutable-js/immutable-js in this project to avoid similar issues

mykola-mokhnach (approved):

app/renderer/actions/Inspector.js
eglitise: Yeah, agreed. I believe React recommends using Immer for this purpose: https://immerjs.github.io/immer/

PR #939, opened 2023-06-26
fix: resolve issue where gestures may get deleted

eglitise:
This resolves a critical issue where newly saved gestures may get deleted upon relaunching the app, due to a state mutation. (I don't think this would have deleted gestures saved before v2023.6.2?)
The fix implementation is once again taken from the code for saving sessions.


eglitise:
And now suddenly I can't reproduce the issue anymore 🤔 well, this implementation seems safer anyway, and there's more than one issue still present in the gesture editor...

PR #941, opened 2023-06-27
chore: migrate most Inspector items to React function components

eglitise:
This PR continues the implementation of #879, covering almost all the components in the Session route. The two exceptions being `GestureEditor` and `Inspector`, which are the most complex components, so I thought it best to leave them for a separate PR.

As I expected, this was overall more complicated than Session, with more uses of React hooks. Like with Session, a few cleanups were also done here, mainly in `ElementLocator`, `LocatedElements`, and `HighlighterRects`.

Each individual component was tested to load and execute its main functionality.


jlipps (approved):
👏

PR #944, opened 2023-06-29
fix: hide ugly scrollbars in Firefox

eglitise:
I opened the browser build in Firefox and was wondering why I was seeing scrollbars all over the place. Realised that `-webkit-scrollbar` doesn't apply to non-WebKit environments - who would have thought.
So here I've added a fix that _is only supported_ in Firefox. ([at least for now](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width#browser_compatibility))

### DIFF
--- a/app/renderer/components/Inspector/Inspector.css
+++ b/app/renderer/components/Inspector/Inspector.css
@@ -98,6 +98,7 @@
 .inspector-main :global(.ant-card-body) {
     height: calc(100% - 48px);
     padding: 12px;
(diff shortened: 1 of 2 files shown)


jlipps:
Ah yes, good point. We used all the webkit only hacks since we were only targeting electron/chrome. I guess now we should care a bit more about cross browser compatibility!

eglitise:
Thankfully, this was really the only problem I noticed with Firefox functionality, and it's good that it was a simple fix.

PR #945, opened 2023-06-29
feat: enhancements + internal improvements for gesture editor

eglitise:
This PR is primarily focused on internal changes for the gesture editor functionality:
* Migrate to React function components (part of #879)
* Make use of `_.cloneDeep` to prevent state mutations

However, some user-facing changes are also included:
* Add support for translations
* Disable Save button for a new gesture (only Save As is enabled, which when pressed will also return to gestures list)
* Fix undefined interaction mode after returning from editor
* Show icon even for ticks with undefined actions (otherwise it's unclear why adding a new pointer causes the tabs to shift downwards)
* Remove unneeded bottom padding (also affects commands and session information tab)


mykola-mokhnach (approved):

jlipps (approved):
nice!

PR #958, opened 2023-07-04
fix: group gesture action buttons + add missing key for drawn gesture points

eglitise:
Two small fixes for the saved gesture usage scenario:
* Use `Button.Group` for the saved gesture actions buttons (looks nicer)
* Add missing key for the drawn gesture points (otherwise React throws a warning)


jlipps:
👍

jlipps:
👍

PR #963, opened 2023-07-05
chore: migrate remaining Inspector items to React function components

### LINKED ISSUE #879: chore: migrate codebase to React function components

eglitise:
This PR migrates the remaining few Inspector items to React function components, and therefore resolves #879! 🎉 
Only 3 files were left: `Inspector`, `Root` and `App` (there is also `ErrorBoundary` but it is not supported as a function component)
* `App` appears to have been used only for the translation injection, but as this is already being done in `InspectorPage` and `SessionPage` directly, the file was found to be obsolete.
* `Inspector` was tricky due to the auto-rescaling part. It relied on both updating state values and immediately using them inside `componentDidMount`, which the `useEffect` hook does not allow. The problem was solved by using a second `useEffect` hook. Testing was done on Mac, Windows and browser, in both regular and MJPEG mode, to confirm that auto-scaling works as expected.


mykola-mokhnach (approved):

app/renderer/components/Inspector/Inspector.js
```diff
@@ -147,103 +107,110 @@ export default class Inspector extends Component {
...
+    return () => {
+      if (windowSize) {
+        window.removeEventListener('resize', updateSourceTreeWidthDebounced);
+        if (mjpegStreamCheckInterval.current) {
+          clearInterval(mjpegStreamCheckInterval.current);
+          mjpegStreamCheckInterval.current = null;
+        }
+      }
+    };
+  }, [windowSize && windowSize.width]);
```
jlipps, line 166: I'm a little confused by this. Isn't this where we're supposed to pass the props that this effect depends on? But `windowSize && windowSize.width` is a boolean, not a prop?

I guess I can see how it works, but I wonder if the code might be a little less arcane if the dependency was just `windowSize` and the boolean check was put into the conditional inside the effect instead.

It's up to you, this just caught my eye.

jlipps (approved):
amazing, what a huge overall contribution to the code maintainability of this project!

app/renderer/components/Inspector/Inspector.js
eglitise: I did originally use just `windowSize`, but the issue with that was that the hook was fired every time `windowSize` was updated (even if the property values did not change), because it is an object and not a primitive type. Ideally this dependency would be just on `windowSize.width`, but the initial state of `windowSize` is undefined. So, I researched a better approach and found this, which also works.
However, I do agree that it's not very obvious. I searched again, and another approach for relatively simple objects (like in this case) is to just wrap the object in `JSON.stringify`. It does appears to work in my tests, so I will change this.

jlipps: ok, that all makes sense!

PR #968, opened 2023-07-06
chore(deps): migrate to `react-router` internal history

### LINKED ISSUE #882: chore: migrate to `react-router`'s internal history
Currently, transitions between the session and inspector paths depend on the history coming from `redux-first-history`.
But otherwise... history is not all that relevant. The goal of this task is to migrate to the `react-router` history component (which seems to only be achievable with hooks). This would allow removing the `history` and `redux-first-history` dependencies.

eglitise:
This PR migrates the route change responsibility from `redux-first-history` to `react-router-dom`, and therefore resolves #882. This is achieved with `react-router`'s `useHistory` hook and `MemoryRouter`.
As a result, `redux-first-history` and `history` have been removed as direct dependencies.

~⚠️ Note, I have also removed the`initFromQueryString` method. I am not really sure how it could have been used, probably only in the browser version by appending additional parameters after the main URL? Even if so, since the app is now based on `MemoryRouter`, the URL itself is not changed anymore.~

Route changes have been tested in both Mac app and browser version.


jlipps:
I haven't fully reviewed this yet, but we do need to keep the init from query string functionality. It's what enables launching the inspector into a session via a url (and yes, this is only for the browser version). But it's depended on by some downstream consumers of the inspector so we shouldn't remove it. Is there a way to preserve the functionality with the new type of router?

eglitise:
Okay, that's fine, I don't think it would be too difficult to add it back and modify the routing call.

eglitise:
@jlipps I've now tested the functionality of `initFromQueryString` with this `MemoryRouter` and it works as expected, populating the session details and autolaunching it, if specified. The only difference is that the route details are not added at the end of the URL anymore (`#/`, `#/inspector`, `#/session`). Is that acceptable?

jlipps:
Yeah, I should build something into the inspector that enables the init from query string. I think it's fine that details aren't included in the URL so long as the navigation works! I'll give this a full review now.

jlipps (approved):

PR #988, opened 2023-07-18
fix(deps): migrate to `electron-extension-installer`

eglitise:
This PR replaces the abandoned `electron-devtools-installer` with `electron-extension-installer`.
Worth noting that with `electron-devtools-installer`, only the Redux devtools were working, while the React devtools loaded, but did not work. Now with `electron-extension-installer`, this issue is fixed, and React devtools are also working.


jlipps:
🎉

PR #989, opened 2023-07-19
chore: tidy up highlighter creation

eglitise:
This PR includes a few fixes and adjustments for the creation of highlighter rectangles over the app screenshot:
* split highlighter for elements and highlighter for bounds into separate React components
  * the functionality of these highlighters was different enough to warrant separation
* remove unused parameter for `unselectHoveredElement`
* most importantly - only render highlighters for elements with non-zero height and width

The main goal of the last change was to improve highlighter load time performance specifically for webview contexts. Unfortunately I did not see a notable difference, but the amount of highlighter elements in the Inspector's DOM tree is greatly reduced, which may still help.
For example, in my tests, a Google search results page on Android Chrome originally generated 1597 highlighter divs, but after this fix, the count dropped to 701 divs - a 56% reduction! Other pages in both Android Chrome and iOS Safari showed similar improvements anywhere between 40 and 65%.


jlipps (approved):
great idea!

KazuCocoa (approved):

PR #995, opened 2023-07-21
docs: update image of inspector window

eglitise:
The screenshot of the inspector window that is included in the README has never been updated. Since several things in the GUI have changed since that time, I thought it's a good idea to update this screenshot. The selected element panel is now also populated.
The app used in the screenshot is the IntegrationApp included in `appium-xcuitest-driver`.


KazuCocoa (approved):
Nice!

PR #1002, opened 2023-07-26
fix: improve screen size calculation for Android landscape mode & when changing orientation

### LINKED ISSUE #665: If the device orientation is changed mid-session, the generated screenshot and skeleton are not aligned
## Appium or Appium Desktop?

## The problem

When the device under test changes from portrait to landscape orientation (or vice versa) mid-session, the generated screenshot and skeleton are not aligned on top of each other:

Screenshot 2022-11-21 at 15 07 13
You can see that the screenshot is correct, but the skeleton is too large, and is only partially accessible, with the rest seemingly continuing into the app source panel.

Similarly, if we launch the session in landscape, then change the device to portrait, the skeleton is far too small:

Screenshot 2022-11-21 at 15 37 22

To me it seems like the skeleton generation does not recalculate the device width, and keeps the same value as it was upon launching the session.

This issue is not new and has been present since Appium Desktop.

Also, I am running Appium 1.22.3. Perhaps this issue depends on the different drivers and is fixed in newer versions?

## Environment

- I am running Appium Desktop version _2022.8.1_.
- I am on (pick one):
    - [x] Mac
    - [ ] Windows
    - [ ] Linux

### LINKED ISSUE #701: Tap by Coordinates does not work for Android apps in landscape
## The problem

If I start a session in portrait mode, and then switch to landscape, tap by coordinates no longer works. It appears that the screen size is incorrect, and so all taps are offset. (I have tested this by finding the right button coordinates, and clicking those instead.) In other words:
1. Put phone in portrait mode
2. Start Appium session
3. Put phone in landscape mode
4. Try to click things

image

To verify this, I've tried a couple of things:

1. I started Appium making sure the phone was in landscape mode already, which does work as expected.
2. The screen capture above is my phone in landscape mode, where I'm hovering over the Facebook icon at 248x272. My phone resolution is actually 1080x2220, so that maps to 509x559 (I think, used aspect ratio). If I find and click on those coordinates instead, it works.  The actual image inside the inspector is about 1080x529. 
 
My guess is that the source of the problem is that old width remains, and height is recalculated based on that. So, X and Y coordinates will always be too small.

I don't *think* this is an Appium problem, because I can see the right coordinates (those generated by Appium Inspector) in the Appium Server logs. I think the problem lies somewhere in how Appium Inspector recalculates the clickable area of the phone after an orientation change.

## Environment

- I am running Appium Desktop version 2022.11.1 (Appium Inspector).
- I am on (pick one):
    - [x] Mac
    - [x] Windows
    - [ ] Linux

eglitise:
This PR improves the screen size calculation specifically when using UiAutomator2, either in landscape mode, or when changing device orientation in any way. It also fixes #665 and fixes #701.
The core issue was caused by the window width and height being effectively hardcoded to portrait mode.

The fix reuses the existing `driver.getWindowRect` call to check the current dimensions (which may be smaller than the full screen size, due to UiAutomator2 calculations that I will discuss in a separate issue). The dimensions are then used only to determine the screen orientation, and the actual values are still taken from the full screen size, as before. This means that portrait mode behavior is entirely unchanged.

Due to the aforementioned UiAutomator2 calculations, as well as some webview adjustments that I need to look into, the highlighter positions are still not perfect. But I believe that this change already improves the situation.
Here is a comparison for the rarest case, webview in landscape. This is how it looked before this fix:
webview-landscape-before
In the app source, I have selected the link element for the Google account button. We can see that in the screenshot, the highlighter is not visible, as it has been cut off. What's more, changing the inspector window size here would further break these positions.
And this is how it looks after this fix:
webview-landscape-after
The widths and X-positions for all highlighters have been fixed, and the requested highlighter is not longer cut off. Changing the inspector window size also does not break this. Still, you can still see that there is an unwanted Y-axis offset, which will be addressed in the future.

I also noticed a code comment that `driver.getWindowRect` was previously not called here due to Chrome not supporting this command - it was apparently added in Chrome 65 (released March 2018), so it should be safe to cut off support for any versions older than that.

### DIFF
--- a/app/renderer/lib/appium-client.js
+++ b/app/renderer/lib/appium-client.js
@@ -201,14 +201,18 @@ export default class AppiumClient {
     let windowSize, windowSizeError;
     const {client: {capabilities: {deviceScreenSize, platformName, automationName}}} = this.driver;
     try {
-      // The call doesn't need to be made for Android for two reasons
-      // - when appMode is hybrid Chrome driver doesn't know this command
-      // - the data is already on the driver
+      windowSize = await this.driver.getWindowRect();
       if (_.toLower(platformName) === 'android' && _.toLower(automationName) === 'uiautomator2') {
+        // returned Android height and width can both be affected by UiAutomator2 calculations
+        // we stick with device dimensions, but swap them depending on detected orientation
         const [width, height] = deviceScreenSize.split('x');
-        windowSize = {width, height, x: 0, y: 0};
-      } else {
-        windowSize = await this.driver.getWindowRect();
+        if (windowSize.height > windowSize.width) { // portrait mode
+          windowSize.height = height;
+          windowSize.width = width;
+        } else { // landscape mode
+          windowSize.height = width;
+          windowSize.width = height;
+        }
       }
     } catch (e) {
       windowSizeError = e;


KazuCocoa (approved):
👍

PR #1006, opened 2023-07-28
feat: enhance hybrid mode controls

### LINKED ISSUE #899: feature request: inform user if no webview contexts were found after selecting hybrid mode

### Current Behavior

When you switch your app from native view to hybrid view, it will try that 20 times.
After 20 times, it will default back to the "NATIVE_APP" behaviour again.

You can see it in appium's log like this:

```
debug] [RemoteDebugger] Sending to Web Inspector took 2ms
[debug] [RemoteDebugger] Empty page dictionary received
[debug] [RemoteDebugger] Error checking application: 'Empty page dictionary received'. Retrying connection
[RemoteDebugger] Could not connect to a valid app after 20 tries.
[debug] [XCUITestDriver@782f (b66a2d6b)] No available web pages: Could not connect to a valid app after 20 tries.
[debug] [XCUITestDriver@782f (b66a2d6b)] No web frames found.
[debug] [XCUITestDriver@782f (b66a2d6b)] No webviews found in 9700ms
[debug] [XCUITestDriver@782f (b66a2d6b)] Responding to client with driver.execute() result: [{"id":"NATIVE_APP"}]
[HTTP] <-- POST /session/b66a2d6b-96c5-423c-9280-e64c6d311d79/execute/sync 200 9702 ms - 31
``` 

However, the icon in inspector is not changed back to "NATIVE_APP" and no indicator of this is shown to the user:

image

### Expected Behavior

I expect the icon 
image
to be going back to the "NATIVE_APP" and show a error message (like a popup?) to the user of this issue.

### Operating System

Mac

### Appium Inspector Version

2023.5.2

### Appium Version

2.0.0-beta.71

### Further Information

_No response_

eglitise:
This PR enhances the UX for the Inspector's hybrid mode with two main changes:
* Move the context selector to the application header. It did not make sense to have it hidden under a selected element. All the old info alert texts have been combined into one info icon+tooltip.
* Add an icon+tooltip when hybrid mode is selected but only the native context is found. Resolves #899.

https://github.com/appium/appium-inspector/assets/37242620/f8a25be6-eb43-47bc-abca-1599a6edc7ce

One small caveat is that the documentation link for hybrid mode now points to the 1.x docs in the GitHub repo, since this info is not yet available in the new docs. It's still an improvement over the old link, which leads to a 404.


app/renderer/components/Inspector/HeaderButtons.js
```diff
@@ -15,13 +16,17 @@ import {
   SearchOutlined,
   CloseOutlined,
   AppstoreOutlined,
-  GlobalOutlined
+  GlobalOutlined,
+  InfoCircleOutlined,
+  ExclamationCircleOutlined
 } from '@ant-design/icons';
 
+const HYBRID_MODE_DOCS_URL = 'https://github.com/appium/appium/blob/1.x/docs/en/writing-running-appium/web/hybrid.md';
```
mykola-mokhnach, line 24: @jlipps Might this doc be outdated?

assets/locales/en/translation.json
```diff
@@ -141,9 +141,8 @@
   "Get Timing": "Get Timing",
   "interactionsNotAvailable": "Interactions for this element may not be available",
   "usingXPathNotRecommended": "Using XPath locators is not recommended and can lead to fragile tests. Ask your development team to provide unique accessibility locators instead!",
-  "usingSwitchContextRecommended": "Webview context(s) detected; certain elements might only be accessible after switching to a webview context, which must be done in an Appium script. See http://appium.io/docs/en/writing-running-appium/web/hybrid/ for more information",
-  "usingWebviewContext": "Using the Webview inspector in Appium Inspector is less accurate in retrieving and selecting DOM elements in comparison to using the DevTools of Chrome and Safari.",
-  "contextSwitcher": "You can switch to a different context here. See http://appium.io/docs/en/writing-running-appium/web/hybrid/ for more information",
+  "noAdditionalContextsFound": "Appium has detected only the native context",
```
mykola-mokhnach, line 144: Just wanted to put my 2 cents there: It would probably be too generic or not even correct to say "Appium has detected" as Appium is just an abstract server. Webview detection is done by the corresponding driver, for example UIA2 or XCUITest one.

mykola-mokhnach (approved):

app/renderer/components/Inspector/HeaderButtons.js
KazuCocoa: If we need the link, maybe https://appium.github.io/appium.io/docs/en/writing-running-appium/web/hybrid/ will be better. but afaik the documentation is outdated. (ideally, each driver that supports hybrid need something)

KazuCocoa (approved):

app/renderer/components/Inspector/HeaderButtons.js
eglitise: I've now changed the link to the appium.github.io version, I agree that it looks better.

assets/locales/en/translation.json
eglitise: Fixed. I decided that instead of explicitly mentioning 'Appium driver' or something similar, it would still be fine to use even more generic wording 😄

app/renderer/components/Inspector/HeaderButtons.js
jlipps: we could link to https://appium.io/docs/en/2.0/guides/context/, and then link from this page to driver-specific context docs as necessary

jlipps (commented):
👍 this is a huge improvement!

app/renderer/components/Inspector/HeaderButtons.js
eglitise: I'm more than happy to update the docs links (including others elsewhere in the code), but I would prefer to keep the current links until the new docs are updated with more information. Right now the context page content is rather abstract and could use more concrete information on webview contexts (and links to driver-specific documentation as well). I would also suggest having (Appium-maintained) client examples like in the old docs.

PR #1014, opened 2023-08-01
fix: improve highlighter positions in webview

eglitise:
This PR improves calculations for generating element highlighters when inspecting a webview. Both iOS and Android had issues in such situations.
Note that highlighter generation in native context is not affected.

### Android
Here the problem was fairly simple - the top offset from the Chrome address bar was being ignored:
android-old-landscape
The issue was fixed by finding the actual webview element (`android.webkit.WebView`), which had correct dimensions, and using its X and Y coordinates as the starting points:
android-new-landscape
This also allowed to fix a problem in landscape mode, where the X axis would also need an offset if the device had a hole punch camera or a similar hardware configuration, resulting in the webpage not extending to that area.

The solution was also confirmed to work when the Chrome address bar was scrolled out of view.

### iOS
At first glance, the highlighter positions were correct... until the phone was turned to landscape mode, or the user set the address bar at the top:
ios-old-landscape
ios-old-portrait
The X and Y offsets each needed their own fix. For the Y offset, the implementation was similar to Android, but since there was no ideal webview element here, the offset was instead obtained from the top status+address bar:
ios-new-portrait
The X offset appeared only in landscape mode, at both ends of the screen, and seemed to match the status bar height, so that was exactly the value that was used:
ios-new-landscape

The solution was also confirmed to work when the top address bar in portrait orientation was scrolled out of view, as well as when the top bar in landscape orientation was scrolled _into_ view.

### DIFF
--- a/app/renderer/lib/appium-client.js
+++ b/app/renderer/lib/appium-client.js
@@ -251,7 +251,7 @@ export default class AppiumClient {
       await this.driver.switchContext(NATIVE_APP);
     }
 
-    const { statBarHeight, viewportRect } = await this.driver.getSession();
+    const sessionDetails = await this.driver.getSession();
     const isAndroid = this.driver.client.isAndroid;
 
     // Get all available contexts (or the error, if one appears)
@@ -274,7 +274,7 @@ export default class AppiumClient {
           webviewLeftOffset = x;
         } else {
           // fallback to default top offset value if element retrieval failed
-          webviewTopOffset = viewportRect.top;
+          webviewTopOffset = sessionDetails.viewportRect.top;
         }
       } else if (this.driver.client.isIOS) {
         // on iOS, find the top status bar and address bar and use its Y endpoint
@@ -285,7 +285,7 @@ export default class AppiumClient {
         }
         // in landscape mode, there is empty space on both sides (at default zoom level), so add offset for that too
         if (windowSize.height < windowSize.width) {
-          webviewLeftOffset = statBarHeight;
+          webviewLeftOffset = sessionDetails.statBarHeight;
         }
       }
 
--- a/app/renderer/lib/appium-client.js
+++ b/app/renderer/lib/appium-client.js
@@ -9,7 +9,7 @@ const { TAP, SWIPE, GESTURE } = SCREENSHOT_INTERACTION_MODE;
 const ANDROID_WEBVIEW_SELECTOR = '//android.webkit.WebView[1]';
 // Selector for the iOS status bar and Safari address bar - not always present
 const IOS_TOP_CONTROLS_SELECTOR = '**/XCUIElementTypeOther[`name CONTAINS "SafariWindow"`]' +
-  '/XCUIElementTypeOther/XCUIElementTypeOther/XCUIElementTypeOther/XCUIElementTypeOther';
+  '/XCUIElementTypeOther/XCUIElementTypeOther/XCUIElementTypeOther/XCUIElementTypeOther[1]';
 
 export const NATIVE_APP = 'NATIVE_APP';
 let _instance = null;
--- a/app/renderer/lib/appium-client.js
+++ b/app/renderer/lib/appium-client.js
@@ -251,8 +251,8 @@ export default class AppiumClient {
       await this.driver.switchContext(NATIVE_APP);
     }
 
-    const { platformName, statBarHeight, viewportRect } = await this.driver.getSession();
-    const isAndroid = _.toLower(platformName) === 'android';
+    const { statBarHeight, viewportRect } = await this.driver.getSession();
+    const isAndroid = this.driver.client.isAndroid;
 
     // Get all available contexts (or the error, if one appears)
     try {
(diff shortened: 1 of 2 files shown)


app/renderer/lib/appium-client.js
```diff
@@ -1,9 +1,15 @@
...
+import { parseSource, setHtmlElementAttributes } from './webview-helpers';
+import { SCREENSHOT_INTERACTION_MODE, APP_MODE } from '../components/Inspector/shared';
 
-const {TAP, SWIPE, GESTURE} = SCREENSHOT_INTERACTION_MODE;
+const { TAP, SWIPE, GESTURE } = SCREENSHOT_INTERACTION_MODE;
+
+// Selector for the Android webview - includes the correct top and bottom boundaries
+const ANDROID_WEBVIEW_SELECTOR = '//android.webkit.WebView[1]';
+// Selector for the iOS status bar and Safari address bar - not always present
+const IOS_TOP_CONTROLS_SELECTOR = '**/XCUIElementTypeOther[`name CONTAINS "SafariWindow"`]' +
```
mykola-mokhnach, line 11: not sure if this locator would be reliable enough

app/renderer/lib/appium-client.js
```diff
@@ -221,90 +227,79 @@ export default class AppiumClient {
...
     }
 
-    // Note: These methods need to be executed in the native context because ChromeDriver behaves differently
+    // The retrieval of all contexts and webview position adjustments require some native context use
     if (currentContext !== NATIVE_APP) {
       await this.driver.switchContext(NATIVE_APP);
     }
 
-    ({platformName, pixelRatio, statBarHeight, viewportRect} = await this.driver.getSession());
+    const { platformName, statBarHeight, viewportRect } = await this.driver.getSession();
```
mykola-mokhnach, line 254: are are looking forward to remove this information from GET /session endpoint completely as it is not a part of W3C standard. Consider using extension calls instead

app/renderer/lib/appium-client.js
```diff
@@ -1,9 +1,15 @@
...
-import {getWebviewStatusAddressBarHeight, parseSource, setHtmlElementAttributes} from './webview-helpers';
-import {SCREENSHOT_INTERACTION_MODE, APP_MODE} from '../components/Inspector/shared';
+import { parseSource, setHtmlElementAttributes } from './webview-helpers';
+import { SCREENSHOT_INTERACTION_MODE, APP_MODE } from '../components/Inspector/shared';
 
-const {TAP, SWIPE, GESTURE} = SCREENSHOT_INTERACTION_MODE;
+const { TAP, SWIPE, GESTURE } = SCREENSHOT_INTERACTION_MODE;
+
+// Selector for the Android webview - includes the correct top and bottom boundaries
+const ANDROID_WEBVIEW_SELECTOR = '//android.webkit.WebView[1]';
```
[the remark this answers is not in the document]
eglitise: I used xpath here mainly because I've seen apps that have more than one webview element (one parent, one child). Admittedly, UiA2 also does not detect any webview contexts in such a screen, but I decided to go the safe route.

app/renderer/lib/appium-client.js
eglitise: Yeah, it's definitely not great... but the child elements sadly don't have any other way to identify them, at least in portrait mode. In landscape mode, the third `XCUIElementTypeOther` has a `UnifiedBar` label, but that's about it.
I did however just realise that this selector finds multiple results, so I'll at least add an index in the end.

app/renderer/lib/appium-client.js
eglitise: Good call, I'll see what I can do.

eglitise: I assume `platformName` can still remain here, since it is part of the standard?
At the same time, according to UiA2 docs, it does not enforce the value to be `Android`, so maybe it's better to check `appium:automationName` instead. But I've never gotten Espresso to work, so I won't know how that behaves...

app/renderer/lib/appium-client.js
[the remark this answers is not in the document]
mykola-mokhnach: class name selectors are faster. Also it would always the first match - exactly what's needed here.
Btw, I am not sure about the correctness of this locator in general - it won't find the first WebView on the page, but rather each first child. Check https://stackoverflow.com/questions/52165349/what-is-the-role-of-parentheses-in-xpath-1-0 for more details.

app/renderer/lib/appium-client.js
```diff
@@ -221,90 +227,79 @@ export default class AppiumClient {
...
-          }
+      if (isAndroid) {
+        // on Android, find the root webview element and use its X and Y startpoints
+        const webview = await this.fetchElement({strategy: 'xpath', selector: ANDROID_WEBVIEW_SELECTOR});
+        if (webview.el) {
+          const { x, y } = await webview.el.getRect();
+          webviewTopOffset = y;
+          webviewLeftOffset = x;
+        }
+      } else {
```
jlipps, line 276: It's probably not a big deal right now cause I'm sure there are lots of these sprinkled in the codebase, but moving forward we should try to take care that we don't assume sessions will always be android or ios. This `else` clause could run in a Roku TV session, for example, where it would be irrelevant.

app/renderer/lib/appium-client.js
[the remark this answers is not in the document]
eglitise: Fixed

app/renderer/lib/appium-client.js
eglitise: I've now changed both `statBarHeight` and `viewportRect` to be retrieved using mobile extensions. In case the driver is too old and does not support these extensions, it will fallback to the /session endpoint.

mykola-mokhnach (approved):

PR #1016, opened 2023-08-03
fix: resolve issue with some commands not working

eglitise:
This PR fixes a problem where some commands in the Commands panel could not be executed (specifically, `rotateDevice`, `setGeoLocation`, `executeScript` and `updateSettings`).
The issue was caused by a state mutation, so the fix involved simply cloning the affected `args` array before mutating it.
Note that `_.cloneDeep` was used instead of the more performant spread syntax, since the `args` array may have multiple levels (e.g. for XCUITest driver's `mobile: pushNotification`)


jlipps (approved):
nice, good find

PR #1021, opened 2023-08-07
feat: add button to save current application screenshot

eglitise:
This PR adds a new button that allows to save the full-size version of the currently shown application screenshot:
screenshot-button

It is one of those features that initially didn't even occur to me, but once it did, I realised it could be extremely useful.
Note that the button is not shown in MJPEG mode.

### DIFF
--- a/app/renderer/components/Inspector/Inspector.js
+++ b/app/renderer/components/Inspector/Inspector.js
@@ -43,7 +43,7 @@ const downloadFile = (href, filename) => {
 
 const downloadXML = (sourceXML) => {
   const href = 'data:application/xml;charset=utf-8,' + encodeURIComponent(sourceXML);
-  const filename = 'source.xml';
+  const filename = `app-source-${new Date().toJSON()}.xml`;
   downloadFile(href, filename);
 };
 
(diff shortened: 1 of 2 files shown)


app/renderer/components/Inspector/Inspector.js
```diff
@@ -41,6 +41,18 @@ const downloadXML = (sourceXML) => {
...
 
+const downloadXML = (sourceXML) => {
+  const href = 'data:application/xml;charset=utf-8,' + encodeURIComponent(sourceXML);
+  const filename = 'source.xml';
+  downloadFile(href, filename);
+};
+
+const downloadScreenshot = (screenshot) => {
+  const href = `data:image/png;base64,${screenshot}`;
+  const filename = 'appium-inspector-screenshot.png';
```
mykola-mokhnach, line 52: maybe it would make sense to add the current timestamp to the screenshot file name

eglitise: The file names for both screenshot and XML source now include the UTC timestamp. (Conversion to local time seemed like it would be quite messy)

PR #1030, opened 2023-08-11
fix: do not open inspector screen if session creation fails

### LINKED ISSUE #1026: bug: proceeds to the page source view while the new session creation got failed

### Current Behavior

Even when a new session creation gets failed (so, an error popup message appears), the main view proceeds to the page source view, then keep sowing below the attached view.

Screenshot 2023-08-10 at 10 38 01 AMshowing

### Expected Behavior

We should go back to the capabilities etc section.

### Operating System

Mac

### Appium Inspector Version

2023.8.1

### Appium Version

any (e.g. 2 base, 1.22.3 base)

### Further Information

_No response_

eglitise:
This PR resolves #1026. The inspector screen should no longer open if session creation fails for any reason.

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -263,7 +263,7 @@ export function newSession (caps, attachSessId = null) {
             description: i18n.t('sauceCredentialsRequired'),
             duration: 4
           });
-          return;
+          return false;
         }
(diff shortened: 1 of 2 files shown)


KazuCocoa (approved):
i see, thank you

PR #1041, opened 2023-08-17
feat: improve Commands tab UX

eglitise:
This PR aims to improve the UX for the Commands tab, making it easier to execute them, as well as providing more information about specific commands, if necessary.
As a first note, I am aware that the vast majority of the currently available commands should probably be turned into `mobile:` execute methods. In the scope of this PR, I intentionally tried to keep any changes for the current commands to a minimum, in order to not break existing workflows. Instead, I plan to add separate tabs for `mobile:` execute methods and settings in future PRs.

Main changes are as follows:
* Group commands using Collapse elements instead of Select. This saves up to 3 clicks for executing a command
* Adjust command groups
  * Combine 'App' and 'Android Activity' into 'App Management'
  * Rename 'File' to 'File Transfer'
  * Rename 'Keys' to 'Keyboard'
  * Rename 'Network' to 'Connectivity'
  * Combine 'Interaction' and 'iOS Simulator' into 'Device Interaction'
  * Move Android 'Fingerprint' to 'Device Interaction'
* Remove custom labels for individual commands. IMO this makes it easier to identify them
* Scale the number of commands per row depending on window width
* Add ability to hide commands depending on driver
  * Currently used only for `startActivity`, `getCurrentActivity`, `getCurrentPackage`, `fingerPrint`, `touchId` and `toggleEnrollTouchId`, since these commands were explicitly labeled for Android or iOS
  * Will be used in future PRs
* Add ability to show extra information for commands
  * Info location depends on whether the command requires parameters. If yes, the info is shown inside the modal; if no, the info is shown as a tooltip
  * Currently used only for `fingerPrint`, `touchId` and `toggleEnrollTouchId`
  * Will be used in future PRs
* Fix `rotateDevice` (parameters were being parsed in incorrect order)

https://github.com/appium/appium-inspector/assets/37242620/caa9edc8-c880-474e-8b86-aa825a91aefa

Additionally, `automationName` is now extracted from `driver` as a separate property upon creating a session. This may resolve situations where `driver` changes (e.g. upon closing the session), but some GUI elements depend on the automation name, causing an app error.


app/renderer/components/Inspector/Commands.js
```diff
@@ -66,35 +65,54 @@ const Commands = (props) => {
     cancelPendingCommand();
   };
 
+  const generateCommandNotes = (notes) => {
+    let generatedNotes = [];
```
mykola-mokhnach, line 69: This could be rewritten in more functional style:

```js
return notes
  .map((note) -> _.isArray(note) ? `${t(note[0])}: ${note[1]}` : t(note))
  .join('; ');
```

app/renderer/components/Inspector/ElementLocator.js
```diff
@@ -33,14 +32,8 @@ const locatorStrategies = (driver) => {
...
 
-const showMissingAutomationNameInfo = (driver, t) => {
-  if (!driver.client.capabilities.automationName) {
-    return <Alert message={t('missingAutomationNameForStrategies')} type={ALERT.INFO} showIcon/>;
-  }
-};
-
 const ElementLocator = (props) => {
-  const { setLocatorTestValue, locatorTestValue, setLocatorTestStrategy, locatorTestStrategy, driver, t } = props;
+  const { setLocatorTestValue, locatorTestValue, setLocatorTestStrategy, locatorTestStrategy, automationName, t } = props;
```
mykola-mokhnach, line 36: maybe it would make sence to integrate [prettify](https://www.npmjs.com/package/prettify) later.

mykola-mokhnach (approved):

jlipps (approved):
This is a significant improvement, thanks! My only thought about the `notes` idea is that, as much as possible, I'd like to shift specific knowledge about platforms back into the drivers and away from the Inspector. Every bit of platform specific knowledge we put into the Inspector is an opportunity for it to drift from the reality of how drivers work, and to make it potentially less easy to integrate future driver support.

I don't have a great idea for how to solve this problem right now. Ideally each driver would expose its own set of supported commands, execute methods, and notes, as part of some kind of API. But this is probably too much to consider right now.

KazuCocoa (approved):

jlipps:
@eglitise whenever you are ready to merge this, let's cut a release!

eglitise:
I certainly agree that ideally all the driver-specific information would come from the drivers themselves. But as you say, right now there's no easy way to retrieve all of this. Currently my plan is to eventually add the `mobile:` extensions and settings for UiA2/XCUITest with their specific notes, but also add a generic field where the user can enter a custom name for an extension/setting, in case it has not yet been added or changed in the Inspector.
This would at least be somewhat future-proof - but I'm open to discussions.

PR #1067, opened 2023-08-28
fix: update robotqa session path for Appium 2

donmezburak:
Changes remote path for Appium 2x. "/wd/hub" removed.


donmezburak:
Hi, We appreciated to you for Appium 2 :) 

We control all parameters on the backend side. So, previous configurations are still valid for our system. By the way, we face a bug for Android dump - we will create a bug after investigating more.

Thanks.

eglitise (approved):

Issue #1103, opened 2023-09-19
feature request: Swipe and touch actions could be done without switching methods


alvarolaserna:
### Current Behavior

Right now if you want to tap by coordinate or swipe by coordinates you have to switch between those two options. It works fine, but it is quite annoying to be switching between one another while interacting with the device

### Suggested Solution

It is possible to merge those two together, and use onMouseDown/onMouseUp instead of onClick from app/renderer/components/Inspector/Screenshot.js . With just a bit of refactoring the UI would be much more user friendly and the inspector can act as if it was a "remote device control software" while using mjpeg server.

### Additional Information

I would be willing to create a PR and work as developer in this project if you like this suggestion.

jlipps:
I think this is a good idea. Now that we have the gestures editor for building up complex gestures one mouse click at a time, it makes sense for swipe/tap to work in a kind of remote control mode.

What do you think @eglitise?

PRs certainly welcome @alvarolaserna!

alvarolaserna:
> I think this is a good idea. Now that we have the gestures editor for building up complex gestures one mouse click at a time, it makes sense for swipe/tap to work in a kind of remote control mode.
> 
> What do you think @eglitise?
> 
> PRs certainly welcome @alvarolaserna!

Awesome, I will create a PR then with the suggestions.

alvarolaserna:
> I think this is a good idea. Now that we have the gestures editor for building up complex gestures one mouse click at a time, it makes sense for swipe/tap to work in a kind of remote control mode.
> 
> What do you think @eglitise?
> 
> PRs certainly welcome @alvarolaserna!

Could you add me to the project so that I can create a PR in the repo? I have it already ready to be pushed. Thanks!

eglitise:
I'm a bit confused how the solution would work. I assume something like, tap would stay unchanged, but swipe functionality would change to a click-hold-drag-release action? If so, that does seem reasonable.
@alvarolaserna you can fork the repo and create a PR through that, it's how I always do it.

alvarolaserna:
> I'm a bit confused how the solution would work. I assume something like, tap would stay unchanged, but swipe functionality would change to a click-hold-drag-release action? If so, that does seem reasonable. @alvarolaserna you can fork the repo and create a PR through that, it's how I always do it.

Yes, thats the idea `click-hold-drag-release`. Okay, I will do that and will create a PR then like that.

[...]

PR #1104, opened 2023-09-21
feat: merge swipe and tap under the same button

alvarolaserna:
This PR is about the enhancement proposed in #1103


eglitise:
There's quite a few things that I would like to be changed here, so instead I've made a branch of my own based on this code. [You can check the changes here](https://github.com/appium/appium-inspector/compare/main...eglitise:appium-inspector:merge-tap-and-swipe?expand=1). Please let me know if it looks okay from your side, behavior-wise.
Note that I haven't created a PR yet, since I still need to test this on iOS (and swiping is somehow broken in Firefox).

alvarolaserna:
> There's quite a few things that I would like to be changed here, so instead I've made a branch of my own based on this code. [You can check the changes here](https://github.com/appium/appium-inspector/compare/main...eglitise:appium-inspector:merge-tap-and-swipe?expand=1). Please let me know if it looks okay from your side, behavior-wise. Note that I haven't created a PR yet, since I still need to test this on iOS (and swiping is somehow broken in Firefox).

I've made all the necessary changes so that it works as you intended, and also I have rename variables so that they make more sense. I have tested it on Firefox/Chrome and iOS/Android, both in mjpeg server mode and non-mjpeg.

app/renderer/components/Inspector/Screenshot.js
```diff
@@ -1,70 +1,80 @@
...
-    } else if (screenshotInteractionMode === SWIPE) {
-      if (!swipeStart) {
-        setSwipeStart(x, y);
-      } else if (!swipeEnd) {
-        setSwipeEnd(x, y);
-        await B.delay(500); // Wait a second to do the swipe so user can see the SVG line
-        await handleDoSwipe({x, y}); // Pass swipeEnd because otherwise it is not retrieved
+  const handleScreenshotDown = async () => {
+    if (screenshotInteractionMode === TAP_SWIPE) {
+      await setCoordStart(x, y);
```
eglitise, line 32: I would put `const { setCoordStart } = props;` at the start of this method, and remove it from the top-level `props` destructuring call.

app/renderer/components/Inspector/Screenshot.js
```diff
@@ -1,70 +1,80 @@
...
+    }
+  };
+
+  const handleScreenshotUp = async () => {
+
```
eglitise, line 37: Like the above, I would put `const { setCoordEnd } = props;` at the start of this method, and remove it from the top-level `props` destructuring call.

app/renderer/components/Inspector/Screenshot.js
```diff
@@ -1,70 +1,80 @@
...
+    if (screenshotInteractionMode === TAP_SWIPE) {
+      await setCoordEnd(x, y);
+      if (Math.abs(coordStart.x - x) < 5 && Math.abs(coordStart.y - y) < 5) {
+        await handleDoTap({x, y}); // Pass coordEnd because otherwise it is not retrieved
+      } else {
+        await handleDoSwipe({x, y}); // Pass coordEnd because otherwise it is not retrieved
       }
     }
```
eglitise, line 45: Would suggest putting `clearCoordAction();` here and removing it from `handleDoTap` and `handleDoSwipe`. You can then also retrieve `clearCoordAction` in a destructuring call at the start of this method, and remove it from the top-level `props` destructuring call.

eglitise (commented):
This looks really good and works as expected! One minor thing is that on Firefox, the circle for swipe start point was either not drawn, or was the same size as the line, but this is not a big deal.
I will test this on iOS tomorrow and then approve :)

app/renderer/components/Inspector/Screenshot.js
alvarolaserna: Yes, agree on all the comments, fixed

app/renderer/components/Inspector/Screenshot.js
alvarolaserna: fixed

eglitise (approved):

jlipps:
thanks a ton @alvarolaserna !

PR #1113, opened 2023-09-27
fix: improve error message for invalid server url

### LINKED ISSUE #1071: feature request: Improve the generic error message for a wrong server url
### Current Behavior

I can observe customers are confused when they see a generic error message by pointing the Inspector to a wrong URL. See, for example, https://discuss.appium.io/t/the-requested-resource-could-not-be-found-or-a-request-was-received-using-an-http-method-that-is-not-supported-by-the-mapped-resource-error/40141

### Suggested Solution

It would make sense to replace the above error with a more helpful message, which could also show the current url and advise to check the actual server log for a valid server url (it should be printed there on the server startup).

### Additional Information

_No response_

eglitise:
This resolves #1071. The wording can probably still be improved.


KazuCocoa (approved):

eglitise:
Made some adjustments to `showError` to accept a new parameter. This allows enhancing the existing error message for invalid URL, instead of adding a second one.

KazuCocoa (approved):

PR #1131, opened 2023-10-16
feat: improve xpath locator suggestions

jlipps:
- Add some attributes that we should consider for uniqueness
- Add search for attributes that might be unique as pairs
- Show a few more attributes by default in the XML source

### DIFF
--- a/app/renderer/components/Inspector/Source.js
+++ b/app/renderer/components/Inspector/Source.js
@@ -10,6 +10,11 @@ const IMPORTANT_ATTRS = [
   'resource-id',
   'AXDescription',
   'AXIdentifier',
+  'text',
+  'label',
+  'value',
+  'id',
+  'class',
 ];
 
 /**
@@ -24,7 +29,7 @@ const Source = (props) => {
     let attrs = [];
 
     for (let attr of Object.keys(attributes)) {
-      if (IMPORTANT_ATTRS.includes(attr) || showAllAttrs) {
+      if ((IMPORTANT_ATTRS.includes(attr) && attributes[attr]) || showAllAttrs) {
         attrs.push(<span key={attr}>&nbsp;
           <i
             className={InspectorStyles.sourceAttrName}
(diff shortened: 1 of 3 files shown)


jlipps:
I've run into this case a few times and feel this is an improvement to our suggested XPath locators.

app/renderer/util.js
```diff
@@ -82,44 +88,92 @@ export function xmlToJSON (source) {
   return firstChild ? translateRecursively(firstChild) : {};
 }
 
+/**
+ * Return information about whether an xpath query results in a unique element, and the non-unique
+ * index of the element in the document if not unique
+ *
+ * @param {string} xpath
+ * @param {DOMDocument} doc
+ * @param {DOMNode} domNode - the current node
```
mykola-mokhnach, line 97: should we also define the returned result type?

package.json
```diff
@@ -161,6 +161,7 @@
     "i18next-localstorage-backend": "4.2.0",
     "ky": "0.25.1",
     "lodash": "4.17.21",
+    "lodash.combinations": "^18.11.1",
```
mykola-mokhnach, line 164: I always being picky about new imports. Perhaps, it would be more effective to just have the single `permutations` function implemented locally using the native stuff: https://stackoverflow.com/questions/43241174/javascript-generating-all-combinations-of-elements-in-a-single-array-in-pairs

mykola-mokhnach (approved):
do we need to modify package.lock in this PR?

jlipps:
> do we need to modify package.lock in this PR?

I suppose we don't! removed

app/renderer/components/Inspector/Source.js
```diff
@@ -10,6 +10,11 @@ const IMPORTANT_ATTRS = [
   'resource-id',
   'AXDescription',
   'AXIdentifier',
+  'text',
+  'label',
+  'value',
+  'id',
+  'class',
```
eglitise, line 17: I don't think `class` should be always shown in the XML view. On Android (which I would expect is the most common use-case for the inspector) this just adds a duplicate of the element type, needlessly bloating every single element item.
At the same time, it would be great if it _was_ shown in webview mode.

jlipps: Hmm yeah good point. I added it thinking of webviews. We probably need to refactor this code so that all of these arrays are sensitive to which platform we're dealing with.

And this could be part of a bigger project that lets drivers declare facts about themselves to the inspector so that we don't have to hardcode these things in the inspector itself.

I've got another open PR, I'll remove class from that one so at least we keep the status quo for now.

PR #1133, opened 2023-10-17
fix: update xpath selector selection logic

jlipps:
I realized there are actually two categories of 'unique' attributes when constructing a selector.

- intentionally unique / static (things like 'id' or 'content-desc' which the developer sets ideally in a way which is unique and doesn't change from run to run. this doesn't mean they are actually always unique. but from the perspective of selector generation, if we find a unique 'id' attribute on an element, then this is sufficient for us to consider the selector unique enough)
- accidentally unique / dynamic (things like 'text' or 'label' which might contain e.g. user generated content or other strings. the fact that they are unique in the xml document is not necessarily a good predictor that they will be unique on subsequent runs of the app.)

This change refactors the selector generation logic to reflect this. If we find an element with an 'accidentally' unique 'text' attribute, we now first attempt to conjoin it with an other attribute (like 'id'), to strengthen the chance that from run to run the selector will not break.

And we still fall back to suggesting a selector based solely on a unique 'text' field, since that is still going to be more likely to work from run to run than a static hierarchical selector.

### DIFF
--- a/app/renderer/util.js
+++ b/app/renderer/util.js
@@ -15,20 +15,29 @@ const UNIQUE_XPATH_ATTRIBUTES = [
   'name',
   'content-desc',
   'id',
+  'resource-id',
   'accessibility-id',
+];
+
+// Attributes that we should recommend as a fallback but ideally only in conjunction with other
+// attributes
+const MAYBE_UNIQUE_XPATH_ATTRIBUTES = [
+  'label',
   'text',
+  'value',
+  'class',
+  'type',
+];
+
+const UNIQUE_CLASS_CHAIN_ATTRIBUTES = [
+  'name',
   'label',
   'value',
 ];
 
-const UNIQUE_CLASS_CHAIN_ATTRIBUTES = [
-  'label',
-  'name',
-  'value',
-];
 const UNIQUE_PREDICATE_ATTRIBUTES = [
-  'label',
   'name',
+  'label',
   'value',
   'type',
 ];
@@ -97,7 +106,7 @@ export function xmlToJSON (source) {
  * @returns {[boolean, number?]} tuple consisting of (1) whether the xpath is unique and (2) its index in
  * the set of other similar nodes if not unique
  */
-function isXpathUnique(xpath, doc, domNode) {
+function determineXpathUniqueness(xpath, doc, domNode) {
   let othersWithAttr = [];
 
   // If the XPath does not parse, move to the next unique attribute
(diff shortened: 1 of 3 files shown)


app/renderer/util.js
```diff
@@ -114,69 +120,124 @@ function isXpathUnique(xpath, doc, domNode) {
...
 
+/**
+ * Given an xml doc and a current dom node, try to find a robust xpath selector qualified by
+ * key attributes, which is unique in the document (or unique plus index).
+ *
+ * @param {string} xpath
+ * @param {DOMDocument} doc
+ * @param {Array<string>|Array<[string, string]>} attrs - a list of attributes to consider, or
+ * a list of pairs of attributes to consider in conjunction
+ * @param {boolean} isPairs - whether we are looking for single attributes or a conjunction of
```
mykola-mokhnach, line 131: `[isPairs=false]` is the proper format

mykola-mokhnach: also, we probably don't need this arg as we could still dynamically detect the format of each item

app/renderer/util.js
```diff
@@ -114,69 +120,124 @@ function isXpathUnique(xpath, doc, domNode) {
...
+ */
+function getUniqueXPath(doc, domNode, attrs, isPairs = false) {
+  let uniqueXpath, semiUniqueXpath;
+  const tagForXpath = domNode.tagName || '*';
+
+  for (const attrName of attrs) {
+    let xpath;
+    if (isPairs) {
+      const [attr1Name, attr2Name] = attrName;
+      const attr1Value = domNode.getAttribute(attr1Name);
```
mykola-mokhnach, line 145: `const [attr1Value, attr2Value] = attrName.map(domNode.getAttribute.bind(domNode))`

app/renderer/util.js
```diff
@@ -114,69 +120,124 @@ function isXpathUnique(xpath, doc, domNode) {
...
+ * key attributes, which is unique in the document (or unique plus index).
+ *
+ * @param {string} xpath
+ * @param {DOMDocument} doc
+ * @param {Array<string>|Array<[string, string]>} attrs - a list of attributes to consider, or
+ * a list of pairs of attributes to consider in conjunction
+ * @param {boolean} isPairs - whether we are looking for single attributes or a conjunction of
+ * attributes
+ *
+ * @returns {[string?, boolean?]} tuple consisting of (1) the xpath selector discovered, and (2)
```
mykola-mokhnach, line 134: actually question marks mean the same as `string|null`. For undefined case above `string|undefined` should be used instead

app/renderer/util.js
```diff
@@ -114,69 +120,124 @@ function isXpathUnique(xpath, doc, domNode) {
...
-    for (const attrName of uniqueAttributes) {
-      const attrValue = domNode.getAttribute(attrName);
-      if (!attrValue) {
-        continue;
-      }
-      const xpath = `//${tagForXpath}[@${attrName}="${attrValue}"]`;
-      const [isUnique, indexIfNotUnique] = isXpathUnique(xpath, doc, domNode);
-      if (isUnique) {
+    const attrsForPairs = [...UNIQUE_XPATH_ATTRIBUTES, ...MAYBE_UNIQUE_XPATH_ATTRIBUTES];
+    const pairAttributes = attrsForPairs.flatMap((v1, i) =>
```
mykola-mokhnach, line 195: attributePairsPermutations

app/renderer/util.js
```diff
@@ -114,69 +120,124 @@ function isXpathUnique(xpath, doc, domNode) {
...
+    ];
+
+    // It's possible that in all of these cases we don't find a truly unique selector. But
+    // a selector qualified by attribute with an index attached like //*[@id="foo"][1] is still
+    // better than a fully path-based selector. We call this a 'semi unique xpath'
+    let semiUniqueXpath;
+
+    // Go through each of our cases and look for selectors for each case in order
+    for (const {attrs, isPairs} of cases) {
+      let [xpath, isFullyUnique] = getUniqueXPath(doc, domNode, attrs, isPairs);
```
mykola-mokhnach, line 220: const?

eglitise (approved):
Overall looks good to me 👍 
I'm mainly curious about the classification and ordering of items in `UNIQUE_XPATH_ATTRIBUTES`/`MAYBE_UNIQUE_XPATH_ATTRIBUTES`. I'm not entirely sure if `name` should be the first choice for the former, and _maybe_ `content-desc` is a better fit for the latter...? I will try to test this on some real apps to get a rough idea.

app/renderer/util.js
```diff
@@ -114,69 +121,122 @@ function isXpathUnique(xpath, doc, domNode) {
...
+      }
+      xpath = `//${tagForXpath}[@${attr1Name}="${attr1Value}" and @${attr2Name}="${attr2Value}"]`;
+    } else {
+      const attrValue = domNode.getAttribute(attrName);
+      if (!attrValue) {
+        continue;
+      }
+      xpath = `//${tagForXpath}[@${attrName}="${attrValue}"]`;
+    }
+    const [isUnique, indexIfNotUnique] = isXpathUnique(xpath, doc, domNode);
```
mykola-mokhnach, line 157: a common agreement for function starting with `is` would be to return a boolean value. How about renaming this one to determineXpathUniqueness?

mykola-mokhnach (approved):

eglitise:
Alright, I checked out some apps with this new logic. Looks good, some suggestions:
* Move `name` from `UNIQUE_XPATH_ATTRIBUTES` to `MAYBE_UNIQUE_XPATH_ATTRIBUTES`. On iOS this parameter generally matches `label`, which is already under `MAYBE_UNIQUE_XPATH_ATTRIBUTES`.
* Put `content-desc` as the last element in `UNIQUE_XPATH_ATTRIBUTES`, or even move it to `MAYBE_UNIQUE_XPATH_ATTRIBUTES`. From what I can tell, it serves a similar purpose to element text.
* Match the item ordering in `MAYBE_UNIQUE_XPATH_ATTRIBUTES` and `UNIQUE_CLASS_CHAIN_ATTRIBUTES`. Like I said, on iOS the `name` and `label` are commonly identical, so right now it may happen that the suggested class chain uses `label`, while the xpath uses `name`.
* Possibly add `class` and `type` to `MAYBE_UNIQUE_XPATH_ATTRIBUTES` to improve situations like this:
image

eglitise:
@jlipps thanks for the info, I did not know about `name` and `content-desc` being linked to the accessibility labels. With that I'm thinking it's not really a big deal if both of them are still included in `UNIQUE_XPATH_ATTRIBUTES`. Same for the ordering - preferring accessibility labels over `id`s feels like it reflects real user behavior better.

At the same time, I would still propose changing the order of `UNIQUE_CLASS_CHAIN_ATTRIBUTES` to `[name, label, value]`, to match the attributes for the suggested class chain and xpath. And if, indeed, the `name` for an element comes from the accessibility label, and differs from `label` and `value`, it should probably be preferred.

jlipps:
ok sounds good will do!

PR #1135, opened 2023-10-19
fix: update xpath selector selection logic yet again

jlipps:
this PR primarily  removes class/type from xpath uniqueness attributes. instead adds logic to look for unique node names and refer to those if necessary (since class/type are always the same as node names)

### DIFF
--- a/app/renderer/util.js
+++ b/app/renderer/util.js
@@ -25,8 +25,6 @@ const MAYBE_UNIQUE_XPATH_ATTRIBUTES = [
   'label',
   'text',
   'value',
-  'class',
-  'type',
 ];
 
 const UNIQUE_CLASS_CHAIN_ATTRIBUTES = [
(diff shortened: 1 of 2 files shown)


app/renderer/util.js
```diff
@@ -139,7 +137,25 @@ function getUniqueXPath(doc, domNode, attrs) {
...
+  // If we're looking for a unique //<nodetype>, return it only if it's actually unique. No
+  // semi-uniqueness here!
+  if (isNodeName) {
+    let xpath = `//${domNode.tagName}`;
+    const [isUnique] = determineXpathUniqueness(xpath, doc, domNode);
+    // even if this node name is unique, if it's the root node, we don't want to refer to it using
+    // '//' but rather '/'
+    if (!domNode?.parentNode?.tagName) {
+      xpath = `/${domNode.tagName}`;
+    }
```
eglitise, line 151: I'm not sure how useful this condition is. From the apps I checked, Android always has two `android.widget.FrameLayout` as the roots, so `isUnique` will be false anyway; and iOS has `XCUIElementTypeApplication` with filled out `name` and `label` fields, so the xpath generator will rely on those, and not even reach this base case. So currently I haven't found a usecase for this.

Also, CI is failing here because we can't use optional chaining (Parcel v1 limitation). It probably needs to be something like `if (!(domNode.parentNode && domNode.parentNode.tagName))`.

Finally, if we really must keep the condition, it should probably be moved inside the `isUnique` block on the next line.

eglitise (changes requested):
Other than the edge case in my other comment, looks good!

PR #1147, opened 2023-10-22
feat: add recorder support for Python client v3

eglitise:
This adds recorder support for Appium Python client v3 and closes #1145.
* Change capabilities parsing to options style
* Remove unsupported methods (`launch_app`, `close_app`, `reset`)

PR #1148, opened 2023-10-22
chore: remove unused template commands

KazuCocoa:
As same as https://github.com/appium/appium-inspector/pull/1147, we can remove launchApp/closeApp/reset commands from template since these commands no longer exists in this inspector library


eglitise (approved):

eglitise:
Can these commands also be removed from the Ruby framework file?

eglitise (changes requested):
Can these commands also be removed from the Ruby framework file?

KazuCocoa:
Nice catch, missed that part.

eglitise (approved):

PR #1152, opened 2023-10-25
fix: improve error for invalid JSON in command argument

eglitise:
Currently, when attempting to call the `executeScript` or `updateSettings` command, if an invalid JSON argument is given, the app will return 2 errors - one for the invalid JSON, and another for the actual method execution.

Changes included in this PR:
* do not call the client method if JSON parsing fails
* add translation key for the JSON error, which also adds the failed JSON string inside the error

### DIFF
--- a/app/renderer/components/Inspector/Commands.js
+++ b/app/renderer/components/Inspector/Commands.js
@@ -17,6 +17,18 @@ const Commands = (props) => {
     }
   };
 
+  const parseJsonString = (jsonString) => {
+    try {
+      return JSON.parse(jsonString);
+    } catch (e) {
+      notification.error({
+        message: t('invalidJson', {json: jsonString}),
+        duration: 5,
+      });
+      return null;
+    }
+  };
+
   const executeCommand = () => {
     const { args, commandName, command } = pendingCommand;
     const { refresh } = command;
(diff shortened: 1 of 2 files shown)


KazuCocoa (approved):

app/renderer/components/Inspector/Commands.js
```diff
@@ -24,6 +24,19 @@ const Commands = (props) => {
     // Make a copy of the arguments to avoid state mutation
     let copiedArgs = _.cloneDeep(args);
 
+    let parsingSuccess = true;
```
mykola-mokhnach, line 27: how about `isJsonValid`

app/renderer/components/Inspector/Commands.js
```diff
@@ -24,6 +24,19 @@ const Commands = (props) => {
...
+    const parseJsonArgument = (index) => {
+      try {
+        copiedArgs[index] = JSON.parse(args[index]);
```
mykola-mokhnach, line 30: Closures might be dangerous if not used carefully. maybe we could simply move it to a separate function and return object/tuple instead?

mykola-mokhnach (approved):

app/renderer/components/Inspector/Commands.js
eglitise: Moved to a separate function

app/renderer/components/Inspector/Commands.js
```diff
@@ -17,13 +17,27 @@ const Commands = (props) => {
     }
   };
 
+  const parseJsonString = (jsonString) => {
+    try {
+      return JSON.parse(jsonString);
+    } catch (e) {
+      notification.error({
```
mykola-mokhnach, line 24: I assume it would be also helpful to show the error message itself

mykola-mokhnach (approved):

app/renderer/components/Inspector/Commands.js
eglitise: Added the error message in the notification description

mykola-mokhnach (approved):

PR #1157, opened 2023-10-27
feat: update Java recorder code

eglitise:
This PR updates the recorder-generated code for the Java client:
* Update boilerplate to support Java client ~8+~ 9+ (use `BaseOptions` instead of `DesiredCapabilities`)
* Add support for non-iOS/Android drivers
* Update element finding strategies to use `AppiumBy`
* Adjust `tap` and `swipe` implementation (add missing imports, improve formatting)
* Implement missing driver commands (`executeDriver`, `startActivity`, `activateApp`, `terminateApp`)
  * `setTimeouts` still remains unimplemented since I am not sure if it even exists in clients or Appium itself


app/renderer/lib/client-frameworks/java.js
```diff
@@ -8,55 +8,46 @@ class JavaFramework extends Framework {
...
 
   wrapWithBoilerplate (code) {
-    let host = JSON.stringify(this.host);
-    let port = JSON.stringify(this.port);
     let [pkg, cls] = (() => {
       if (this.caps.platformName) {
         switch (this.caps.platformName.toLowerCase()) {
           case 'ios': return ['ios', 'IOSDriver'];
           case 'android': return ['android', 'AndroidDriver'];
+          case 'windows': return ['windows', 'WindowsDriver'];
```
mykola-mokhnach, line 16: ❤️

app/renderer/lib/client-frameworks/java.js
```diff
@@ -72,31 +63,26 @@ ${this.indent(code, 4)}
...
-      'accessibility id': 'AccessibilityId',
-      'id': 'Id',
-      'class name': 'ClassName',
-      'name': 'Name',
-      '-android uiautomator': 'AndroidUIAutomator',
-      '-android datamatcher': 'AndroidDataMatcher',
-      '-android viewtag': 'AndroidViewTag',
-      '-ios predicate string': 'IosNsPredicate',
-      '-ios class chain': 'IosClassChain',
+      'xpath': 'xpath',
```
mykola-mokhnach, line 68: do we really need to quote object keys if these don't contain special characters?

app/renderer/lib/client-frameworks/java.js
```diff
@@ -72,31 +63,26 @@ ${this.indent(code, 4)}
...
+      'name': 'name',
+      '-android uiautomator': 'androidUIAutomator',
+      '-android datamatcher': 'androidDataMatcher',
+      '-android viewtag': 'androidViewTag',
+      '-ios predicate string': 'iOSNsPredicateString',
+      '-ios class chain': 'iOSClassChain',
     };
     if (!suffixMap[strategy]) {
-      throw new Error(`Strategy ${strategy} can't be code-gened`);
+      throw new Error(`Strategy '${strategy}' can't be code-gened`);
```
mykola-mokhnach, line 80: We did not know how to generate the code for the '${strategy}' location strategy

app/renderer/lib/client-frameworks/java.js
```diff
@@ -119,65 +105,72 @@ ${this.indent(code, 4)}
...
+  // Execute Script
+
+  codeFor_executeScript (varNameIgnore, varIndexIgnore, scriptCmd, jsonArg) {
+    let assembledCommand;
+    if (jsonArg === undefined) {
+      assembledCommand = `driver.executeScript("${scriptCmd}");`;
+    } else {
+      // change the JSON object into a format accepted by Map.ofEntries:
+      // combine the keys and values into a flattened array, then stringify it
+      const argsValuesArray = [].concat(...(_.toPairs(JSON.parse(jsonArg))));
```
mykola-mokhnach, line 149: You can use _.flatMap call here

app/renderer/lib/client-frameworks/java.js
```diff
@@ -119,65 +105,72 @@ ${this.indent(code, 4)}
...
-    return `String packageName = driver.currentPackage()`;
+  // App Management
+
+  codeFor_startActivity (varNameIgnore, varIndexIgnore, ...args) {
+    const argNames = ['appPackage', 'appActivity', 'appWaitPackage', 'intentAction', 'intentCategory',
+      'intentFlags', 'optionalIntentArguments', 'dontStopAppOnReset'];
+    // zip argument names and values into tuples, but only if the value was provided
+    // then flatten the array of tuples and stringify it
+    const argsValuesArray = argNames.flatMap((k, i) => (args[i] === undefined) ? [] : [k, args[i]]);
+    const argsValuesString = JSON.stringify(argsValuesArray).slice(1, -1);
```
mykola-mokhnach, line 164: maybe we could dedupe this into a separate function and at the same time make it more readable

app/renderer/lib/client-frameworks/java.js
```diff
@@ -200,6 +201,8 @@ ${this.indent(code, 4)}
     return `Map<String, String> appStrings = driver.getAppStringMap(${language ? `${language}, ` : ''}${stringFile ? `"${stringFile}` : ''});`;
```
mykola-mokhnach, line 201: since Java11, that we only support now we could same some space and use `var appStrings = ...`  syntax

app/renderer/lib/client-frameworks/java.js
```diff
@@ -208,6 +211,57 @@ ${this.indent(code, 4)}
...
+  codeFor_unlock () {
+    return `driver.unlockDevice();`;
+  }
+
+  codeFor_isLocked () {
+    return `boolean isLocked = driver.isDeviceLocked();`;
+  }
+
+  codeFor_rotateDevice (varNameIgnore, varIndexIgnore, x, y, radius, rotation, touchCount, duration) {
+    return `driver.rotate(new DeviceRotation(${x}, ${y}, ${radius}, ${rotation}, ${touchCount}, ${duration}));`;
```
mykola-mokhnach, line 248: this may probably be confusing because we only support the limited set of arguments for this API. Random args won't work

app/renderer/lib/client-frameworks/java.js
```diff
@@ -119,65 +105,72 @@ ${this.indent(code, 4)}
...
-    Point tapPoint = new Point(${x}, ${y});
-    Sequence tap = new Sequence(FINGER, 1);
-        tap.addAction(FINGER.createPointerMove(Duration.ofMillis(0),
-                PointerInput.Origin.viewport(), tapPoint.x, tapPoint.y));
-        tap.addAction(FINGER.createPointerDown(PointerInput.MouseButton.LEFT.asArg()));
-        tap.addAction(FINGER.createPointerMove(Duration.ofMillis(50),
-                PointerInput.Origin.viewport(), tapPoint.x, tapPoint.y));
-        tap.addAction(FINGER.createPointerUp(PointerInput.MouseButton.LEFT.asArg()));
-        driver.perform(Arrays.asList(tap));
+final PointerInput FINGER = new PointerInput(PointerInput.Kind.TOUCH, "finger");
```
mykola-mokhnach, line 111: java linter usually freaks out if it detects local variables named in uppercase. the rule of thumb is to use camelCase for local variable names. Uppercase is only used for class-level constants

app/renderer/lib/client-frameworks/java.js
mykola-mokhnach: same APIs below are only valid for mobile platforms. Not sure how much they would make sense for gecko or safari drivers

eglitise: True, but in that case the execution of the action will also return an error. I guess the best way to handle this would be to only record an action if its execution succeeds, but I haven't looked into that yet.
IMO adding the remaining drivers here is still better, since now at least it will no longer be `UnknownDriver`.

app/renderer/lib/client-frameworks/java.js
eglitise: I just added them for consistency, but I'll remove them

app/renderer/lib/client-frameworks/java.js
eglitise: Ok, I just read that all caps is recommended for `final` variables, but I'll change it to camelcase

app/renderer/lib/client-frameworks/java.js
eglitise: Alright, will change these occurrences elsewhere in the code as well

app/renderer/lib/client-frameworks/java.js
eglitise: I'll just rewrite this so that it reuses `codeFor_executeScript`

app/renderer/lib/client-frameworks/java.js
eglitise: Reading into this, it seems that the parameters coming from Inspector are for the deprecated `mobileRotation` command, instead of Selenium's rotation endpoints. I'd propose leaving this here for now, so I can later adjust it everywhere.

mykola-mokhnach (approved):

app/renderer/lib/client-frameworks/java.js
```diff
@@ -8,55 +8,46 @@ class JavaFramework extends Framework {
...
-            e.printStackTrace();
-        }
-        return null;
-    }
+  public void setUp() {
+    var options = new BaseOptions()
+${capStr};
 
-    driver = new ${cls}(this.remoteUrl(HOST, PORT), desiredCapabilities);
+    driver = new ${cls}(new URL("${this.serverUrl}"), options);
```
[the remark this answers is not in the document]
eglitise: it's been a while since I worked with Java 😅 I figured the try/catch was optional

PR #1162, opened 2023-10-31
fix: adjust highlighters for Android tablets

### LINKED ISSUE #1154: bug: The click point does not match the accurate position of the element in latest version(2023.10.4)

### Current Behavior

In the latest version [Appium-Inspector-mac-2023.10.4.dmg](https://github.com/appium/appium-inspector/releases/download/v2023.10.4/Appium-Inspector-mac-2023.10.4.dmg), the click point does not match the accurate position of the icon.
Screen Shot 2023-10-26 at 16 26 08

### Expected Behavior
The click point should match the accurate position of the icon.
There is no such problem when using the older version.(for example [Appium-Inspector-mac-2023.6.1.dmg](https://github.com/appium/appium-inspector/releases/download/v2023.6.1/Appium-Inspector-mac-2023.6.1.dmg)) 
Screen Shot 2023-10-26 at 16 36 31

### Operating System

Mac

### Appium Inspector Version

Appium-Inspector-mac-2023.10.4.dmg

### Appium Version

2.2.0

### Further Information

_No response_

eglitise:
This PR fixes an issue where element highlighters for Android tablets were being drawn incorrectly.
The core issue stems from the `deviceScreenSize` parameters, which appear to match portrait mode for phones, but landscape mode for tablets. This has now been adjusted for.
Resolves #1154.

### DIFF
--- a/app/renderer/lib/appium-client.js
+++ b/app/renderer/lib/appium-client.js
@@ -211,11 +211,13 @@ export default class AppiumClient {
       if (_.toLower(platformName) === 'android' && _.toLower(automationName) === 'uiautomator2') {
         // returned Android height and width can both be affected by UiAutomator2 calculations
         // we stick with device dimensions, but swap them depending on detected orientation
+        // deviceScreenSize value fits portrait mode for phones, but landscape mode for tablets
         const [width, height] = deviceScreenSize.split('x');
-        if (windowSize.height > windowSize.width) { // portrait mode
+        if ((windowSize.height > windowSize.width && height > width) || // phone portrait mode
+            (windowSize.height < windowSize.width && height < width)) { // tablet landscape mode
           windowSize.height = height;
           windowSize.width = width;
-        } else { // landscape mode
+        } else { // phone landscape mode / tablet portrait mode
           windowSize.height = width;
           windowSize.width = height;
         }


app/renderer/lib/appium-client.js
```diff
@@ -211,11 +211,13 @@ export default class AppiumClient {
       if (_.toLower(platformName) === 'android' && _.toLower(automationName) === 'uiautomator2') {
         // returned Android height and width can both be affected by UiAutomator2 calculations
         // we stick with device dimensions, but swap them depending on detected orientation
+        // deviceScreenSize value fits portrait mode for phones, but landscape mode for tablets
         const [width, height] = deviceScreenSize.split('x');
-        if (windowSize.height > windowSize.width) { // portrait mode
+        if ((windowSize.height > windowSize.width && height > width) || // phone portrait mode
```
mykola-mokhnach, line 216: should it be `>=` instead of `>` ?

mykola-mokhnach (approved):

app/renderer/lib/appium-client.js
eglitise: It's definitely an edgecase. If the dimensions are equal for both `windowSize` and `deviceScreenSize`, then either branch is fine. The only way problems could arise is if `width == height`, but `windowSize.height != windowSize.width`. I'm not sure how likely this is to actually happen.

jlipps: I'm also finding this condition a bit hard to read. Would it make more sense to reverse it?

eglitise: Created #1163 to adjust this

PR #1167, opened 2023-11-02
feat: update Python recorder code

eglitise:
This PR updates the recorder-generated code for the Python client:

* Add missing driver commands (`activateApp`, `terminateApp`, `sendSMS`)
* Fix `executeScript` implementation
* Apply conversion to Python values also for parameters inside arrays and objects
* Improve formatting for generated code (remove semicolons, add variable assignment where needed)

### DIFF
--- a/app/renderer/lib/client-frameworks/python.js
+++ b/app/renderer/lib/client-frameworks/python.js
@@ -9,6 +9,14 @@ class PythonFramework extends Framework {
   getPythonVal (jsonVal) {
     if (typeof jsonVal === 'boolean') {
       return jsonVal ? 'True' : 'False';
+    } else if (Array.isArray(jsonVal)) {
+      const convertedItems = jsonVal.map((item) => this.getPythonVal(item));
+      return `[${convertedItems.join(', ')}]`;
+    } else if (typeof jsonVal === 'object') {
+      const convertedItems = Object.keys(jsonVal).map((k) =>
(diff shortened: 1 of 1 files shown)


app/renderer/lib/client-frameworks/python.js
```diff
@@ -121,25 +132,35 @@ actions.perform()
...
-    return `is_app_installed = driver.is_app_installed('${app}');`;
+    return `is_app_installed = driver.is_app_installed('${app}')`;
   }
 
   codeFor_background (varNameIgnore, varIndexIgnore, timeout) {
     return `driver.background_app(${timeout})`;
   }
 
+  codeFor_activateApp (varNameIgnore, varIndexIgnore, app) {
+    return `driver.activate_app('${app}')`;
```
KazuCocoa, line 147: I wondered if this command ideally needs to be mobile: command -> https://github.com/appium/appium-uiautomator2-driver?tab=readme-ov-file#mobile-activateapp
but for now, looks good to me. the argument name should be changed for mobile command for android/ios and other drivers

KazuCocoa (approved):

app/renderer/lib/client-frameworks/python.js
eglitise: I think this is fine, because the implementation in the Python client calls `mobile: activateApp` anyway: https://github.com/appium/python-client/blob/master/appium/webdriver/extensions/applications.py#L187

PR #1172, opened 2023-11-03
feat: update Robot Framework recorder code

eglitise:
This PR updates the recorder-generated code for the Robot Framework Appium client. Not sure how well maintained the client is, but the recorder for it should at least be usable now.

* Fix valid locator strategies
* Fix conversion to Robot variable values (wrap numbers and booleans in `${}`)
* For element actions, first save the locator as a variable (matches the approach of other frameworks)
* Implement missing driver commands


app/renderer/lib/client-frameworks/robot.js
```diff
@@ -50,285 +43,314 @@ ${this.indent(code, 4)}
...
+    return `&{scriptArgument}    Create Dictionary    ${argsValuesStrings.join('    ')}
+Execute Script    ${scriptCmd}    &{scriptArgument}`;
   }
 
-  // TODO: Add these robot framework commands
+  // App Management
+
   codeFor_getCurrentActivity () {
-    return '';
+    return '${activity_name} =    Get Activity';
```
mykola-mokhnach, line 106: I assume we can still call `mobie: ` extensions from RF

mykola-mokhnach: same below

eglitise: I think we can, but I'm not really sure of the best approach. The user might have an older driver version that does not support all of the replacement extensions. Ideally I would like to have separate GUI buttons for the various `mobile` extension commands, and leave the current buttons linked to their original, deprecated functionality, with added deprecation labels.
(I also realise I already replaced many such deprecated methods in #1173, so now I'm not sure if I should change that back)

mykola-mokhnach: if they have older driver then an error message would be shown, so this is ok. Lets always aim the most recent version available

mykola-mokhnach (approved):

app/renderer/lib/client-frameworks/robot.js
eglitise: I've now changed all the methods for which the equivalent Ruby code was calling mobile extensions. Common methods (i.e. without language-specific variable assignment) were moved to the common class file.

mykola-mokhnach (approved):

PR #1176, opened 2023-11-05
feat: add Hungarian language support

KazuCocoa:


eglitise (changes requested):
We also need a translation key, otherwise the language menu will only use `hu`.

eglitise (approved):

eglitise:
I'm guessing the CI failures are due to some incompatibility between Node 20.9.0 and Python 3.12?

Issue #1115, opened 2023-09-29
Update crowdin api from v1 to v2


KazuCocoa:
### Current Behavior

v1

### Suggested Solution

move to v2

KazuCocoa:
`https://api.crowdin.com/api/v2/projects` with proper request format: https://developer.crowdin.com/api/v2/#section/Introduction/Requests

Project id is `346705`.

KazuCocoa:
- https://github.com/appium/appium-inspector/blob/main/ci-jobs/crowdin-sync-translations.js#L20
- https://github.com/appium/appium-inspector/blob/main/ci-jobs/crowdin-update-resources.js#L18

KazuCocoa:
https://developer.crowdin.com/assets/api/api_v1_docs_crowdin.pdf is v1.

> Upload the latest version of your source files (the ones, that should be localized) to your Crowdin project.

maybe this will be:

https://developer.crowdin.com/api/v2/#operation/api.projects.files.put


hm... it has `storageId`, but I am not sure what this means. We must learn the new API first...

PR #1177, opened 2023-11-05
refactor: update Crowdin API to v2

mykola-mokhnach:
Fixes https://github.com/appium/appium-inspector/issues/1115

Also replaced long-time-deprecated request libs with axios


.github/workflows/crowdin-update-resources.yml
```diff
@@ -22,6 +22,6 @@ jobs:
     - name: Crowdin Update
       run: npm run crowdin-update
       env:
-        CROWDIN_PROJECT_ID: appium-desktop
-        CROWDIN_PROJECT_KEY: ${{ secrets.CROWDIN_PROJECT_KEY }}
-      
+        # appium-desktop
+        CROWDIN_PROJECT_ID: 346705
+        CROWDIN_TOKEN: ${{ secrets.CROWDIN_TOKEN }}
```
KazuCocoa, line 27: Do you use this from your api token, or from 1pass word (the one I previously created, which ties to my account)
Just want to remove the token from 1pass if that was not used to reduce confusion

ci-jobs/crowdin-sync-translations.js
```diff
@@ -1,72 +1,54 @@
...
+    intervalMs: 1000,
   });
+  const {data: downloadData} = await performApiRequest(`/translations/builds/${buildId}/download`);
+  log.info(`Downloading translations to '${dstPath}'`);
+  await net.downloadFile(downloadData.url, dstPath);
 }
 
 async function main () {
-  await exportTranslations();
+  const buildId = await buildTranslations();
```
KazuCocoa, line 48: oh, new one must call build api first...

KazuCocoa (commented):
looks good to me. Did you run the script once?

.github/workflows/crowdin-update-resources.yml
mykola-mokhnach: The v2 API uses a new token format. I have added it to the `appium` account at crowdin and to github secrets.

KazuCocoa: Yes, i added the new format's one before (maybe last month or so), so to avoid duplications. Let me delete mine from the 1pass, that has `Crowdin's v2 access token. It ties with a personal access token by Kazu` in the note

KazuCocoa: btw, in the Appium vaults?

KazuCocoa (approved):

.github/workflows/crowdin-update-resources.yml
KazuCocoa: (deleted mine)

ci-jobs/crowdin-common.js
```diff
@@ -0,0 +1,50 @@
...
+    method,
+    headers: {
+      Authorization: `Bearer ${API_TOKEN}`,
+      'Content-Type': 'application/json',
+      'User-Agent': USER_AGENT,
+      ...(headers || {}),
+    },
+    url,
+    data: payload,
+  })).data;
```
eglitise, line 45: I think it would be useful to have handling/logging of non-2XX response codes

ci-jobs/crowdin-update-resources.js
```diff
@@ -1,39 +1,51 @@
...
-request(options, (err, res, body) => {
-  if (err) {
-    throw err;
-  }
+async function updateFile(fileId, storageId) {
+  log.info(`Updating the project with the newly uploaded '${RESOURCE_NAME}' instance`);
+  await performApiRequest(`/files/${fileId}`, {
+    method: 'PUT',
+    payload: {
+      storageId,
```
eglitise, line 40: Looking at the API for this call, it seems that it clears all translations and approvals by default. We probably need to add `updateOption: "keep_translations_and_approvals"` to the payload.

ci-jobs/crowdin-sync-translations.js
```diff
@@ -1,72 +1,54 @@
...
-        if (res.statusCode >= 400) {
-          return reject(`Cannot export the translated resources in Crowdin. Error code: ${res.statusCode}`);
-        }
-        log.info(`Successfully exported Crowdin translations`);
-      })
-      .on('close', resolve);
+async function buildTranslations () {
+  log.info('Building project translations');
+  const {data: buildData} = await performApiRequest('/translations/builds', {
+    method: 'POST',
```
eglitise, line 20: The Crowdin documentation doesn't say whether `skipUntranslatedStrings`, `skipUntranslatedFiles` and `exportApprovedOnly` are set to `false` by default, but I assume they are.

ci-jobs/crowdin-common.js
```diff
@@ -0,0 +1,50 @@
+const { logger } = require('@appium/support');
+const axios = require('axios');
+const path = require('path');
+const _ = require('lodash');
```
eglitise, line 4: Can we use ES6 `import`/`export` here and in the other 2 files?

eglitise (commented):
Thanks!
I noticed that Crowdin also has their own JS client (https://github.com/crowdin/crowdin-api-client-js), maybe it could simplify some parts of the code? It could even be added to `devDependencies` instead of `dependencies`.

ci-jobs/crowdin-common.js
mykola-mokhnach: such handling is present by default in axios. Check the documentation on `validateStatus` property.

ci-jobs/crowdin-update-resources.js
mykola-mokhnach: > Defines whether to keep existing translations and approvals for updated strings

It basically defines if we want to keep translations for strings that were **updated** in the original locale. We probably don't as they require translation updates

ci-jobs/crowdin-sync-translations.js
mykola-mokhnach: as far as I can see they are

ci-jobs/crowdin-common.js
mykola-mokhnach: Not sure how to enforce that in raw node.js. I usually rely on babel or TS transpilers

mykola-mokhnach:
> Thanks! I noticed that Crowdin also has their own JS client (https://github.com/crowdin/crowdin-api-client-js), maybe it could simplify some parts of the code? It could even be added to `devDependencies` instead of `dependencies`.

I am usually not a big fan of adding new dependencies in case the original implementation is simple and does not require a lot of boilerplate code.

ci-jobs/crowdin-common.js
mykola-mokhnach: nvm, found a tutorial on manual CJS->ESM conversion

.github/workflows/crowdin-update-resources.yml
mykola-mokhnach: > btw, in the Appium vaults?

I don't have 1pass on my machine
I could send the token to you, so you could update 1pass

eglitise (approved):

.github/workflows/crowdin-update-resources.yml
KazuCocoa: Oh, i see. either is ok for me, or I can send an invitation for you. I wanted to clarify where the token was.

PR #1178, opened 2023-11-06
chore: Replace uuid usage with native primitives

mykola-mokhnach:


eglitise:
Unfortunately this doesn't work :( I tried to save a gesture in both the Electron and browser versions, and both returned an error about randomUUID not being a function.

mykola-mokhnach:
> Unfortunately this doesn't work :( I tried to save a gesture in both the Electron and browser versions, and both returned an error about randomUUID not being a function.

I am now importing it from main module. Could you please try it again?

eglitise:
No luck, sadly. The returned error in the browser version is: `TypeError: (0 , _crypto.randomUUID) is not a function`
Worth noting that the desktop app's minified code looks similar:
```js
        function i() {
            return (0,
            e.randomUUID)()
        }
```
However, I changed the code to `import 'crypto'` / `return crypto.randomUUID();`, and that somehow fixed the issue in the browser version. Unfortunately, the desktop app still doesn't work, likely because the current Electron version (13.6.9) bundles Node 14.16, which doesn't support this function.

mykola-mokhnach:
> No luck, sadly. The returned error in the browser version is: `TypeError: (0 , _crypto.randomUUID) is not a function` Worth noting that the desktop app's minified code looks similar:
> 
> ```js
>         function i() {
>             return (0,
>             e.randomUUID)()
>         }
> ```
> 
> However, I changed the code to `import 'crypto'` / `return crypto.randomUUID();`, and that somehow fixed the issue in the browser version. Unfortunately, the desktop app still doesn't work, likely because the current Electron version (13.6.9) bundles Node 14.16, which doesn't support this function.

should we then leave it until Electron gets updated above v14?

eglitise:
I would suggest closing it. The main blocker for upgrading to Electron 14 is that it breaks `electron-settings`, which in turn is also fairly outdated and should probably be replaced with some alternative. At the same time, the current user settings data (most importantly the saved sessions and gestures) should be migrated. Not sure if there is an easy way to resolve this.

mykola-mokhnach:
> I would suggest closing it. The main blocker for upgrading to Electron 14 is that it breaks `electron-settings`, which in turn is also fairly outdated and should probably be replaced with some alternative. At the same time, the current user settings data (most importantly the saved sessions and gestures) should be migrated. Not sure if there is an easy way to resolve this.

In the worst case we might just release a new major version and ask customers to migrate their data manually or create an external migration tool. Electron must be updated in any case though

jlipps:
At this point customers love their saved data so much. We might need to first implement an import/export feature for capability json or something

PR #1182, opened 2023-11-07
feat: add .NET NUnit recorder

### LINKED ISSUE #804: Feature request: add Nunit as a recorder test framework
Hi,

The recorder feature in appium inspector is really helpful. 
But I could not utilize this feature, because I am using NUnit as a test framework.

So it would be great if we can have an option for NUnit also.

eglitise:
This PR adds a new recorder for .NET NUnit.
Resolves #804.


eglitise:
cc @Dor-bl

KazuCocoa (approved):
lgtm... while my knowledge is limited

PR #1183, opened 2023-11-08
feat: enhance recorder GUI & fix syntax highlighting

eglitise:
This PR updates the GUI for the recorder functionality:

* Change location from a toggleable panel to a separate tab
  * Prevents confusion between buttons for pausing the recording, closing the panel, and clearing recorded actions
  * Makes it easier to read the code when boilerplate is enabled
  * Avoids reducing the visible height of all other tabs (which sometimes broke the application layout)
* Show language dropdown even when no code has been recorded
* Update some icons and element sizes
  * Also applies to the session code box
* Fix syntax highlighting (not shown in the video)
  * Also applies to the session code box and capabilities box

https://github.com/appium/appium-inspector/assets/37242620/9e479036-ec4e-4906-a5ff-c6c38a5026b7

Additionally, there is a fix for the copy code functionality in the session code box - it was not copying raw code.


jlipps:
Have not yet reviewed code. From a feature perspective I think this is great. My only concern is that when the user is interacting with the app on a different tab (e.g. source tab), there's no feedback that stuff is happening on the recording tab. Some ideas:

- near the "Recording" tab header, keep a badge which is a count of commands recorded, so the user can see it ticking up
- some kind of info message that quickly disappears showing the command added to the recorder

also open to other ideas!

eglitise:
Hmm, I was thinking that having the recorder icon change to red would be enough to indicate that any action (other than gestures) will now generate code. Worst case, the user can always switch to the recorder tab to check.
Badge is probably doable, but counting actions like this does feel somewhat arbitrary.
Info message I'm not too keen on, it could easily get in the way of the user flow.

eglitise:
For another improvement to the recorder GUI, I've fixed the syntax highlighting (this fix also applies to the session code box and capabilities box). Also removed the `format-json` package, whose utilized functionality could be replaced with a one-liner.

mykola-mokhnach (approved):

PR #1200, opened 2023-11-13
chore(deps): remove ky

eglitise:
This PR replaces `ky` with `axios`. There's no need for two HTTP clients, and `ky` was outdated anyway.
Local session discovery was tested to still work.

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -866,18 +866,15 @@ export function getRunningSessions () {
 
     try {
       const adjPath = path.endsWith('/') ? path : `${path}/`;
+      const url = `http${ssl ? 's' : ''}://${hostname}:${port}${adjPath}sessions`;
       const res = username && accessKey
-        ? await axios({
-          url: `http${ssl ? 's' : ''}://${hostname}:${port}${adjPath}sessions`,
-          headers: {
-            'Authorization': `Basic ${btoa(`${username}:${accessKey}`)}`,
-            'content-type': HEADERS_CONTENT
-          }
-        })
-        : await axios({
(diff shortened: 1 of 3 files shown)


app/renderer/actions/Session.js
```diff
@@ -866,17 +866,16 @@ export function getRunningSessions () {
...
-          headers: {
-            'Authorization': `Basic ${btoa(`${username}:${accessKey}`)}`,
-            'content-type': HEADERS_CONTENT
-          }
-        }).json()
-        : await ky(`http${ssl ? 's' : ''}://${hostname}:${port}${adjPath}sessions`, {
-          headers: {'content-type': HEADERS_CONTENT}
-        }).json();
-      dispatch({type: GET_SESSIONS_DONE, sessions: res.value});
+        ? await axios({url, headers: {
```
mykola-mokhnach, line 871: we may only add the necessary header based on the condition:

```
await axios({url, headers: {
  'content-type': HEADERS_CONTENT,
  ...(username && accessKey ? {'Authorization': `Basic ${btoa(`${username}:${accessKey}`)}`} : {})
}});
```

mykola-mokhnach (approved):

jlipps (commented):
I forget why I switched to Ky in the first place. I'm sure there was a reason. But this is no doubt OK.

KazuCocoa:
That was probably to make https://inspector.appiumpro.com/ work. Axios did not work on browser, but worked on nodejs before. It could be wrong right now

eglitise:
Yes, I just checked this in the browser as well (created a session via app, and tried to discover it via browser) - it worked fine.

jlipps:
Ah yes that was why. Well as long as it works in the browser build, we should be fine!

PR #1206, opened 2023-11-16
chore: enhance linting & formatting

eglitise:
This PR enhances the current linting rules and adds reformatting capabilities:
* Remove unused ESLint plugin `eslint-plugin-jsx-a11y `
* Add React-specific ESLint rules from `eslint-plugin-react`
* Add Prettier reformatter
* Add formatting check as part of the CI lint job
* Apply new lint rules and reformatting to the whole repository code (this makes up almost all the code changes in this PR)
  * Code changes also include reordering of imports by using [prettier-plugin-sort-imports](https://github.com/trivago/prettier-plugin-sort-imports). However, the plugin itself is not included here, as it adds even more babel, and requires moving the Prettier configuration to a separate JS file, in order to import this plugin.

I realise this is a huge PR to review, but since most of it is due to automatic reformatting, I think it should be fine to only check all other code changes (ESLint/Prettier configuration, package.json, etc.). I've used multiple commits to make this easier.


KazuCocoa (approved):

.eslintrc.json
```diff
@@ -1,47 +1,26 @@
 {
-  "extends": "@appium/eslint-config-appium",
+  "extends": ["plugin:react/recommended", "@appium/eslint-config-appium"],
```
mykola-mokhnach, line 2: Thanks for doing this. How about switching from babel to ts and doing the corresponding change for the linter as well? Similarly to what was done in https://github.com/appium/appium-mac2-driver/pull/216

typescript compiler does not require everything to be rewritten into TS, but it can also validate vanilla JS files and does it good.

mykola-mokhnach (approved):

.eslintrc.json
eglitise: Thanks, I'll take a look at this in a separate task. There will probably be some differences because babel isn't directly used to build the app, so I'm not sure how that will change.

PR #1223, opened 2023-11-26
Update dev install doc and single test run

ath0mas:
1. Better use `npm ci` than `npm install`, to fully rely on `package-lock.json` and not possibly modify it unintentionally

1. Add note about `node-gyp` required installs for python and c/c++ compiler tools

1. Running `webview-helpers-specs.js` only in my IDE resulted in
```
TypeError: Cannot read properties of undefined (reading 'eql')
```
And I found that it is the only spec that calls `chai.use(chaiAsPromised);` but not `chai.should();`.


eglitise (approved):
Thank you!

PR #1224, opened 2023-11-27
style: remove custom anticon :hover style

ath0mas:
The existing style is mostly [antd](https://github.com/ant-design/ant-design/tree/4.24.15) default blue, and custom sets `anticon` `:hover` to be like red.
But in most buttons with icon only a reduced part is colored like this ; this change ~makes it be applied to the whole button~ removes this custom style.

btn-hover-style

~Looking at [antd mixins for buttons](https://github.com/ant-design/ant-design/blob/4.24.15/components/button/style/mixin.less#L14), I deal with the same pseudo-classes (`:hover`, `:focus`, and `:active`), and then I set the same 3 style properties (`color`, `border-color`, and `background`).~
And this keeps the default style when the button is `disabled`.

Some not directly targeted elements were previously having the custom `:hover` style too, like some arrows for toggle blocks or dropdowns, and also the link icon at the bottom-left.
They all have the default style back as not targeted by the selectors anymore.

### DIFF
--- a/app/renderer/components/Session/Session.css
+++ b/app/renderer/components/Session/Session.css
@@ -17,9 +17,16 @@
   margin-left: 10px;
 }
 
-.sessionContainer :global(.anticon:hover) {
-  background-color: #f08080;
+.sessionContainer :global(.ant-btn-icon-only:not([disabled]):hover),
+.sessionContainer :global(.ant-btn-icon-only:not([disabled]):focus) {
   color: white;
+  border-color: #f08080;
+  background: #f08080;
+}
+.sessionContainer :global(.ant-btn-icon-only:not([disabled]):active) {
+  color: white;
+  border-color: #dc3d32;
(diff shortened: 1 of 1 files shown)


eglitise:
Thanks for bringing this up. While this definitely looks better, it seems that this custom style was only ever used on the session page, so all of the icon-only buttons in the inspector page have always had the default hover style.
For consistency I would instead suggest simply removing this custom style entirely. The default one seems fine IMO, and sticking to defaults would make the eventual migration to antd 5 and dark mode just a bit easier.

ath0mas:
I completely agree with you, I just thought this red style may have been decided on purpose time ago.
I let this breaking choice of removing it up to the maintainers, like you 😉.

So can I remove it?

ath0mas:
OK done, ready.

eglitise (approved):

PR #1259, opened 2023-12-19
fix: adjust highlighter positions again

### LINKED ISSUE #1154: bug: The click point does not match the accurate position of the element in latest version(2023.10.4)

eglitise:
Simple fix for highlighter generation. I had somehow missed that the `deviceScreenSize` values are strings, not integers, whose comparison works differently... (`'1280' > '800' -> false`)
Fixes #1154 (hopefully)

### DIFF
--- a/app/renderer/lib/appium-client.js
+++ b/app/renderer/lib/appium-client.js
@@ -235,7 +235,7 @@ export default class AppiumClient {
         // returned Android height and width can both be affected by UiAutomator2 calculations
         // we stick with device dimensions, but swap them depending on detected orientation
         // deviceScreenSize value fits portrait mode for phones, but landscape mode for tablets
-        const [width, height] = deviceScreenSize.split('x');
+        const [width, height] = deviceScreenSize.split('x').map((param) => parseInt(param, 10));
         // check if the orientation for windowSize matches orientation for deviceScreenSize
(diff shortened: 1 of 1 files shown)


jlipps (commented):
oh good catch. weird that they are strings!

eglitise:
It makes sense that they are strings, after all, `deviceScreenSize` is something like `1280x800`. The Inspector just splits this on the `x` to get the numerical values, but it didn't actually convert these values to integers.

jlipps:
ohhh, I see.

Issue #1255, opened 2023-12-18
bug: Start Session button require another click from time to time


Dor-bl:
### Is this an issue specifically with Appium Inspector?

- [X] I have verified that my issue does _not_ occur with Appium, and should be investigated as an Appium Inspector issue

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently, when I try to start a new session, the click won't work on the first try, and it requires another click.

https://github.com/appium/appium-inspector/assets/59066376/2f19d4ff-6325-4c12-a5e9-2085653cc05e

### Expected Behavior

The session should start on the first try.

### Operating System

Mac

### Appium Inspector Version

2023.11.1

### Appium Version

2.2.1

### Further Information

Is it possible to get logs of the Appium inspector from somewhere?

eglitise:
Hmm, I've never encountered this behavior.
As for the logs, I guess the closest thing could be the browser console (right click anywhere in the Inspecor -> Inspect Element), but I don't think the release version logs as much as the dev version.

Dor-bl:
@eglitise is this log any good?
```
immutableStateInvariantMiddleware.ts:25 Uncaught (in promise) Error: Invariant failed: A state mutation was detected between dispatches, in the path 'session.savedSessions.7.caps.9.name'.  This may cause incorrect behavior. (https://redux.js.org/style-guide/style-guide#do-not-mutate-state)
    at invariant (immutableStateInvariantMiddleware.ts:25)
    at immutableStateInvariantMiddleware.ts:258
    at Object.measureTime (utils.ts:10)
    at immutableStateInvariantMiddleware.ts:251
    at actionCreatorInvariantMiddleware.ts:32
    at dispatch (/Applications/Appium Inspector.app/Contents/Resources/app.asar/node_modules/redux/lib/redux.js:699)
    at Session.js:173
    at Ne (Session.js:216)
    at Session.js:229
    at /Applications/Appium Inspector.app/Contents/Resources/app.asar/node_modules/redux-thunk/lib/index.js:23
    at immutableStateInvariantMiddleware.ts:266
    at actionCreatorInvariantMiddleware.ts:32
    at boundActionCreators.<computed> (/Applications/Appium Inspector.app/Contents/Resources/app.asar/node_modules/react-redux/lib/utils/bindActionCreators.js:13)
    at F (Session.js:39)
    at onClick (Session.js:108)
[...]

PR #1299, opened 2024-01-22
fix: resolve issue with Start Session sometimes needing a second click

orionlize:
I found `start session` would throw error `Invariant failed: A state mutation was detected between dispatches, in the path 'session.caps.0.name'.  This may cause incorrect behavior.` first time. function `loadNewSession` use caps directly would cause this bug.

This pr will fix this bug by `cloneDeep(caps)`

### DIFF
--- a/app/renderer/components/Session/Session.js
+++ b/app/renderer/components/Session/Session.js
@@ -1,6 +1,6 @@
 import {LinkOutlined} from '@ant-design/icons';
 import {Badge, Button, Spin, Tabs} from 'antd';
-import _ from 'lodash';
+import _, {cloneDeep} from 'lodash';
 import React, {useEffect} from 'react';
 import {useNavigate} from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


app/renderer/components/Session/Session.js
```diff
@@ -53,7 +53,7 @@ const Session = (props) => {
   };
 
   const loadNewSession = async (caps, attachSessId = null) => {
-    if (await newSession(caps, attachSessId)) {
+    if (await newSession(cloneDeep(caps), attachSessId)) {
```
eglitise, line 56: It's probably simpler to use `_.cloneDeep` and leave the import statement unchanged.

orionlize:
> Thanks for spotting this! I think this might fix #1255. Could you please also sign the CLA?

done ;-)

eglitise:
Thank you!

Issue #1216, opened 2023-11-21
feature request: increase screenshot size


martamedovova-ext91522:
### Current Behavior

Hello,

the tablet screen display is very small for portrait and landscape mode (especially for iPad 12.9" Pro)
Hard to see the icons or messages - everything.

![image](https://github.com/appium/appium-inspector/assets/130643867/02e95fd6-098e-4d44-b774-c7a01d06d8a3)

### Suggested Solution

Make a larger tablet screen display or add the zoom function

Can you do it?

Thank you :)

eglitise:
Your screenshot is from an older application version. Please try with the latest version (2023.11.1).

eglitise:
But I do think it should be possible to increase the size of the screenshot part in this case. Right now I believe the screenshot height is limited by the application window height, and the screenshot width is limited to 500px - both conditions must be true. In your case, the application window is large enough that the 500px width limit causes a significant limitation.

martamedovova-ext91522:
I cannot increase the size at the Inspector, there is no zoom. 

I will try the new version, hope it is better and more dynamic :-)

luprochazka-cen63872:
I agree with Marta. The screen is too small especially in landscape mode. I think a better way would be to use a floating window (with only device) or enabling resize of panels inside the application.

nicerloop:
Same situation here, we are in the process of automating a desktop application through appium, and the screenshot display with element handles overlay is way too small to be useable.
A floating window with only the device would probably be the most usefull in our case.
Being able to change the relative sizes of the different sections would also be helpful.
Reorganizing the panels could also help.

Any pointers to help me to start working on these modifications?

eglitise:
I did start looking into this and initially just removed the 500px hard limit for screenshot width, so it continued scaling further with the window size. But right now this requires additionally manually refreshing the source/screenshot, which I would like to resolve.

Pointer-wise I'm not really sure; the current resizing approach relies a lot on CSS adjustments, which might be difficult to combine with features like adjustable section widths. But the main resizing method for the screenshot is in [Inspector.js](https://github.com/appium/appium-inspector/blob/main/app/renderer/components/Inspector/Inspector.js#L113).

Also note that some of the other suggestions have already been mentioned in #766.

eglitise:
Inspector version 2024.2.1 increases the screenshot size for larger Inspector window sizes, and also adds the zoom functionality.
Closing this for now - the other suggestions can be covered in #766.

PR #1306, opened 2024-01-29
fix: improve scaling of screenshot width

eglitise:
This PR removes the 500px hard limit for max screenshot width, and instead sets it to 40% of the window width. This allows for much better scaling at larger window widths.
I didn't notice any changes for portrait mode, but in landscape mode, with default window size, the screenshot was 9px wider (meaning the source/selected element is 9px narrower). I don't think this should negatively affect usability.
Should improve the situation for #1216.


mykola-mokhnach (approved):

jlipps (commented):
👍

Issue #391, opened 2022-05-06
pcloudy caps need updating to W3C compatible style


Dcode78:
## Appium or Appium Desktop?

You are reporting an issue at the Appium Desktop repository. Appium Desktop is a wrapper around Appium. If you are having trouble running tests, it is much more likely that the problem you are encountering is not a problem with Appium Desktop but with Appium. For that reason we require you to have tried your testcase with Appium CLI before reporting issues. Check the checkbox below to confirm that you have proven your issue does _not_ reproduce on Appium itself:

- [X] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## The problem

Unable to start the session successfully to pCloudy by configuring desired capabilities 

"Invalid or unsupported WebDriver capabilities found ("pCloudy_Username", "pCloudy_ApiKey"). Ensure to only use valid W3C WebDriver capabilities (see https://w3c.github.io/webdriver/#capabilities).If you run your tests on a remote vendor, like Sauce Labs or BrowserStack, make sure that you put them into vendor specific capabilities, e.g. "sauce:options" or "bstack:options". Please reach out to to your vendor support team if you have further questions."

Couldn't find any way to configure desired capabilities for Username and ApiKey. I end up getting above error every time. 

## Environment

- I am running Appium Desktop version 
   - App Version 2022.5.1
   - Electron  13.6.9
   -  Node.js 14.16.0
                                                  
- I am on (pick one):
    - [X] Windows
    
![pCloudy_Error](https://user-images.githubusercontent.com/32234887/167125764-49aaaf48-b0e0-4185-8db0-b4bee3ca4dc2.jpg)

KazuCocoa:
Probably they should update https://github.com/appium/appium-inspector/blob/bee6ef9effbd125affc82a545b9316551653a27b/app/renderer/actions/Session.js#L389-L390 to set the capability as `pCloud:`(?) prefix.

Could you send an email to their support team, please?

Dcode78:
Dropped a mail at their contact us address, will update once reply received. Thank you for your time @KazuCocoa .

Dcode78:
pCloudy support has responded requesting to use [Appium desktop](https://github.com/appium/appium-desktop/releases/tag/v1.18.2) instead of appium-inspector and I tried it works.

KazuCocoa:
thank you for the confirmation.
Maybe their system hasn't been upgraded? to work with w3c capabilities.

jlipps:
[...]

PR #1324, opened 2024-02-06
docs: address possible removal in cloud providers

KazuCocoa:
I think it would be nice to address possible removal in case they are not compatible with our inspector.
https://github.com/appium/appium-inspector/issues/391

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,10 @@ The Inspector has built-in integrations with various cloud service provider plat
   </tr>
 </table>
 
+> [!NOTE]
+>
+> We may remove cloud providers if they no longer work properly with this inspector, as there is no compatibility.
+
 ## Reporting Issues
 
 If you run into a problem, first check whether you can reproduce it _without_ the Inspector, by using a different Appium client.


eglitise (approved):
Makes sense. I'm also thinking of removing Remote TestKit; their website hasn't worked since around June of last year.

KazuCocoa:
> 'm also thinking of removing Remote TestKit; their website hasn't worked since around June of last year.

btw, @eglitise which website?
It looks like https://appkitbox.com/en/category/news/ has some updates thus I assume they keep maintaining the tool. We could ask the original contributor for in an issue to check it, though. In the Japanese website also had some notifications in 2024.

Screenshot 2024-02-15 at 1 19 37 AM

https://appkitbox.com/category/news/

eglitise:
That... is odd, because whenever I try to access this website (including just now), it returns 403 Forbidden. Even tried multiple browsers and locations. Maybe there's some geoblocking going on that wasn't present early last year?

KazuCocoa:
Hm, possibly yes. I saw that GDPR made some websites shut down from EU countries. Could it be related if your access was from around there

KazuCocoa:
hm, then we probably need to address such limitations somehow (maybe in README...?, no idea right now)

KazuCocoa:
I have confirmed with VPN about the 403. Maybe we could do something

eglitise:
If this is indeed due to GDPR, I think the least they could do is change that HTTP 403 to HTTP 451 😅

PR #1330, opened 2024-02-08
fix: reduce getSession endpoint calls in hybrid mode

KazuCocoa:
W3C WebDriver spec does not support `getSession` endpoint call. chrome session with w3c mode returned 404 for the endpoint recently, then `this.driver.getSession` got 404.
Newer drivers do not need to got to fallback, which needs to call getSession, so lets minimize the endpoint call in `appium-client.js`.

In Session.js, it has a case that MJSONWP mode returns `platform` instead of `platformName` in chrome.

e.g, with:
```
  "appium:chromeOptions": {
    "w3c": false
  }
```

then, `attachedSessionCaps.platformName` returns `null` but `attachedSessionCaps.platform` has expected value. So lets get the platformName with `attachedSessionCaps.platformName || attachedSessionCaps.platform`

### DIFF
--- a/app/renderer/lib/appium-client.js
+++ b/app/renderer/lib/appium-client.js
@@ -277,7 +277,6 @@ export default class AppiumClient {
       await this.driver.switchContext(NATIVE_APP);
     }
 
-    const sessionDetails = await this.driver.getSession();
     const isAndroid = this.driver.client.isAndroid;
 
     // Get all available contexts (or the error, if one appears)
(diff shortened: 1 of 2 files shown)


app/renderer/lib/appium-client.js
```diff
@@ -327,8 +330,11 @@ export default class AppiumClient {
             const deviceScreenInfo = await this.driver.executeScript('mobile:deviceScreenInfo', []);
             webviewLeftOffset = deviceScreenInfo.statusBarSize.height;
           } catch (e) {
-            // in case driver does not support mobile:deviceScreenInfo
-            webviewLeftOffset = sessionDetails.statBarHeight;
+            try {
+              const sessionDetails = await this.driver.getSession();
```
jlipps, line 334: should we memoize this so we don't make the network call twice?

KazuCocoa: Not necessary since these are either Android or iOS, so basically it occurs once

eglitise: It will still be called on every refresh if using hybrid mode - but now only in a few specific cases.

eglitise (approved):
Good catch!

PR #1343, opened 2024-02-12
docs: build docs for appium-inspector@2024.2

eglitise:
I'm not really sure why CI for publishing docs fails on the `mike deploy` call, yet for some reason running `npm run publish:docs` works perfectly fine on my Mac. So for now let's just update the docs this way, until we investigate further.


eglitise:
Good, that seems to have gone smoothly.

jlipps:
also fine with me if you want to publish direct to the docs-site branch instead of opening a PR. it's what i do with appium 😬

jlipps:
Ooops! OK I just ran appium docs publish too. Could it be something like a github deploy token missing in the env?

eglitise:
It's possible... honestly I don't really know how to set it, I assumed since it's the credentials of GitHub Actions, it would already be available. What's interesting is that the publish for 2023.12 went just fine.

PR #1378, opened 2024-03-04
fix: correctly account for position and scroll values of webviews

jlipps:
This PR fixes two things:

1. On iOS, webviews that didn't occupy the full screen (as in Safari) were not previously considered.
2. On both platforms, `window.scrollX` and `window.scrollY` were also not considered, which are necessary when thinking about the positions of web elements on the screen (web element position is absolute, not viewport relative as mobile elements are).

### DIFF
--- a/app/renderer/lib/appium-client.js
+++ b/app/renderer/lib/appium-client.js
@@ -315,26 +315,50 @@ export default class AppiumClient {
           }
         }
       } else if (this.driver.client.isIOS) {
-        // on iOS, find the top status bar and address bar and use its Y endpoint
-        const topBar = await this.fetchElement({
-          strategy: '-ios class chain',
-          selector: IOS_TOP_CONTROLS_SELECTOR,
-        });
-        if (topBar.el) {
-          const {y, height} = await topBar.el.getRect();
-          webviewTopOffset = y + height;
-        }
-        // in landscape mode, there is empty space on both sides (at default zoom level), so add offset for that too
-        if (windowSize.height < windowSize.width) {
-          try {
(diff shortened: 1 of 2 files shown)


KazuCocoa (approved):
Tested a few with a webview app. Looks good to me.

app/renderer/lib/appium-client.js
```diff
@@ -315,26 +315,44 @@ export default class AppiumClient {
...
-          const {y, height} = await topBar.el.getRect();
-          webviewTopOffset = y + height;
-        }
-        // in landscape mode, there is empty space on both sides (at default zoom level), so add offset for that too
-        if (windowSize.height < windowSize.width) {
-          try {
-            const deviceScreenInfo = await this.driver.executeScript('mobile:deviceScreenInfo', []);
-            webviewLeftOffset = deviceScreenInfo.statusBarSize.height;
-          } catch (e) {
+        const isSafari = this.driver.client?.capabilities?.browserName?.toLowerCase() === 'safari';
```
eglitise, line 318: I believe CI is failing because Parcel 1 doesn't support optional chaining. Maybe it's worth adding it as a prop in [`main/app/renderer/reducers/Inspector.js`](https://github.com/appium/appium-inspector/blob/main/app/renderer/reducers/Inspector.js) the same way as `automationName`?

jlipps: yeah just noticed this. i went about the problem in a slightly different way, what do you think

eglitise: That works too :)

eglitise (approved):

PR #1389, opened 2024-03-13
fix: generate suggested locators upon selecting element instead of upon retrieving source

### LINKED ISSUE #1359: bug: improve translateRecursively performance

### Current Behavior

The result is ok, but I saw `translateRecursively`'s slowness when a page source had huge nodes.
For example, a webview context recursively calculated `translateRecursively` 3653 times, then it took `76786` ms entirely.

I roughly calculated this time with:

```
  const startAt = new Date();
  console.log(`=======> translateRecursively`);
  try {
    return firstChild ? translateRecursively(firstChild) : {};
  } finally {
    const endAt = new Date();
    //     =======> done: 3653. duration 76786 ms
    console.log(`=======> done: ${recursiveCount}. duration ${endAt - startAt} ms`);
```

I wondered if we could improve this performance.

I have modified `getSourceUpdate` forcefully with https://gist.github.com/KazuCocoa/07b427765da523e2a488329ba36639d3

### Expected Behavior

not sure if we could improve this performance for now. I haven't dug into this possibility anything yet. I have just detected the long-taken method.

### Operating System

Mac

### Appium Inspector Version

2024.2.2

### Appium Version

_No response_

### Further Information

_No response_

eglitise:
This is a sizeable refactor of the logic behind generating suggested locators.
Until now, locator generation was invoked right after retrieving the app source, and it was run for every single node in the XML source, which is entirely unnecessary. For smaller XMLs, this delay was barely noticeable, but bigger XMLs could show a delay of several seconds or even more. iOS was also affected more than Android, due to the generation of class chains and predicate strings.
With this change, locator generation is invoked only upon selecting an element in an already-retrieved source, and it is run only for that element. In my experience it is also fast enough to not even be noticeable.
As a result of this change, the delay after refreshing the app source is reduced. I didn't do extensive testing, but for one sample app, the delay on Android was reduced by around 100ms, and on iOS by an entire second.
Fixes #1359.

Notable changes:
* Invoke locator generation upon selecting an element instead of after retrieving the app source
* Add a dedicated Redux event for locator generation
* Remove unused `selectedElementVariableName` and `electedElementVariableType` from Redux storage
* Move retrieval of simple locators to `util.js` (which already contained methods for retrieving complex locators)
  * Test spec files were merged accordingly
* Move `findElementByPath` to `util.js` (it did not belong in the reducers file)
* Rename ambiguous instances of `source` to clarify whether it is XML, JSON or Document 
* Use a single `DOMParser` instance instead of creating new ones for every conversion
* Add `findDOMNodeByPath` method for finding a Node by path, similarly to `findElementByPath` for a JSON element
* Update unit tests and add new ones for `findDOMNodeByPath`, `findJSONElementByPath`, `getOptimalClassChain`, `getOptimalPredicateString`
* Enable predicate string for `XCUIElementTypeApplication` ([which is a valid type](https://developer.apple.com/documentation/xctest/xcuielementtype/xcuielementtypeapplication)). Not really sure why class chain did not work, so that was left disabled.


mykola-mokhnach:
I love the optimization idea itself

I would say the whole current implementation of location heuristics might/should be refactored and nicely isolated to separate logical units (classes and methods). It is just a pain to read long methods containing hundreds of lines of code, especially without proper understanding of the background logic.

Would it make sense for you @eglitise if the logic for each locator type would be isolated into classes with shorter methods and cleaner hierarchy?

app/renderer/components/Inspector/SelectedElement.js
```diff
@@ -116,65 +121,28 @@ const SelectedElement = (props) => {
...
 
-  // Add XPath to the data source as well
-  if (xpath) {
-    findDataSource.push({
-      key: 'xpath',
-      find: 'xpath',
-      selector: xpath,
-    });
+  // If XPath is the only optimal selector, warn the user about its brittleness
+  let showXpathWarning = false;
```
mykola-mokhnach, line 143: Sometimes this warning might be annoying. How about having a possibility of hiding it forever or at least until the current inspector session finishes?

app/renderer/components/Inspector/SelectedElement.js
```diff
@@ -116,65 +121,28 @@ const SelectedElement = (props) => {
...
+  if (findDataSource.length === 1) {
+    showXpathWarning = true;
```
mykola-mokhnach, line 145: let showXpathWarning = findDataSource.length === 1

test/unit/util-specs.js
```diff
@@ -1,10 +1,20 @@
...
 import xpath from 'xpath';
 
-import {addVendorPrefixes, getOptimalXPath, xmlToJSON} from '../../app/renderer/util';
+import {
+  addVendorPrefixes,
+  areAttrAndValueUnique,
+  domParser,
+  findDOMNodeByPath,
+  findJSONElementByPath,
+  getOptimalClassChain,
```
mykola-mokhnach, line 12: not sure these methods should all be in util
They all serve particular purpose (finding locators) and it makes sense to have them isolated in a separate module/class (maybe multiple)

mykola-mokhnach: same applies to tests

eglitise: Yeah, I was considering some separation of the utils file as well. Say, one file for the source conversion (XML/JSON/Document), one for the locator generation, and one with everything else (currently only `withTranslation` and `addVendorPrefixes`)

app/renderer/components/Inspector/SelectedElement.js
eglitise: Hmm, I personally don't find it too annoying since it only really appears if only the xpath locator applies. In fact, an unintentional change in this PR is also counting class chain and predicate string towards this length (previously the length was checked _before_ adding these two strategies).

eglitise:
>I would say the whole current implementation of location heuristics might/should be refactored and nicely isolated to separate logical units (classes and methods).

I'm not sure about separating each locator type into its own module/class - really it's only xpath that has multiple methods. But separating all locator generation-specific methods from e.g. the source translation methods sounds very reasonable.

test/unit/util-specs.js
```diff
@@ -16,7 +26,190 @@ function testXPath(doc, node, expectedXPath) {
   xpath.select(expectedXPath, doc)[0].should.equal(node);
 }
 
+// Helper for converting source from XML to Document format
+function xmlToDoc(sourceXML) {
+  return domParser.parseFromString(sourceXML);
+}
+
 describe('util.js', function () {
+  describe('#areAttrAndValueUnique', function () {
```
KazuCocoa, line 35: 👍

KazuCocoa (approved):
lgtm, overall. Thank you for this change, I tried to move the timing into select element, but did not have much time to understand and track the implementation... I tested with large source like I attached in the original issue.

I hope this will improve inspector users' experience after the page source loading

app/renderer/utils/locator-generation.js
```diff
@@ -30,72 +16,103 @@ const UNIQUE_CLASS_CHAIN_ATTRIBUTES = ['name', 'label', 'value'];
...
 
-    const result = [];
-    for (let childIdx = 0; childIdx < xmlNode.childNodes.length; ++childIdx) {
-      const childNode = xmlNode.childNodes.item(childIdx);
-      if (childNode.nodeType === 1) {
-        result.push(childNode);
+/**
+ * Get suggested selectors for simple locator strategies (which match a specific attribute)
+ *
+ * @param {Object} attributes element attributes
```
mykola-mokhnach, line 49: this typedef could be improved

app/renderer/utils/locator-generation.js
```diff
@@ -30,72 +16,103 @@ const UNIQUE_CLASS_CHAIN_ATTRIBUTES = ['name', 'label', 'value'];
...
+ * @param {Document} sourceDoc
+ * @param {boolean} isNative whether native context is active
+ * @returns {Object} mapping of strategies to selectors
```
mykola-mokhnach, line 52: same as above

mykola-mokhnach (approved):

jlipps:
Didn't make it here in time to do a full review, but this is an extremely useful improvement! Thank you

PR #1394, opened 2024-03-18
feat: add suggested selector for UiAutomator strategy

eglitise:
This PR adds generation of suggested selectors for the UiAutomator locator strategy, when using the UiAutomator2 driver.
While it does not cover the full breadth of capabilities for this strategy (e.g. no UiScrollable or child selectors), it can still assemble selectors out of 4 attributes (`resource-id`, `text`, `content-desc`, `class`), and add indices, which should cover most use cases.

It is also oftentimes faster than the `id` strategy... 👀 
image

⚠️ Important to note that this strategy seems to only find elements _in the last direct child of the hierarchy_ - elements under any other direct child are not visible. I assume this is some sort of limitation (whether intentional or not) from Google's end. This is why the implementation requires recalculating the full DOM Document, the DOM node, and path.

There is another related change in this PR: when reading the suggested locator table, the entry `-android uiautomator (docs)` was quite lengthy and difficult to read, therefore the presentation of entries in the selected element tables was adjusted. Until now, some columns had a fixed width, and others took up the remaining width, potentially resulting in linebreaks - this could look very unreadable at small window sizes:
image
Now the rows support horizontal scrolling, while still keeping the previously-fixed-width columns in place:
image
This does mean that very long selectors (e.g. complex xpaths/class chains) will require scrolling even for large screen sizes. I think this is still acceptable.

### DIFF
--- a/app/renderer/utils/locator-generation.js
+++ b/app/renderer/utils/locator-generation.js
@@ -78,10 +78,11 @@ export function getSimpleSuggestedLocators(attributes, sourceDoc, isNative = tru
  *
  * @param {string} path a dot-separated string of indices
  * @param {Document} sourceDoc
+ * @param {boolean} isNative whether native context is active
  * @param {string} automationName
  * @returns {Object.<string, string>} mapping of strategies to selectors
  */
-export function getComplexSuggestedLocators(path, sourceDoc, automationName) {
+export function getComplexSuggestedLocators(path, sourceDoc, isNative, automationName) {
   let complexLocators = {};
   const domNode = findDOMNodeByPath(path, sourceDoc);
   if (domNode.tagName.includes('XCUIElement')) {
@@ -89,7 +90,7 @@ export function getComplexSuggestedLocators(path, sourceDoc, automationName) {
     const optimalClassChain = getOptimalClassChain(sourceDoc, domNode);
     complexLocators['-ios class chain'] = optimalClassChain ? '**' + optimalClassChain : null;
     complexLocators['-ios predicate string'] = getOptimalPredicateString(sourceDoc, domNode);
-  } else if (automationName.toLowerCase() === 'uiautomator2') {
+  } else if (automationName.toLowerCase() === 'uiautomator2' && isNative) {
     complexLocators['-android uiautomator'] = getOptimalUiAutomatorSelector(
       sourceDoc,
       domNode,
@@ -121,6 +122,7 @@ export function getSuggestedLocators(selectedElement, sourceXML, isNative, autom
   const complexLocators = getComplexSuggestedLocators(
     selectedElement.path,
     sourceDoc,
+    isNative,
     automationName,
   );
   return _.toPairs({...simpleLocators, ...complexLocators});
--- a/app/renderer/utils/locator-generation.js
+++ b/app/renderer/utils/locator-generation.js
@@ -54,10 +54,10 @@ export function areAttrAndValueUnique(attrName, attrValue, sourceDoc) {
 /**
  * Get suggested selectors for simple locator strategies (which match a specific attribute)
  *
- * @param {Object.<string, string|number>} attributes element attributes
+ * @param {Record<string, string|number>} attributes element attributes
  * @param {Document} sourceDoc
  * @param {boolean} isNative whether native context is active
- * @returns {Object.<string, string>} mapping of strategies to selectors
+ * @returns {Record<string, string>} mapping of strategies to selectors
  */
 export function getSimpleSuggestedLocators(attributes, sourceDoc, isNative = true) {
   const res = {};
@@ -80,22 +80,29 @@ export function getSimpleSuggestedLocators(attributes, sourceDoc, isNative = tru
  * @param {Document} sourceDoc
  * @param {boolean} isNative whether native context is active
  * @param {string} automationName
- * @returns {Object.<string, string>} mapping of strategies to selectors
+ * @returns {Record<string, string>} mapping of strategies to selectors
  */
 export function getComplexSuggestedLocators(path, sourceDoc, isNative, automationName) {
   let complexLocators = {};
   const domNode = findDOMNodeByPath(path, sourceDoc);
-  if (domNode.tagName.includes('XCUIElement')) {
-    // XCUI native context
-    const optimalClassChain = getOptimalClassChain(sourceDoc, domNode);
-    complexLocators['-ios class chain'] = optimalClassChain ? '**' + optimalClassChain : null;
-    complexLocators['-ios predicate string'] = getOptimalPredicateString(sourceDoc, domNode);
-  } else if (automationName.toLowerCase() === 'uiautomator2' && isNative) {
-    complexLocators['-android uiautomator'] = getOptimalUiAutomatorSelector(
-      sourceDoc,
-      domNode,
-      path,
-    );
+  if (isNative) {
+    switch (automationName.toLowerCase()) {
+      case 'xcuitest':
+      case 'mac2': {
+        const optimalClassChain = getOptimalClassChain(sourceDoc, domNode);
+        complexLocators['-ios class chain'] = optimalClassChain ? '**' + optimalClassChain : null;
+        complexLocators['-ios predicate string'] = getOptimalPredicateString(sourceDoc, domNode);
+        break;
+      }
+      case 'uiautomator2': {
+        complexLocators['-android uiautomator'] = getOptimalUiAutomatorSelector(
+          sourceDoc,
+          domNode,
+          path,
+        );
+        break;
+      }
+    }
   }
   complexLocators.xpath = getOptimalXPath(sourceDoc, domNode);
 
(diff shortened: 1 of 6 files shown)


mykola-mokhnach:
We may also show the full selector properly wrapped in a tooltip upon moseover cell event

app/renderer/utils/locator-generation.js
```diff
@@ -70,16 +78,24 @@ export function getSimpleSuggestedLocators(attributes, sourceDoc, isNative = tru
  *
  * @param {string} path a dot-separated string of indices
  * @param {Document} sourceDoc
+ * @param {boolean} isNative whether native context is active
+ * @param {string} automationName
  * @returns {Object.<string, string>} mapping of strategies to selectors
  */
-export function getComplexSuggestedLocators(path, sourceDoc) {
+export function getComplexSuggestedLocators(path, sourceDoc, isNative, automationName) {
```
mykola-mokhnach, line 85: mathods having more than 3 arguments are hard to read. Maybe we could combine some args into an object

app/renderer/utils/locator-generation.js
```diff
@@ -70,16 +78,24 @@ export function getSimpleSuggestedLocators(attributes, sourceDoc, isNative = tru
...
   let complexLocators = {};
   const domNode = findDOMNodeByPath(path, sourceDoc);
   if (domNode.tagName.includes('XCUIElement')) {
```
mykola-mokhnach, line 88: it would make sense to distinguish branches here based on the automation name

app/renderer/utils/locator-generation.js
```diff
@@ -70,16 +78,24 @@ export function getSimpleSuggestedLocators(attributes, sourceDoc, isNative = tru
  *
  * @param {string} path a dot-separated string of indices
  * @param {Document} sourceDoc
+ * @param {boolean} isNative whether native context is active
+ * @param {string} automationName
  * @returns {Object.<string, string>} mapping of strategies to selectors
```
mykola-mokhnach, line 83: Record<string, string>

app/renderer/utils/locator-generation.js
mykola-mokhnach: same below

app/renderer/utils/locator-generation.js
```diff
@@ -417,12 +431,96 @@ export function getOptimalPredicateString(doc, domNode) {
...
+export function getOptimalUiAutomatorSelector(doc, domNode, path) {
+  try {
+    // BASE CASE #1: If this isn't an element, or we're above the root, return empty string
+    if (!domNode.tagName || domNode.nodeType !== 1) {
+      return '';
+    }
 
+    // UiAutomator can only find elements inside the last direct child of the hierarchy.
+    // hierarchy is the child of doc (which is <xml/>), so need to get the children of its child
+    const hierarchyChildren = childNodesOf(childNodesOf(doc)[0]);
```
mykola-mokhnach, line 458: is the length of childNodesOf(doc) always > 0?

app/renderer/utils/locator-generation.js
```diff
@@ -417,12 +431,96 @@ export function getOptimalPredicateString(doc, domNode) {
...
+    const newPath = pathArray.join('.');
+    const newDomNode = findDOMNodeByPath(newPath, newDoc);
+
+    // BASE CASE #3: Check all attributes and try to find unique ones
+    let uiSelector, othersWithAttr, mostUniqueSelector;
+    let othersWithAttrMinCount = 99;
+
+    for (const [attrName, attrTranslation] of CHECKED_UIAUTOMATOR_ATTRIBUTES) {
+      const attrValue = newDomNode.getAttribute(attrName);
+      if (_.isNil(attrValue) || (_.isString(attrValue) && attrValue.length === 0)) {
```
mykola-mokhnach, line 487: you could use _.isEmpty

app/renderer/utils/locator-generation.js
mykola-mokhnach: also the result of childNodesOf might be null according to the docstring

app/renderer/utils/locator-generation.js
```diff
@@ -70,16 +78,24 @@ export function getSimpleSuggestedLocators(attributes, sourceDoc, isNative = tru
...
-export function getComplexSuggestedLocators(path, sourceDoc) {
+export function getComplexSuggestedLocators(path, sourceDoc, isNative, automationName) {
   let complexLocators = {};
   const domNode = findDOMNodeByPath(path, sourceDoc);
   if (domNode.tagName.includes('XCUIElement')) {
     // XCUI native context
     const optimalClassChain = getOptimalClassChain(sourceDoc, domNode);
     complexLocators['-ios class chain'] = optimalClassChain ? '**' + optimalClassChain : null;
     complexLocators['-ios predicate string'] = getOptimalPredicateString(sourceDoc, domNode);
+  } else if (automationName.toLowerCase() === 'uiautomator2' && isNative) {
```
KazuCocoa, line 93: [nits]
Could refer to https://github.com/appium/appium-inspector/blob/0405a47f764c5a3d637ac6dcf105fb1da6956c3e/app/renderer/components/Inspector/shared.js#L90 to avoid using writing `uiautomator2` directly? Possbly the shared components need to move to another place?

eglitise: I agree with splitting/moving `shared.js` elsewhere, but this would involve other files unrelated to this PR, so I would like to do that in a separate PR.

app/renderer/utils/locator-generation.js
eglitise: Among all 3 related functions (`getSuggestedLocators`, `getSimpleSuggestedLocators`, `getComplexSuggestedLocators`), there is only one common parameter (`isNative`). 4 parameters isn't _that_ bad in my opinion.

eglitise:
>We may also show the full selector properly wrapped in a tooltip upon mouseover cell event

Unfortunately this does not appear to work in combination with the existing onClick tooltip. Another idea I had is to wrap the selector only after a certain width, but this would still make it quite difficult to read it for small window sizes.

eglitise:
Also, while I've been using UiAutomator selectors for some time now, I'm still not sure whether the `new UiSelector().` part is necessary, because `resourceId("android:id/content")` works just fine.

app/renderer/utils/locator-generation.js
mykola-mokhnach: I would say there is still a lot space for [improvements](https://www.baeldung.com/java-best-practices-many-parameters-method). Ofc, not necessarily as part of this PR.

mykola-mokhnach (approved):

app/renderer/utils/locator-generation.js
```diff
@@ -391,38 +415,126 @@ export function getOptimalPredicateString(doc, domNode) {
...
+    const newXml = xmlSerializer.serializeToString(lastHierarchyChild);
+    // wrap the new XML in a dummy tag which will have the node type Document
+    const newDoc = domParser.parseFromString(`<dummy>${newXml}</dummy>`);
+    pathArray[0] = '0';
+    const newPath = pathArray.join('.');
+    const newDomNode = findDOMNodeByPath(newPath, newDoc);
+
+    // BASE CASE #4: Check all attributes and try to find unique ones
+    let uiSelector, othersWithAttr, mostUniqueSelector;
+    let othersWithAttrMinCount = 99;
```
mykola-mokhnach, line 497: what is this magic number?

mykola-mokhnach:
> Also, while I've been using UiAutomator selectors for some time now, I'm still not sure whether the `new UiSelector().` part is necessary, because `resourceId("android:id/content")` works just fine.

- https://github.com/appium/appium-uiautomator2-server/blob/master/app/src/test/java/io/appium/uiautomator2/utils/UiScrollableParserTests.java
- https://github.com/appium/appium-uiautomator2-server/blob/master/app/src/test/java/io/appium/uiautomator2/utils/UiSelectorParserTests.java
- https://github.com/appium/appium-uiautomator2-server/blob/master/app/src/test/java/io/appium/uiautomator2/utils/UiAutomatorParserTests.java

KazuCocoa (approved):

app/renderer/utils/locator-generation.js
eglitise: Oh certainly - a lot of the Inspector code is very much in need of better code quality.

app/renderer/utils/locator-generation.js
eglitise: I guess `Number.MAX_SAFE_INTEGER` would make the intent more obvious. Another approach would be to skip the initial value here, and assign it in the following comparison if it is undefined.

eglitise:
> https://github.com/appium/appium-uiautomator2-server/blob/master/app/src/test/java/io/appium/uiautomator2/utils/UiScrollableParserTests.java
https://github.com/appium/appium-uiautomator2-server/blob/master/app/src/test/java/io/appium/uiautomator2/utils/UiSelectorParserTests.java
https://github.com/appium/appium-uiautomator2-server/blob/master/app/src/test/java/io/appium/uiautomator2/utils/UiAutomatorParserTests.java

I looked through these and it still isn't fully clear to me. It looks like `createUiSelectorParser` is used for this purpose, but I couldn't figure out its exact implementation in `UiExpressionParser` 😅 but that's fine.

PR #1401, opened 2024-03-22
fix: rename TestingBot key and secret to username and accessKey

jochen-testingbot:
This PR renames the TestingBot `key` and `secret` params to `username` and `accessKey`.

The reason for this change is because `getRunningSessions` checks if both `username` and `accessKey` are set to determine if an `Authorization` header needs to be sent with the Axios GET request to `/sessions`.

https://github.com/appium/appium-inspector/blob/main/app/renderer/actions/Session.js#L935-L944


eglitise (approved):
Thank you!

PR #1402, opened 2024-03-25
chore: add dedicated files for constants

eglitise:
This PR is a refactor with the following changes:
* Add a new `constants` directory for storing various constants used throughout the app
* Get rid of the `shared.js` and `AntdTypes.js` files whose purpose was not obvious
* Save the `automationName` value in lowercase, avoiding the need to change it to lowercase in other files


jlipps (approved):
seems reasonable to me!

app/renderer/components/ErrorBoundary/ErrorMessage.js
```diff
@@ -2,13 +2,12 @@ import {CopyOutlined} from '@ant-design/icons';
 import {Alert, Button, Tooltip} from 'antd';
 import React from 'react';
 
+import {ALERT} from '../../constants/ANTD_TYPES';
+import {LINKS} from '../../constants/COMMON';
```
mykola-mokhnach, line 6: Not sure we need to have uppercase file names. Is it mentioned in any standard?

app/renderer/utils/other.js
```diff
@@ -28,3 +28,36 @@ export function addVendorPrefixes(caps) {
...
+}
+
+export function percentageToPixels(pct, maxPixels) {
+  if (!isNaN(pct)) {
+    return Math.round(maxPixels * (pct / 100));
+  }
+}
+
+export function parseCoordinates(element) {
+  let {bounds, x, y, width, height} = element.attributes || {};
```
mykola-mokhnach, line 45: why cannot we use const?

eglitise: I didn't modify these methods, but some variables can indeed be converted to consts. Not these though - only `bounds` is not changed. We can however just create new const variables, allowing to change these to consts too.

app/renderer/components/ErrorBoundary/ErrorMessage.js
eglitise: I saw this as a suggestion when searching for best practices for constants, and I thought it made sense. But it's probably not specified in any standard, no.

mykola-mokhnach: lets only use lowercase for file names. It is already clear these are constants as they are located in the `constants` subdir

KazuCocoa (approved):
lg

mykola-mokhnach (approved):

PR #1404, opened 2024-03-28
ci: update electron-builder and Azure pipeline configurations

### LINKED ISSUE #1400: bug: Unable to launch .AppImage from arm64 Ubuntu 23.10

### Current Behavior

When launching AppImage with command ./Appium-Inspector-linux-2024.3.1.AppImage, error is received:
```./Appium-Inspector-linux-2024.3.1.AppImage: 1: �: not found
./Appium-Inspector-linux-2024.3.1.AppImage: 2: }�P7pys�H��_f��h: not found
./Appium-Inspector-linux-2024.3.1.AppImage: 1: �b�@@: not found
./Appium-Inspector-linux-2024.3.1.AppImage: 1: ELFAI: not found
./Appium-Inspector-linux-2024.3.1.AppImage: 19: Syntax error: Unterminated quoted string```

File downloaded from: [releases - 2024.3.1](https://github.com/appium/appium-inspector/releases)
Instructions followed from: [Installation Instructions](https://appium.github.io/appium-inspector/latest/quickstart/installation/)


### Expected Behavior

Application should launch

### Operating System

Linux

### Appium Inspector Version

2024.3.1

### Appium Version

2.5.1

### Further Information

Device is Raspberry Pi 5
Running Ubuntu 23.10
Device is up to date, Appium is function correctly (able to connect to devices and control)

### PR DESCRIPTION
**Pull request by eglitise**
This PR updates the configuration for `electron-builder` (which generates OS-specific application packages) and the Azure pipeline that is triggered during the release process (which is where `electron-builder` is run):
* Update Azure VMs to latest OS images
* Bump NodeTool@0 & Node 16 to UseNode@1 & Node 20
* Move `electron-builder` config to a separate file for clarity
* Generate x64 and arm64 builds for all 3 OSes
  * Remove Windows ia32 builds
  * Split macOS universal artifacts into x64 and arm64 artifacts (smaller download)
* Use common naming format for all artifacts
  * Note that Linux x64 AppImage is for some reason generated with the `x84_64` suffix instead of `x64`
* Add `tar.gz` artifacts for Linux
* Add entitlements file for macOS (likely only relevant after updating to Electron 20+)

As a result of these changes, I believe the number of generated artifacts will now increase to 26.
Resolves #1400.

### CODE REVIEW

**Review by jlipps — APPROVED** [review]
looks great!

File: ci-jobs/templates/package.yml [hunk 1]
```diff
@@ -21,9 +21,9 @@ jobs:
     steps:
       - checkout: self
         persistCredentials: true
-      - task: NodeTool@0
+      - task: UseNode@1
```
mykola-mokhnach, line 24: consider using lts  node version alias instead of the hardcoded one

eglitise: I'm not sure if Azure supports the `lts` value. [The documentation](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/use-node-v1?view=azure-pipelines#inputs) only refers to semver values.

mykola-mokhnach: Oh, I forgot it's Azure. Perhaps it makes sense to migrate to GitHub CI

eglitise: I agree that GitHub CI would be more convenient, though I'm guessing there was a reason why Azure was originally selected. Maybe the machines are more powerful? It's also used in other Appium projects (e.g. xcuitest driver).

mykola-mokhnach: it was selected before github had any CI. Yes, it is still used, although not by xcuitest driver. It's used in webdriveragent because of the high amount of e2e tests we run there and the limited capability of free macOS-based runners on Github. Otherwise I would also migrate them.

KazuCocoa: Perhaps I tried to convert this project to GHA before. Then, I needed maybe Window's signature which probably needed Dan's help. We could restart migrating them.

PR #1407, opened 2024-04-03
fix: fix updater

### LINKED ISSUE #733: Updating Appium inspector with app is not working
## Appium or Appium Desktop?

- [ ] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## The problem

_Replace this section with a description of your issue_

## Environment

- I am on (pick one):
    - [* ] Mac
    - [ ] Windows
    - [ ] Linux

When I tried to update Appium inspector on my Mac (macOS 13.2.1) it did not work. This happened on the apple silicon computer and the Intel computer. The only way I could update the program was to download the new version of the application.

eglitise:
This PR fixes the long-outstanding issue with the updater not working. It reverts from the custom implementation based on the stock Electron updater, to a simpler version based on `electron-updater`.

For now the implementation only works through the manual 'Check for Updates' button, but it should not be too difficult to also add automatic checking in the future, once it's certain that the manual approach works fine.

Most of the code related to the old updater implementation has also been removed.

I did also test this functionality. This was done by changing the application version in `package.json` to a smaller number, packaging it into an installer, installing it, and then checking for updates. On macOS I ran into an error when clicking restart after the 'newer' version was downloaded - I believe this is likely due to different code signing identities. On Windows the process worked fine, although there was an odd ENOENT warning upon launching the updated version, which did not appear upon reopening the app - maybe related to different `publish` options.

Despite the above problems, I think the change is still an improvement and is worth merging in its current state. Perhaps the issues will disappear when updating between two production versions.

Fixes #733.


app/main/updater.js
```diff
@@ -0,0 +1,56 @@
...
+autoUpdater.on('update-available', async ({version, releaseDate}) => {
+  const pubDate = new Date(releaseDate).toDateString();
+  const {response} = await dialog.showMessageBox({
+    type: 'info',
+    message: t('appiumIsAvailable', {name: version}),
+    buttons: [t('Install Now'), t('Install Later')],
+    detail: t('updateDetails', {pubDate, notes: RELEASES_LINK}),
+  });
+  if (response === 0) {
+    dialog.showMessageBox({
```
mykola-mokhnach, line 33: not sure if await is missing

app/main/menus.js
```diff
@@ -67,7 +65,7 @@ function optionAbout() {
 function optionCheckForUpdates() {
   return {
     label: t('Check for Updates…'),
-    click: () => checkNewUpdates(true),
+    click: () => checkForUpdates(),
```
mykola-mokhnach, line 68: maybe it makes sense to add an option to check for updates on startup automatically (which is enabled by default). This way we make sure users have the latest and the greatest version

app/main/updater.js
```diff
@@ -0,0 +1,56 @@
...
+  });
+});
+
+autoUpdater.on('update-available', async ({version, releaseDate}) => {
+  const pubDate = new Date(releaseDate).toDateString();
+  const {response} = await dialog.showMessageBox({
+    type: 'info',
+    message: t('appiumIsAvailable', {name: version}),
+    buttons: [t('Install Now'), t('Install Later')],
+    detail: t('updateDetails', {pubDate, notes: RELEASES_LINK}),
```
mykola-mokhnach, line 30: would it be possible to show release notes for the particular version?

app/main/updater.js
```diff
@@ -0,0 +1,56 @@
...
+  });
+  if (response === 0) {
+    dialog.showMessageBox({
+      type: 'info',
+      buttons: [t('OK')],
+      message: t('updateIsBeingDownloaded'),
+    });
+    autoUpdater.downloadUpdate();
```
mykola-mokhnach, line 38: is this done silently in background?

app/main/menus.js
eglitise: I would like to first have a version with just manual updating to confirm that it works, and then enable automatic update checks in the next version.

app/main/updater.js
eglitise: Yes. However, unlike stock Electron, `electron-updater` has the `download-progress` event, which could theoretically be used to show the progress percentage. But from a quick glance, it would require creating a dedicated window and sending the progress text as a message event to this window. I think it's still doable, but perhaps not as critical to have at the moment.

app/main/updater.js
eglitise: I did investigate this option, but decided against it for several reasons:
* The raw release notes string requires post-processing to make it readable (remove HTML tags with regex, handle multi-level bullet points, etc)
* The release notes can get quite large and cause the floating box buttons to go out of bounds, preventing any user interaction. I mainly observed this with the raw release notes string, but there is still a possibility that it can apply to the post-processed notes in the future. A quick search for scrolling options inside this box also didn't give me any results.
* The user may skip one or more versions during the update, and showing the changelog only for the most recent version is not sufficient, in my opinion

Still, the floating box does have another problem - it cannot render clickable links, so the link shown here is unfortunately just text.

app/main/updater.js
mykola-mokhnach: makes sense

app/main/updater.js
mykola-mokhnach: if that is an async call then the next call will happen anyway unless it is put into `then` callback

eglitise: I tested this just now with the await added - the download does not actually start until the dialog is dismissed. I believe the await causes Electron to wait for user input (in this case, dismissal of the dialog box)

mykola-mokhnach: yep, that is why I wanted to know if you want to start the update download progress only after the dialog is accepted or asynchronously while it's shown

eglitise: I think it's better to start it asynchronously - from a user perspective, the message box text implies that the download has already started.

mykola-mokhnach: then this code is OK. maybe just add a comment that this behaviour is expected

mykola-mokhnach (approved):

jlipps (commented):
Incredible fix for this issue, thank you!

Issue #1440, opened 2024-04-29
feature request: When copying an accessibility ID that contains a newline, please use "\n" instead of the actual newline character. This will make it easier to paste the ID directly into code. 


miwtoo:
### Current Behavior

```
Setting
Tab 3 of 3
```

### Suggested Solution

```
Setting\nTab 3 of 3
```

### Additional Information
When copying an accessibility ID that contains a newline, please use "\n" instead of the actual newline character. This will make it easier to paste the ID directly into code. 

<img width="516" alt="image" src="https://github.com/appium/appium-inspector/assets/12011082/c858de31-dd90-4898-ad9c-376b8eb1382a">

accessibility id : 
<img width="1212" alt="image" src="https://github.com/appium/appium-inspector/assets/12011082/038ae6d6-7e58-4160-8a58-3e368db72d1e">

miwtoo:
@eglitise 
```XML
<?xml version='1.0' encoding='UTF-8' standalone='yes' ?>
<hierarchy index="0" class="hierarchy" rotation="0" width="1008" height="2077">
  <android.widget.FrameLayout index="0" package="com.example.automation_appium" class="android.widget.FrameLayout" text="" checkable="false" checked="false" clickable="false" enabled="true" focusable="false" focused="false" long-clickable="false" password="false" scrollable="false" selected="false" bounds="[0,0][1008,2244]" displayed="true">
    <android.widget.LinearLayout index="0" package="com.example.automation_appium" class="android.widget.LinearLayout" text="" checkable="false" checked="false" clickable="false" enabled="true" focusable="false" focused="false" long-clickable="false" password="false" scrollable="false" selected="false" bounds="[0,0][1008,2190]" displayed="true">
      <android.widget.FrameLayout index="0" package="com.example.automation_appium" class="android.widget.FrameLayout" text="" resource-id="android:id/content" checkable="false" checked="false" clickable="false" enabled="true" focusable="false" focused="false" long-clickable="false" password="false" scrollable="false" selected="false" bounds="[0,0][1008,2190]" displayed="true">
        <android.widget.FrameLayout index="0" package="com.example.automation_appium" class="android.widget.FrameLayout" text="" checkable="false" checked="false" clickable="false" enabled="true" focusable="true" focused="true" long-clickable="false" password="false" scrollable="false" selected="false" bounds="[0,0][1008,2190]" displayed="true">
[...]

PR #1442, opened 2024-04-29
feat: :sparkles: support element that has newline replace to \n (#1440)

miwtoo:
I created a pull request for feature request #1440. I would appreciate if you could please review it and provide any suggestions.

Screenshot 2567-04-30 at 00 31 52

### DIFF
--- a/app/renderer/components/Inspector/SelectedElement.js
+++ b/app/renderer/components/Inspector/SelectedElement.js
@@ -39,18 +39,19 @@ const SelectedElement = (props) => {
   const isDisabled = selectedElementSearchInProgress || isFindingElementsTimes;
 
   const selectedElementTableCell = (text, copyToClipBoard) => {
+    const textString = String(text);
     if (copyToClipBoard) {
       return (
         <div className={styles['selected-element-table-cells']}>
           <Tooltip title={t('Copied!')} trigger="click">
-            <span className={styles['element-cell-copy']} onClick={() => clipboard.writeText(text)}>
-              {text}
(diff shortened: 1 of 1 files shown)


eglitise (changes requested):
Thanks for the quick PR!
While the code seems reasonable to me, this does not fix how the text is displayed in the application source tree. I think a proper fix might require changes in the `xmlToJSON` method instead.
(edit: I would also call this a fix rather than a feature, since the XML is correct, but the Inspector GUI is not)

miwtoo:
> Thanks for the quick PR! While the code seems reasonable to me, this does not fix how the text is displayed in the application source tree. I think a proper fix might require changes in the `xmlToJSON` method instead. (edit: I would also call this a fix rather than a feature, since the XML is correct, but the Inspector GUI is not)

@eglitise 
Screenshort: 
Screenshot 2567-04-30 at 04 02 14

test/unit/utils-source-parsing-specs.js
```diff
@@ -470,5 +470,111 @@ describe('utils/source-parsing.js', function () {
         path: '',
       });
     });
+
+    it('should convert xml to json for Newline', function () {
```
eglitise, line 474: Thanks for adding a new test as well! I think that since this test is specifically for the newline conversion functionality, you can probably remove all other attributes that don't have the newline symbols, which would keep the test shorter.

eglitise (approved):

test/unit/utils-source-parsing-specs.js
miwtoo: removed. Thanks, @eglitise for the advice.

KazuCocoa (approved):

eglitise (approved):
Thank you!

jlipps:
thanks @miwtoo !

PR #1444, opened 2024-04-30
chore: use .jsx and .module.css file extensions where appropriate

eglitise:
This PR is a preparatory step towards (finally) migrating away from Parcel v1.

I've done a bit of experimenting on a fresh Electron+React project with Parcel v2, and it did not look great. Their Electron integration is unfortunately not as good as it could be, seemingly requiring a rewrite of some parts into CJS (though I didn't fully figure this part out).
Then I tried Webpack, but it required _a lot_ of tinkering with its config files - too much for my limited knowledge.
Finally I tried Vite... and I actually managed to get it to work, for both web and Electron formats. So I am proposing **to drop Parcel entirely and instead migrate to Vite**.

I already started working on this migration, and managed to get the web version fully working (with one exception for i18n). But, as part of getting this to work, Vite required changing the extensions for React and CSS module files. Since this affected a lot of files, I thought of separating the changes into its own PR - and after some brief testing with Parcel v1, I did not observe any issues.


mykola-mokhnach (approved):
the change looks safe to me

jlipps:
I'm OK with choosing Vite over Parcel v2!

PR #1447, opened 2024-05-01
chore: refactor

eglitise:
This PR continues the preparation for migrating to Vite, and primarily changes many file locations:
* add a new `assets` directory in `renderer`, and move `images`, `locales` and `stylesheets` directories to it
* move `main.js` to `main` directory
* merge `Root.jsx` and `routes.jsx` (originally I only moved `Root.jsx` to its parent directory, but merging it with `routes.jsx` further improved clarity)
* add a new `src` directory in `renderer` and move all `renderer` contents to it, except HTML entrypoint files

Browser and Electron versions were tested to still work fine.


jlipps (approved):
seems good to me! we'll want to double check the CrowdIn integration works without issue after we merge.

eglitise:
I think it should be fine, since the path for translation files was only defined in one file.

Issue #1478, opened 2024-05-29
bug: Boilerplate python code has undefined item in options.load_capabilities , created python code won't run.


CEN8888:
### Is this an issue specifically with Appium Inspector?

- [X] I have verified that my issue does _not_ occur with Appium, and should be investigated as an Appium Inspector issue

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I am using appium inspector with a windows app .   point to a windows application . Observe the window opens . Press record to generate python code .  Open  the record tab and display the boilerplate code and recorded code .

observe boiler code includes undefined capabilities and will not run as expected .

# This sample code supports Appium Python client >=2.3.0
# pip install Appium-Python-Client
# Then you can paste this into a file and simply run with Python

from appium import webdriver
from appium.options.common.base import AppiumOptions
from appium.webdriver.common.appiumby import AppiumBy

# For W3C actions
from selenium.webdriver.common.action_chains import ActionChains
from selenium.webdriver.common.actions import interaction
from selenium.webdriver.common.actions.action_builder import ActionBuilder
from selenium.webdriver.common.actions.pointer_input import PointerInput

options = AppiumOptions()
options.load_capabilities({
	"platformName": "windows",
	"appium:automationName": "windows",
	"appium:app": "C:*.exe",
	"appium:undefined": undefined,
	"appium:newCommandTimeout": 3600,
	"appium:connectHardwareKeyboard": True
})

driver = webdriver.Remote("http://127.0.0.1:4723", options=options)

el1 = driver.find_element(by=AppiumBy.XPATH, value="//Pane/Group/Group")
el1.click()

driver.quit()

### Expected Behavior

The boilerplate and recorded code should run as a python program. 

### Operating System

Windows

### Appium Inspector Version

2024.3.4

### Appium Version

2.5.4

### Further Information

Winappdriver version 1.2.2009.02003  Appium-Python-Client 2.5.0   selenium 4.20.0

KazuCocoa:
Btw, the appium python client and selenium version combination need to follow https://github.com/appium/python-client?tab=readme-ov-file#compatibility-matrix

eglitise:
Could you please provide your Python version, and the exact error message returned by this code?

eglitise:
Closing due to no response

PR #1479, opened 2024-05-29
chore: update python template code

KazuCocoa:
https://github.com/appium/appium-inspector/issues/1478

Updating based on the latest python client


jlipps (approved):

eglitise:
I'm not sure if there are any client changes that would require adjusting the Inspector's recorder. For instance, `load_capabilities` should still be working with 2.3.0.

KazuCocoa:
I was wondering the main reason was "undefined" in the capabilities set

PR #1483, opened 2024-06-01
fix: Add validation for duplicate capability name

### LINKED ISSUE #1196: feature request: Duplicate Saved capability sets name
### Current Behavior

Currently when saving caps set with same name no message is shown
image

### Suggested Solution

I believe we should avoid having the same capability set names 

### Additional Information

_No response_

sudharsan-selvaraj:
Fix #1196

Added validations for duplicate capability set names upon saving.

1. In Save As modal

Screenshot 2024-06-02 at 3 25 54 AM

2. In Inline editor:

Screenshot 2024-06-02 at 3 26 30 AM

### DIFF
--- a/app/renderer/src/actions/Session.js
+++ b/app/renderer/src/actions/Session.js
@@ -76,6 +76,7 @@ export const SET_PROVIDERS = 'SET_PROVIDERS';
 
 export const SET_ADD_VENDOR_PREFIXES = 'SET_ADD_VENDOR_PREFIXES';
 
+export const SET_CAPABILITY_NAME_ERROR = 'SET_CAPABILITY_NAME_ERROR';
 export const SET_STATE_FROM_URL = 'SET_STATE_FROM_URL';
 export const SET_STATE_FROM_SAVED = 'SET_STATE_FROM_SAVED';
 
@@ -662,8 +663,15 @@ export function newSession(caps, attachSessId = null) {
 export function saveSession(server, serverType, caps, params) {
   return async (dispatch) => {
     let {name, uuid} = params;
-    dispatch({type: SAVE_SESSION_REQUESTED});
     let savedSessions = (await getSetting(SAVED_SESSIONS)) || [];
+    let isDuplicateName = savedSessions.find(
+      (session) => session.name?.toLowerCase() === name?.toLowerCase(),
+    );
+    if (isDuplicateName) {
+      return dispatch({type: SET_CAPABILITY_NAME_ERROR});
+    }
+    dispatch({type: SAVE_SESSION_REQUESTED});
+
     if (!uuid) {
       // If it's a new session, add it to the list
       uuid = UUID();
--- a/app/renderer/src/actions/Session.js
+++ b/app/renderer/src/actions/Session.js
@@ -665,7 +665,7 @@ export function saveSession(server, serverType, caps, params) {
     let {name, uuid} = params;
     let savedSessions = (await getSetting(SAVED_SESSIONS)) || [];
     let duplicateSessions = savedSessions.filter((session) => session.name === name);
-    // We need to save the details if user is updating an already saved capability set
+    // Ignore the check if the user is updating an existing capability set with duplicate names
     let isEditingExistingCapability = duplicateSessions.find((session) => session.uuid === uuid);
     if (duplicateSessions.length > 0 && !isEditingExistingCapability) {
       return dispatch({type: SET_CAPABILITY_NAME_ERROR});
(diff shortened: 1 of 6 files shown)


app/renderer/src/assets/locales/en/translation.json
```diff
@@ -281,5 +281,6 @@
   "noResultsFound": "No results found",
   "invalidCapType": "Invalid capability type: {{type}}",
   "whitespaceDetected": "Text Starts and/or Ends With Whitespace",
-  "pcloudyCredentialsRequired": "pCloudy username and api key are required!"
+  "pcloudyCredentialsRequired": "pCloudy username and api key are required!",
+  "duplicateCapabilityNameError": "Capability with the same name already present"
```
Dor-bl, line 285: The error text seems a bit confusing, one might think he used the same capability name such as platformName. Rather than the capability set settings

sudharsan-selvaraj: @Dor-bl Got it. How about rephrasing is as `Capability Set with the same name already present`? Open for suggestions.

app/renderer/src/actions/Session.js
```diff
@@ -662,8 +663,15 @@ export function newSession(caps, attachSessId = null) {
 export function saveSession(server, serverType, caps, params) {
   return async (dispatch) => {
     let {name, uuid} = params;
-    dispatch({type: SAVE_SESSION_REQUESTED});
     let savedSessions = (await getSetting(SAVED_SESSIONS)) || [];
+    let duplicateSession = savedSessions.find(
+      (session) => session.name?.toLowerCase() === name?.toLowerCase(),
```
eglitise, line 668: We cannot use optional chaining (Parcel 1 does not support it). Also I think it's fine to only do case-sensitive matching, meaning the `toLowerCase` can be removed, and the optional chaining would not be necessary anyway.

app/renderer/src/assets/locales/en/translation.json
```diff
@@ -281,5 +281,6 @@
   "noResultsFound": "No results found",
   "invalidCapType": "Invalid capability type: {{type}}",
   "whitespaceDetected": "Text Starts and/or Ends With Whitespace",
-  "pcloudyCredentialsRequired": "pCloudy username and api key are required!"
+  "pcloudyCredentialsRequired": "pCloudy username and api key are required!",
+  "duplicateCapabilityNameError": "Capability Set with the same name already present"
```
eglitise, line 285: Maybe something like 'A capability set with the same name already exists'

eglitise (changes requested):
Overall I think the error message is fine, but the implementation breaks functionality for people who already have sets with duplicate names. Here's a simple way to reproduce:
* Have 2 sets with the same name
* Edit the raw JSON of one set -> save -> save
* Dev tools shows the `SET_CAPABILITY_NAME_ERROR` event being fired, but nothing happens in the UI
* Click on a different set and then back to the first one -> previously saved JSON changes are reverted

app/renderer/src/actions/Session.js
sudharsan-selvaraj: @eglitise Makes sense. I have updated the snipped to remove the optional and case sensitive checks.

app/renderer/src/assets/locales/en/translation.json
sudharsan-selvaraj: @eglitise Sounds good.

sudharsan-selvaraj:
> Overall I think the error message is fine, but the implementation breaks functionality for people who already have sets with duplicate names. Here's a simple way to reproduce:
> 
> * Have 2 sets with the same name
> * Edit the raw JSON of one set -> save -> save
> * Dev tools shows the `SET_CAPABILITY_NAME_ERROR` event being fired, but nothing happens in the UI
> * Click on a different set and then back to the first one -> previously saved JSON changes are reverted

@eglitise Nice catch. I have now enabled a check to make sure updating an existing capability set should be saved with any errors until user update the name.

eglitise (approved):

app/renderer/src/actions/Session.js
```diff
@@ -662,8 +663,15 @@ export function newSession(caps, attachSessId = null) {
 export function saveSession(server, serverType, caps, params) {
   return async (dispatch) => {
     let {name, uuid} = params;
-    dispatch({type: SAVE_SESSION_REQUESTED});
     let savedSessions = (await getSetting(SAVED_SESSIONS)) || [];
+    let duplicateSessions = savedSessions.filter((session) => session.name === name);
```
mykola-mokhnach, line 667: consider using const instead of let of a variable is not mutable

mykola-mokhnach (approved):

KazuCocoa (approved):

jlipps (approved):

jlipps:
thanks @sudharsan-selvaraj !

PR #1485, opened 2024-06-04
feat: add locator details to find element results

### LINKED ISSUE #1241: feature request: Add searched element to popup window
### Current Behavior

Currently when searching for an element, in case the element is not found the user see a general message saying: "Could not find any elements"

### Suggested Solution

the popup should include the element we searched for as well.
i.e. Could not find any elements matching the below criteria: {Element selector value}"
image

### Additional Information

This will help users when they are working closely with dev teams to add unique identifiers to their app, so they can show them what is missing for what value of selector used.

sudharsan-selvaraj:
Fix #1241 

Screenshot 2024-06-04 at 7 00 51 AM

### DIFF
--- a/app/renderer/src/components/Inspector/LocatedElements.jsx
+++ b/app/renderer/src/components/Inspector/LocatedElements.jsx
@@ -48,11 +48,11 @@ const LocatedElements = (props) => {
       {locatedElements.length === 0 && (
         <Space className={InspectorStyles.spaceContainer} direction="vertical" size="small">
           <Row>
-          <i>{t('couldNotFindAnyElements')}</i>
+            <i>{t('couldNotFindAnyElements')}</i>
           </Row>
           <Row>
             {t('locatorStrategy')} {locatorTestStrategy}
-            <br/>
+            <br />
             {t('selector')} {locatorTestValue}
           </Row>
           {showIdAutocompleteInfo()}
--- a/app/renderer/src/components/Inspector/LocatedElements.jsx
+++ b/app/renderer/src/components/Inspector/LocatedElements.jsx
@@ -48,8 +48,7 @@ const LocatedElements = (props) => {
       {locatedElements.length === 0 && (
         <Space className={InspectorStyles.spaceContainer} direction="vertical" size="small">
           <Row>
-            <i>{`${t('couldNotFindAnyElements')} with locator `}</i>
-            <b>{`[${locatorTestStrategy} = ${locatorTestValue}]`}</b>
+            <i>{t('couldNotFindAnyElements', {locatorTestStrategy, locatorTestValue})}</i>
           </Row>
           {showIdAutocompleteInfo()}
         </Space>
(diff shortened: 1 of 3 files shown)


app/renderer/src/components/Inspector/LocatedElements.jsx
```diff
@@ -46,7 +48,8 @@ const LocatedElements = (props) => {
       {locatedElements.length === 0 && (
         <Space className={InspectorStyles.spaceContainer} direction="vertical" size="small">
           <Row>
-            <i>{t('couldNotFindAnyElements')}</i>
+            <i>{`${t('couldNotFindAnyElements')} with locator `}</i>
```
mykola-mokhnach, line 51: Please update the original message in the resources instead. Changing its payload here would lead to broken localizations. The message itself should be changed to (pay attention to the format specifiers presence)

`The locator [{{locatorTestStrategy}} = {{locatorTestValue}}] did not match any elements`

sudharsan-selvaraj: @mykola-mokhnach That makes sense. I have updated the message in the translation now.

app/renderer/src/components/Inspector/LocatedElements.jsx
```diff
@@ -46,7 +48,7 @@ const LocatedElements = (props) => {
       {locatedElements.length === 0 && (
         <Space className={InspectorStyles.spaceContainer} direction="vertical" size="small">
           <Row>
-            <i>{t('couldNotFindAnyElements')}</i>
+            <i>{t('couldNotFindAnyElements', {locatorTestStrategy, locatorTestValue})}</i>
           </Row>
```
eglitise, line 52: I think it's cleaner if the text string is not modified, and instead, additional lines are added with the locator details:
```suggestion
            <i>{t('couldNotFindAnyElements')}</i>
          </Row>
          <Row>
            {t('locatorStrategy')} {locatorTestStrategy}
            <br />
            {t('selector')} {locatorTestValue}
          </Row>
```
It would look like this:
image

sudharsan-selvaraj: @eglitise This look good to me as well. Let's also see what @mykola-mokhnach thinks?

mykola-mokhnach: this might bite us in response for some locales where the order of words in acetone is different from English. Always prefer to use format specifier instead of hardcoded string catenation logic.

sudharsan-selvaraj: @mykola-mokhnach I like the idea pointed out by @eglitise to show the locators details in new line to give better readability. Shall we proceed with the suggested design?

eglitise: I would still propose the above approach, because the order of words is the same one that is already used in the search panel modal landing page. Combining the text + strategy + selector in the same sentence makes it a lot more difficult to read, especially when italicized.

sudharsan-selvaraj: @eglitise @mykola-mokhnach Initially i was not sure how the localisation has been implemented. Since we are using the translations as Label, it should work pretty much the same way as it is in the Search panel model.

eglitise (approved):

KazuCocoa (approved):

Issue #1004, opened 2023-07-27
feature request: Search option at the App Source section


maioni:
### Current Behavior

Why there's no SEARCH option at the App Source section (at Appium Inspector)?

It could help a lot, sometimes I can't find elements by clicking on it, so I have to copy all the source to a note text application, find something I need, and after it, goes back to inspector to make sure I can find it with a selector I created. 

An option to search here could help.

Thank you! 

### Suggested Solution

Add a Search text option to the App Source section.

After type the input, it would highlight the element or elements found. (With a find next or find previous option)

### Additional Information

I got this idea when automating a sample application SWAG LABS, after opened the menu and try to click on options (the logout option), it was selecting the elements under the menu, and not from menu. 

So I had to copy all the SOURCE, and try to find the menu elements by text, and goes back to Inspector to validate my current selector.

https://github.com/saucelabs/sample-app-mobile/releases

jlipps:
Yeah I suppose a simple CMD+F doesn't work in the source box. Seems like a reasonable request to be able to search inside the source tree. Not sure how easy it would be to implement though! Might not be worth it if it's high effort

sudharsan-selvaraj:
@jlipps @maioni This seems like a reasonable request and a valuable feature to have. I've managed to develop it and would love to hear your feedback. Here's the link to [PR](https://github.com/appium/appium-inspector/pull/1494) and i have added a quick demo as well.

danieltlf:
@sudharsan-selvaraj Thank you for your PR! I really appreciate it! Good work!

maioni:
@sudharsan-selvaraj thank you! it is really good!

PR #1494, opened 2024-06-07
feat: add functionality to search through app source tree

sudharsan-selvaraj:
Fixes https://github.com/appium/appium-inspector/issues/1004

**Demo**

https://github.com/appium/appium-inspector/assets/20136913/70ee6680-6504-47ac-83e3-bb1b32c709b1

### DIFF
--- a/app/renderer/src/components/Inspector/Inspector.jsx
+++ b/app/renderer/src/components/Inspector/Inspector.jsx
@@ -11,8 +11,9 @@ import {
   SelectOutlined,
   TagOutlined,
   ThunderboltOutlined,
+  SearchOutlined,
 } from '@ant-design/icons';
-import {Button, Card, Modal, Space, Spin, Switch, Tabs, Tooltip} from 'antd';
+import {Button, Card, Modal, Row, Space, Spin, Switch, Tabs, Tooltip, Input} from 'antd';
 import {debounce} from 'lodash';
 import React, {useEffect, useRef, useState} from 'react';
 import {useNavigate} from 'react-router-dom';
@@ -87,6 +88,7 @@ const Inspector = (props) => {
     toggleShowAttributes,
     isSourceRefreshOn,
     windowSize,
+    setPageSourceSearchText,
     t,
   } = props;
 
@@ -314,7 +316,14 @@ const Inspector = (props) => {
                         </span>
                       }
                       extra={
-                        <span>
+                        <Row>
+                          <Input
+                            size="middle"
+                            allowClear
+                            placeholder="Search in page source"
+                            onChange={(e) => setPageSourceSearchText(e.target.value)}
+                            prefix={<SearchOutlined />}
+                          />
                           <Tooltip title={t('Toggle Attributes')}>
                             <Button
                               type="text"
@@ -339,7 +348,7 @@ const Inspector = (props) => {
                               onClick={() => downloadXML(sourceXML)}
                             />
                           </Tooltip>
-                        </span>
+                        </Row>
                       }
                     >
                       <Source {...props} />
--- a/app/renderer/src/components/Inspector/Inspector.module.css
+++ b/app/renderer/src/components/Inspector/Inspector.module.css
@@ -73,6 +73,10 @@
   height: 100%;
 }
 
+.inspector-main .interaction-tab-container :global(.ant-input-affix-wrapper) {
+  width: 200px;
+}
+
 .inspector-main .interaction-tab-container {
   position: relative;
   flex-grow: 1;
@@ -124,6 +128,11 @@
 .inspector-main .tree-container {
   height: 100%;
   overflow: auto;
+}
+
+.inspector-main .tree-search-value {
+  color: #272625;
+  background: #b6d932;
 }
 
 .session-info-table {
--- a/app/renderer/src/components/Inspector/Source.jsx
+++ b/app/renderer/src/components/Inspector/Source.jsx
@@ -1,10 +1,11 @@
 import {Spin, Tree} from 'antd';
-import React from 'react';
-
+import React, {useEffect} from 'react';
+import {renderToString} from 'react-dom/server';
 import {IMPORTANT_SOURCE_ATTRS} from '../../constants/source';
 import InspectorStyles from './Inspector.module.css';
 import LocatorTestModal from './LocatorTestModal.jsx';
 import SiriCommandModal from './SiriCommandModal.jsx';
+import {uniq} from 'lodash';
 
 /**
  * Shows the 'source' of the app as a Tree
@@ -20,8 +21,72 @@ const Source = (props) => {
     methodCallInProgress,
     mjpegScreenshotUrl,
     isSourceRefreshOn,
+    pageSourceSearchText,
     t,
   } = props;
+
+  /**
+   * Highlights the part of the node text in source tree that matches the search term.
+   */
+  const higlightNodeMatchingSearchTerm = (nodeText, searchText) => {
+    if (!searchText) {
+      return nodeText;
+    }
+
+    const index = nodeText.indexOf(searchText);
+    if (index < 0) {
+      return nodeText;
+    }
+    const prefix = nodeText.substring(0, index);
+    const suffix = nodeText.slice(index + searchText.length);
+    //Matched word will be wrapped in a separate span for custom highlighting
+    const matchedWord = nodeText.slice(index, index + searchText.length);
+
+    return (
+      <>
+        {prefix}
+        <span className={InspectorStyles['tree-search-value']} data-match={searchText}>
+          {matchedWord}
+        </span>
+        {suffix}
+      </>
+    );
+  };
+
+  useEffect(() => {
+    if (!treeData || !pageSourceSearchText) {
+      return;
+    }
+
+    const nodesMatchingSearchTerm = [];
+
+    /**
+     * If any search text is entered, we will try to find matching nodes in the tree.
+     * and expand their parents to make the nodes visible that matches the
+     * search text.
+     *
+     * hierarchy is an array of node keys representing the path from the root to the
+     * current node.
+     */
+    const findNodesToExpand = (node, hierarchy) => {
+      /* Node title will an object representing a react element.
+       * renderToString method will construct a HTML DOM string
+       * which can be used to match against the search text.
+       *
+       * If any node that matches the search text is found, we will add all its
+       * parents to the 'nodesMatchingSearchTerm' array to make them automatically expand.
+       */
+      const nodeText = renderToString(node.title);
+      if (nodeText.indexOf(pageSourceSearchText) > -1) {
+        nodesMatchingSearchTerm.push(...hierarchy);
+      }
+      if (node.children) {
+        node.children.forEach((c) => findNodesToExpand(c, [...hierarchy, node.key]));
+      }
+    };
+    treeData.forEach((node) => findNodesToExpand(node, [node.key]));
+    setExpandedPaths(uniq(nodesMatchingSearchTerm));
+  }, [treeData, pageSourceSearchText]);
 
   const getFormattedTag = (el, showAllAttrs) => {
     const {tagName, attributes} = el;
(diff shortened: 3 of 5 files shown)


app/renderer/src/components/Inspector/Inspector.jsx
```diff
@@ -314,7 +316,14 @@ const Inspector = (props) => {
                         </span>
                       }
                       extra={
-                        <span>
+                        <Row>
+                          <Input
+                            size="middle"
+                            allowClear
+                            placeholder="Search in page source"
```
eglitise, line 323: Please add this to the translations file

app/renderer/src/components/Inspector/Source.jsx
```diff
@@ -20,27 +21,98 @@ const Source = (props) => {
...
     mjpegScreenshotUrl,
     isSourceRefreshOn,
+    pageSourceSearchText,
     t,
   } = props;
 
+  /**
+   * Highlights the part of the node text in source tree that matches the search term.
+   */
+  const higlightNodeMatchingSearchTerm = (nodeText, searchText) => {
```
eglitise, line 31: Typo in variable name

app/renderer/src/components/Inspector/Source.jsx
```diff
@@ -20,27 +21,98 @@ const Source = (props) => {
...
+    }
+
+    const index = nodeText.indexOf(searchText);
+    if (index < 0) {
+      return nodeText;
+    }
+    const prefix = nodeText.substring(0, index);
+    const suffix = nodeText.slice(index + searchText.length);
+    //Matched word will be wrapped in a separate span for custom highlighting
+    const matchedWord = nodeText.slice(index, index + searchText.length);
```
eglitise, line 43: I assume `matchedWord` is used due to potentially different casing from `searchText`?

eglitise (changes requested):
Thank you! Will test this a bit later.
Similarly to your other PR, would it be possible to also update the documentation? I am referring to this page: https://appium.github.io/appium-inspector/latest/session-inspector/source/
* Update the top two images
* Add a new section in Application Source (between Refreshing the Source and Toggle Attributes Button)

app/renderer/src/components/Inspector/Inspector.jsx
sudharsan-selvaraj: Moved the string to translations file.

app/renderer/src/components/Inspector/Source.jsx
sudharsan-selvaraj: Yup. Searching is current not case sensitive. So to preserve the highlighted text casing we are using the `matchedWord` to extract the exact word from the source.

sudharsan-selvaraj:
> Thank you! Will test this a bit later. Similarly to your other PR, would it be possible to also update the documentation? I am referring to this page: https://appium.github.io/appium-inspector/latest/session-inspector/source/
> 
> * Update the top two images
> * Add a new section in Application Source (between Refreshing the Source and Toggle Attributes Button)


Thanks for pointing out to the docs. I have updated all relevant images and added a new section with the gist of search functionality. Do let me know for any improvements.

app/renderer/src/components/Inspector/Source.jsx
```diff
@@ -20,27 +21,98 @@ const Source = (props) => {
...
+    pageSourceSearchText,
     t,
   } = props;
 
+  /**
+   * Highlights the part of the node text in source tree that matches the search term.
+   */
+  const highlightNodeMatchingSearchTerm = (nodeText, searchText) => {
+    if (!searchText) {
+      return nodeText;
```
mykola-mokhnach, line 33: It is not very clear to me why we return the nodeText if no matches have been found. I would expect nothing to be returned/highlighted in such case

app/renderer/src/components/Inspector/Source.jsx
```diff
@@ -20,27 +21,98 @@ const Source = (props) => {
...
+    }
+
+    const index = nodeText.toLowerCase().indexOf(searchText.toLowerCase());
```
mykola-mokhnach, line 36: is nodeText always defined?

app/renderer/src/components/Inspector/Source.jsx
mykola-mokhnach: can we just do `matchedWord = suffix`?

app/renderer/src/components/Inspector/Source.jsx
```diff
@@ -20,27 +21,98 @@ const Source = (props) => {
...
+    const findNodesToExpand = (node, hierarchy) => {
+      /* Node title will an object representing a react element.
+       * renderToString method will construct a HTML DOM string
+       * which can be used to match against the search text.
+       *
+       * If any node that matches the search text is found, we will add all its
+       * parents to the 'nodesMatchingSearchTerm' array to make them automatically expand.
+       */
+      const nodeText = renderToString(node.title).toLowerCase();
+      if (nodeText.indexOf(pageSourceSearchText.toLowerCase()) > -1) {
```
mykola-mokhnach, line 80: consider using `.includes`

app/renderer/src/components/Inspector/Source.jsx
```diff
@@ -20,27 +21,98 @@ const Source = (props) => {
     methodCallInProgress,
     mjpegScreenshotUrl,
     isSourceRefreshOn,
+    pageSourceSearchText,
     t,
   } = props;
 
+  /**
+   * Highlights the part of the node text in source tree that matches the search term.
```
mykola-mokhnach, line 29: Please add a description for the returned value

app/renderer/src/components/Inspector/Source.jsx
```diff
@@ -20,27 +21,98 @@ const Source = (props) => {
...
+   */
+  const highlightNodeMatchingSearchTerm = (nodeText, searchText) => {
```
mykola-mokhnach, line 31: it would be nice to add a unit test for this method

app/renderer/src/components/Inspector/Source.jsx
sudharsan-selvaraj: Node Text represents each XML node in the app source. So it's unlikely to be undefined. But i have added a safety check to make sure it is defined before processing.

app/renderer/src/components/Inspector/Source.jsx
sudharsan-selvaraj: Updated the logic.

app/renderer/src/components/Inspector/Source.jsx
sudharsan-selvaraj: Added necessaty unit tests.

package.json
```diff
@@ -150,6 +150,7 @@
     "eslint-plugin-promise": "6.2.0",
     "eslint-plugin-react": "7.34.2",
     "eslint-plugin-react-native": "4.1.0",
+    "ignore-styles": "^5.0.1",
```
eglitise, line 153: Is this import necessary? The repo has been archived as of 2 years ago.

app/renderer/src/assets/stylesheets/common.global.less
```diff
@@ -99,3 +99,8 @@ body::-webkit-scrollbar-corner {
 .ant-tabs-tab {
   font-size: 14px;
 }
+
+.search-word-highlighted {
```
eglitise, line 103: I would suggest moving this back to the Inspector-specific stylesheet, since this will not be used for the Session page

sudharsan-selvaraj: I moved it to common module to improve the testability and since it is a common class which can be used across any module in future it makes sense to keep it in common css.

package.json
eglitise: Hmm, I can suggest a slightly different approach: instead of exporting the whole `highlightNodeMatchingSearchTerm` function, separate its text search functionality into its own method and place it in `utils/source-parsing.js`. Something like this:
```js
export function findNodeMatchingSearchTerm(nodeText, searchText) {
  if (!searchText || !nodeText) {
    return null;
  }

  const index = nodeText.toLowerCase().indexOf(searchText.toLowerCase());
  if (index < 0) {
    return null;
  }
  const prefix = nodeText.substring(0, index);
  const suffix = nodeText.slice(index + searchText.length);
  // Matched word will be wrapped in a separate span for custom highlighting
  const matchedWord = nodeText.slice(index, index + searchText.length);
  return {prefix, matchedWord, suffix};
}
```

Then `highlightNodeMatchingSearchTerm` can be as follows:
```js
const highlightNodeMatchingSearchTerm = (nodeText, searchText) => {
  const searchResults = findNodeMatchingSearchTerm(nodeText, searchText);
  if (!searchResults) {
    return nodeText;
  }
  const {prefix, matchedWord, suffix} = searchResults;
  return (
    <>
      {prefix}
      <span className="search-word-highlighted" data-match={searchText}>
        {matchedWord}
      </span>
      {suffix}
    </>
  );
};
```

Then you can change the unit tests to validate `findNodeMatchingSearchTerm` (in `utils-source-parsing-specs.js`), and this extra import will not be necessary.

app/renderer/src/assets/stylesheets/common.global.less
eglitise: See my other comment - if nothing is exported from the JSX, this is not needed for test purposes. Also, this file does not define any custom styles, but only overrides default ones. All custom styles are contained in the page-specific stylesheets.

jlipps (commented):
cool feature!

app/renderer/src/assets/stylesheets/common.global.less
sudharsan-selvaraj: Got it. i was banging my head around looking for ways to enable the testability. This make more sense and will make the necessary changes.

package.json
sudharsan-selvaraj: Got it. i was banging my head around looking for ways to enable the testability. This make more sense and will make the necessary changes.

app/renderer/src/components/Inspector/Source.jsx
```diff
@@ -20,27 +52,70 @@ const Source = (props) => {
...
+      if (nodeText.includes(pageSourceSearchText.toLowerCase())) {
+        nodesMatchingSearchTerm.push(...hierarchy);
+      }
+      if (node.children) {
+        node.children.forEach((c) => findNodesToExpand(c, [...hierarchy, node.key]));
+      }
+    };
+    treeData.forEach((node) => findNodesToExpand(node, [node.key]));
+    setExpandedPaths(uniq(nodesMatchingSearchTerm));
+  }, [treeData, pageSourceSearchText]);
```
eglitise, line 92: I wanted to test this but unfortunately I got an error: `Cannot access 'treeData' before initialization`.
If I moved the `useEffect` hook below the `treeData` definition, then I got another error: `Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.`
I checked and this happens due to too many updates in `treeData`.

app/renderer/src/components/Inspector/Inspector.module.css
```diff
@@ -73,6 +73,10 @@
   height: 100%;
 }
 
+.inspector-main .interaction-tab-container :global(.ant-input-affix-wrapper) {
```
eglitise, line 76: This does not look good at the minimum window width:
image
Although I'm not sure how to best scale this. Ideally it should shrink without affecting the card title, but the actual behavior is the opposite.

eglitise: Also I would suggest defining a class for the searchbar if you intend to change its CSS. This particular line is rather long and it's not obvious what it's meant to affect.

eglitise (changes requested):
It seems to work well in the browser, but not so much in the Electron version :(

mykola-mokhnach:
@sudharsan-selvaraj Would you have some time to finish this PR?

sudharsan-selvaraj:
> @sudharsan-selvaraj Would you have some time to finish this PR?

@mykola-mokhnach Yup, I will be able to close the pending items.

app/renderer/src/components/Inspector/Inspector.module.css
sudharsan-selvaraj: I just tweaked a bit to make it better for smaller screens. Do let me know if it looks good.

Screenshot 2024-08-19 at 5 38 49 PM

app/renderer/src/components/Inspector/Source.jsx
sudharsan-selvaraj: Got it. It appears that `treeData` isn't necessary as a dependency for the hook. Instead, we just need to listen for changes in the search input and re-render the tree as needed.

Interestingly, this issue didn't come up when the PR was initially created

app/renderer/src/components/Inspector/Inspector.module.css
eglitise: It is better in that there's no linebreak, but ideally the search bar and icons should not hide the 'App Source' title - instead the title should hide the search bar/buttons. Not sure how possible this is.

eglitise:
The good thing is that it does work in the Electron version now :) But I have some issues with the tree node text extraction:
* It uses `renderToString` which is imported from `react-dom/server`, and [as per React docs](https://react.dev/reference/react-dom/server/renderToString#removing-rendertostring-from-the-client-code), it 'unnecessarily increases your bundle size and should be avoided'. There do seem to be alternatives but I haven't checked them in detail.
* The converted node titles have additional wrapper text, which is unfortunately also taken into account when searching. For example, a node `<android.widget.LinearLayout>` would be converted to `<span>&lt;<b class="_sourcetag_zwtnd_433">android.widget.linearlayout</b>&gt;</span>`. Meaning, if I search for `span`, the entire DOM tree will be unwrapped, and potentially not a single node will be highlighted.

eglitise:
I would also suggest exploring the idea of not converting the entire DOM upon every change in the search string. Maybe the raw title of each node can be calculated as part of `recursive`, or at least the DOM should be converted only once upon starting search.

saikrishna321:
@sudharsan-selvaraj any updates on this PR ?

eglitise:
@sudharsan-selvaraj I will close this PR as it has been more than a year since its opening, but you are welcome to reopen it after resolving the merge conflicts and addressing the comments.

PR #1514, opened 2024-06-28
chore: separate web and Electron-specific code

eglitise:
This is another refactor aimed at simplifying the eventual migration to Vite:
* Introduce a 3-folder structure in `/app`: `common`, `electron`, and `web`
* The `common` folder retains all code shared between both app versions
* The `electron` folder contains:
  * Electron-specific stuff (`main` process code, `splash.html`, unique imports for polyfills)
  * A copy of `index.html` to ensure unique entrypoints
* The `web` folder contains:
  * Web-specific stuff (unique imports for polyfills)
  * A copy of `index.html` to ensure unique entrypoints
* Moved `polyfills/index.js` to `shared/polyfills.js` since the other polyfill files are now version-specific
  * For now the file still retains the `require`s from individual platforms. I could not yet find a way to avoid this using dependency injections.
* Moved `getSaveableState` from `main` to `renderer` since it was not used anywhere in `main` 
* Moved `i18next.config.js` to `electron/main`, this avoids unneeded imports for web

The main benefit of this refactor is that both the web and Electron versions no longer import dependencies specific to the other version, which reduces warnings/errors and likely also the app bundle size.

Web and Electron versions were both tested to still work fine, including the Electron version's i18n functionality.


KazuCocoa (approved):
lgtm, I like the idea to clarify web/electron.
(I have also tested briefly)

mykola-mokhnach (approved):

PR #1529, opened 2024-07-10
chore: use common logger class

eglitise:
This PR adds a basic `Logger` class for abstracting away some logging implementations for better maintenance. The class itself still executes basic `console` calls under the hood, but this can be improved in the future.
The following logging implementations have been replaced with `Logger`:
* All uses of `console` in `common` files
* All uses of the `log` method imported from the shared `polyfills` file (there were only 2 such instances)
  * The browser polyfill still used `console`, so there is no difference
  * The Electron polyfill used `electron-log`, so this dependency has been removed

The actual main goal of this PR is to remove `electron-log`, as it is only being called in the renderer process anyway, and also it was causing issues for the Vite migration.

### DIFF
--- a/app/common/utils/other.js
+++ b/app/common/utils/other.js
@@ -15,6 +15,22 @@ const VALID_W3C_CAPS = [
   'unhandledPromptBehavior',
 ];
 
+class Logger {
+  info(...args) {
+    console.info(...args); // eslint-disable-line no-console
+  }
+
+  warn(...args) {
+    console.warn(...args); // eslint-disable-line no-console
+  }
+
+  error(...args) {
+    console.error(...args); // eslint-disable-line no-console
+  }
+}
+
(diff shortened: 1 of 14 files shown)


app/common/utils/other.js
```diff
@@ -15,6 +15,22 @@ const VALID_W3C_CAPS = [
   'unhandledPromptBehavior',
 ];
 
+class Logger {
```
mykola-mokhnach, line 18: I would rather keep the logger in a separate module

mykola-mokhnach (approved):

eglitise:
Indeed, using `@appium/logger` was actually my first choice, but since it uses `node:` imports which are not supported by the current Electron version, it was not an option.

mykola-mokhnach (approved):

PR #1531, opened 2024-07-10
chore: remove filepath input type in capability builder

eglitise:
This PR removes the filepath input type in the capability builder. There are 3 main reasons for this:
* The file picker button has never worked in the browser version
* The capability value is formatted as a standard string, same as the text and object input types
* It is the only instance of using Electron's remote module, which is causing issues with the Vite migration

For users who are currently using this capability type, the capability should still work the same as before, but the value input field will no longer have the file picker button. I tested this in both web and Electron versions.

### DIFF
--- a/app/web/polyfills.js
+++ b/app/web/polyfills.js
@@ -36,7 +36,6 @@ class BrowserSettings {
   }
 }
 
-const log = console;
 const settings = new BrowserSettings();
 const {clipboard, shell, ipcRenderer} = browser;
 const i18NextBackendOptions = {
@@ -49,4 +48,4 @@ const i18NextBackendOptions = {
   ],
 };
 
-export {log, settings, clipboard, shell, ipcRenderer, i18NextBackend, i18NextBackendOptions};
+export {settings, clipboard, shell, ipcRenderer, i18NextBackend, i18NextBackendOptions};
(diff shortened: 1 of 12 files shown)


app/web/polyfills.js
```diff
@@ -9,11 +9,6 @@ const browser = {
   shell: {
     openExternal: (url) => window.open(url, ''),
   },
-  remote: {
```
mykola-mokhnach, line 12: how this change is related to the original PR?

mykola-mokhnach:
We may still keep the `filepath` combo item, just by making it an alias to a "text", just without the button. We may also add an additional validator on this field for file path existence.

app/web/polyfills.js
eglitise: It is the browser version's polyfill equivalent of Electron's `remote` polyfill. Since the latter has been removed, the former can be removed, too.

eglitise:
This PR does still keep backwards compatibility with the `file` type, but I don't really see the point in keeping a dedicated type that's just an alias for an existing one. IMO, the button was the main benefit of the filepath type.
Checking the path existence is certainly useful, but the browser cannot access the filesystem, so this functionality will also not work there.

mykola-mokhnach (approved):

PR #1558, opened 2024-07-19
chore: migrate from Parcel v1 to Vite

### LINKED ISSUE #1453: chore: replace Parcel with Vite
Parcel v1 is greatly outdated and is holding back many important dependency upgrades, like Electron and antd.
Several possible replacements for it exist (Parcel v2, Webpack, Vite, etc.), which were examined and tested on a fresh Electron+React project. Vite seemed to be the most viable candidate, though, like the others, it also requires some refactoring beforehand. This issue will be used to track the migration progress.

Related tasks:
* #1444 
* #1447
* #1514
* #1529
* #1531
* #1542
* #1544
* #1546 
* #1550 
* #1551
* #1554 
* #1558 (the actual replacement PR)

eglitise:
The time has finally come. This PR ditches Parcel v1 and migrates to the very latest (!) version of Vite! 🎉 
Other changes include:
* Simplifying the `@import` paths for the `.less` styling file (this didn't work in Parcel)
* Removing the additional HMR code in `index.jsx` (Vite supports this by default)
* Using an alias to solve the problem with handling web/Electron polyfill imports (probably also possible in Parcel, but likely not as straightforward, since there was no dedicated Parcel config file)
* Changing i18n locale paths depending on whether dev or production mode is used
* Skipping the browser build in Azure pipelines during publishing (this is not directly related to Vite)
* Updating the node scripts for consistency and ease of use
* Moving to `vitest` for unit and integration tests. E2E tests were left as-is, but I don't think they have ever worked anyway.

Some explicitly imported dependencies used with Parcel can now also be removed (e.g. `postcss`), but I am trying to keep this PR as small as possible, so that will come in a separate PR. It will also include the removal of `mocha`/`chai` and E2E tests being commented out, to be rebuilt in the future.

Basic functionality was tested on both macOS and Windows for:
* Browser dev build
* Browser production build
* Electron dev build
* Electron production build
* Electron packaged app

One unexpected benefit from this migration is that the app seems to open notably faster as well!

Closes #1453.


package.json
```diff
@@ -17,37 +17,28 @@
   "keywords": [
     "appium"
   ],
-  "assetsPath": "app/common/public/locales",
```
eglitise, line 20: This line was used by `parcel-plugin-asset-copier`, which is being removed now

eglitise:
Hmm that's odd, I had the tests run just fine previously...

app/electron/main/i18next.js
```diff
@@ -1,12 +1,17 @@
 import i18n from 'i18next';
 import i18NextBackend from 'i18next-fs-backend';
-import path from 'path';
+import {join} from 'path';
```
mykola-mokhnach, line 3: should we use `node:...` format for system packages?

app/electron/main/i18next.js
```diff
@@ -1,12 +1,17 @@
...
 
+const localesPath =
+  process.env.NODE_ENV === 'development'
+    ? 'app/common/public/locales' // from project root
+    : join(__dirname, '../renderer/locales'); // from 'main' in package.json
+
 const i18NextBackendOptions = {
-  loadPath: path.join(__dirname, '{{lng}}/{{ns}}.json'),
-  addPath: path.join(__dirname, '{{lng}}/{{ns}}.json'),
+  loadPath: `${localesPath}/{{lng}}/{{ns}}.json`,
```
mykola-mokhnach, line 13: should we also use path.join here to have valid path separators?

mykola-mokhnach: same below

app/electron/main/windows.js
```diff
@@ -1,13 +1,18 @@
...
 import i18n from './i18next';
 import {openFilePath} from './main';
 import {APPIUM_SESSION_EXTENSION, isDev} from './helpers';
 import {rebuildMenus} from './menus';
 
-const mainUrl = `file://${__dirname}/index.html`;
-const splashUrl = `file://${__dirname}/splash.html`;
+const mainPath = isDev
+  ? process.env.ELECTRON_RENDERER_URL
+  : join(__dirname, '../renderer/index.html'); // from 'main' in package.json
```
mykola-mokhnach, line 12: consider using the `join(__dirname, '..', 'renderer', 'index.html')` format as it should properly handle path reparators

app/electron/main/windows.js
```diff
@@ -52,12 +59,11 @@ function buildSessionWindow() {
...
   const splashWindow = buildSplashWindow();
-  mainWindow = buildSessionWindow();
-
-  splashWindow.loadURL(splashUrl);
+  isDev ? splashWindow.loadURL(splashPath) : splashWindow.loadFile(splashPath);
   splashWindow.show();
 
-  mainWindow.loadURL(mainUrl);
+  mainWindow = buildSessionWindow();
+  isDev ? mainWindow.loadURL(mainPath) : mainWindow.loadFile(mainPath);
```
mykola-mokhnach, line 66: could also be ```mainWindow[`load${isDev ? 'URL' : 'File'}`](mainPath)```

package.json
```diff
@@ -132,6 +122,7 @@
     "@types/react": "18.3.3",
     "@types/redux-logger": "3.0.13",
     "@types/sinon": "17.0.3",
+    "@vitejs/plugin-react": "4.3.1",
     "asyncbox": "3.0.0",
     "chai": "4.4.1",
```
mykola-mokhnach, line 127: do we still use chai...?

app/common/renderer/polyfills.js
```diff
@@ -1,39 +1,18 @@
...
-    clipboard,
-    shell,
-    ipcRenderer,
-    i18NextBackend,
-    i18NextBackendOptions,
-    fs,
-    util,
-  } = require('../../electron/renderer/polyfills'));
-}
+import {
```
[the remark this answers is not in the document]
eglitise: True, but I think it's cleaner to do this, rather than adding `eslint-disable-line` in many more files. Plus since neither VS Code nor ESLint are aware of this alias, the imports won't be auto-adjusted if file locations change, so having it all in one place would also reduce the manual effort.

app/electron/main/i18next.js
eglitise: I'm not sure if they will work here, since the currently used Electron version uses Node 14.16, whereas the `node:...` format was added in Node 14.18.

package.json
eglitise: No, but there are several other dependencies that can also be removed, so I'd rather keep this PR smaller and remove all other dependencies in a separate one

mykola-mokhnach: I can observe some test files already have `should` being replaced with `expect`.

app/electron/main/i18next.js
mykola-mokhnach: node14 is so ancient :(

app/common/renderer/polyfills.js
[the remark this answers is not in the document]
mykola-mokhnach: why `eslint-disable-line ` would be needed? Are we talking about the same thing: 

```js
export {
  settings,
  clipboard,
  shell,
  ipcRenderer,
  i18NextBackend,
  i18NextBackendOptions,
  fs,
  util,
} from '#local-polyfills'
```

app/electron/main/i18next.js
eglitise: That's why this PR is needed to unblock the Electron+Node upgrade :)

app/common/renderer/polyfills.js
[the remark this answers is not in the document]
eglitise: Yes - this last line currently has the `eslint-disable-line` comment, and if I remove the comment, I get `Unable to resolve path to module '#local-polyfills'`. Normally this would be resolved by adding the alias to the ESLint config file, but since the alias can point to two different locations, I thought it would better to just disable ESLint and leave an explanatory comment above it.

package.json
eglitise: Yes, unit and integration tests are now fully based on `vitest`, and only the E2E test files still use `chai`. But again, I don't recall them ever working in the first place, so I'm thinking of just commenting out their contents and removing `chai` completely. Not in this PR though.

app/common/renderer/polyfills.js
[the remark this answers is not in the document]
mykola-mokhnach: makes sense. Make leave a todo there for now

package.json
mykola-mokhnach: fine with me

app/common/renderer/polyfills.js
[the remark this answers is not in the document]
eglitise: Okay, I think I misunderstood, I didn't realise the two lines can just be combined. Will adjust this.

jlipps (approved):
amazing work!

KazuCocoa:
lg overall, I'll do some testing on my local as well tomorrow.

KazuCocoa (approved):
lg overall. Ran them on my local as well. Had the same comments for the path related (to use `join`) which is already addressed, although others look good for me.

electron.vite.config.mjs
```diff
@@ -0,0 +1,57 @@
+import react from '@vitejs/plugin-react';
+import {defineConfig, externalizeDepsPlugin} from 'electron-vite';
+import {join} from 'path';
+import renderer from 'vite-plugin-electron-renderer';
+
+export default defineConfig({
+  main: {
+    build: {
+      outDir: join(__dirname, 'dist/main'),
```
mykola-mokhnach, line 9: prefer to split path as separate arguments for all join calls

test/unit/webview-helpers.spec.js
```diff
@@ -0,0 +1,18 @@
...
+import {join} from 'path';
+import {describe, expect, it} from 'vitest';
+
+import {parseSource} from '../../app/common/renderer/lib/webview-helpers';
+
+describe('webview-helpers.js', function () {
+  describe('#parseSource', function () {
+    it('should parse html to proper xml', async function () {
+      const original = await fs.readFile(
+        join(__dirname, './mocks/appium.page.original.html'),
```
mykola-mokhnach, line 11: the `./` part is redundant

vite.config.mjs
```diff
@@ -0,0 +1,35 @@
...
+      less: {
+        javascriptEnabled: true,
+        additionalData: '@root-entry-name: default;',
+      },
+    },
+  },
+  plugins: [react()],
+  resolve: {
+    alias: {
+      '#local-polyfills': join(__dirname, 'app/web/polyfills'),
```
mykola-mokhnach, line 27: same as above

mykola-mokhnach (approved):

PR #1565, opened 2024-07-22
chore: clean up dependencies

eglitise:
This PR is a follow-up to #1558 and removes various dependencies that are no longer needed (as well as some that have not been needed for a while now):
* `chai`, `mocha`, `sinon`, `spectron`: these are no longer needed after removing their uses from the currently non-functional E2E tests
* `cross-env`: its last remaining use was in the npm script to run E2E tests
* `devtron`, `check-engines`: both required running an additional command to activate them, and I don't see much use in them anyway
* `postcss-modules`, `babel`: no longer (explicitly) needed after Vite
* `concurrently`,  `node-libs-browser`: not sure if these were ever used

For good measure, after removing the above dependencies, I also ran `npm run clean:npm`, which rebuilt the lockfile.
As a result of these changes, security is also improved - with #1558, the number of vulnerabilities in this project decreased from 80 to 27. With this PR, this number has gone down to 4.

Regarding changes in the E2E files - I don't feel like they should be removed entirely, since they can be used as a basis when fixing them in the future. I only made the minimum amount of changes to remove `chai` and `spectron`.


.eslintrc.json
```diff
@@ -1,10 +1,8 @@
 {
   "extends": ["plugin:react/recommended", "@appium/eslint-config-appium"],
```
mykola-mokhnach, line 2: would it be possible to use eslint-config-appium-ts instead of eslint-config-appium later and have .ts files supported?

package.json
```diff
@@ -27,7 +27,7 @@
     "test:format": "prettier . -c",
     "test:unit": "vitest run unit",
     "test:integration": "vitest run integration",
-    "test:e2e": "cross-env E2E_TIMEOUT=600000 NODE_ENV=test RUNNING_IN_SPECTRON=true mocha --reporter mocha-multi-reporters --reporter-options configFile=./test/mochareporters.json ./test/e2e",
+    "test:e2e": "echo 'Currently not functional'",
```
mykola-mokhnach, line 30: should we leave a TODO for these?

mykola-mokhnach: also what else need to be done to resurrect e2e tests?

.eslintrc.json
eglitise: Changed to the TS module. In order to pass linting now, I had to change `electron-debug` and `electron-extension-installer` `require()`s to top-level `import`s (and therefore also change them from dev deps to main deps), but this is also how their README suggests them to be used.

package.json
eglitise: Added a TODO in the script itself. As for getting them to work, well... I guess the first step is picking a framework, where my choices would be WDIO or Playwright. Then comes the part of figuring out how to integrate it, and I guess once that's done, the suite can be extended further.

mykola-mokhnach:
it looks like build fails now :(

eglitise:
I rebuilt the package-lock file without deleting node_modules and npm decided to delete optional dependencies 🙃
should be fine now

mykola-mokhnach (approved):

KazuCocoa (commented):
🚀

PR #1596, opened 2024-08-09
feat: add infobox for elements at snapshotMaxDepth

eglitise:
This PR adds a new infobox when selecting an element located at `snapshotMaxDepth`. It helps users understand that some elements may be hidden, explains the reason why they are hidden, and provides a way to expose them.
image
The hope is that this infobox will also provide additional help for users testing React Native apps on iOS, since such apps [are known to have issues with high element depth](https://appium.github.io/appium-inspector/latest/troubleshooting/#cannot-see-full-source-tree-with-xcuitest-driver-react-native-app).

### DIFF
--- a/app/common/public/locales/en/translation.json
+++ b/app/common/public/locales/en/translation.json
@@ -133,6 +133,7 @@
   "contextDropdownInfo": "Multiple contexts detected; certain elements might only be accessible after switching to a different context. Note that webview inspection in Appium Inspector is less accurate than the DevTools of Chrome or Safari. For more information, see:",
   "idAutocompletionCanBeDisabled": "The requested id selector does not have a package name prefix. This Appium session has package name autocompletion enabled, which may be the reason why no elements were found. To disable this behavior, relaunch this session with the capability 'appium:disableIdLocatorAutocompletion' set to 'true'.",
   "missingAutomationNameForStrategies": "Additional locator strategies may be available. To view them, add the capability 'appium:automationName' when launching the session. Note that this capability is mandatory in Appium 2.",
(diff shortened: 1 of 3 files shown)


KazuCocoa (approved):
👍

app/common/public/locales/en/translation.json
```diff
@@ -133,6 +133,7 @@
   "contextDropdownInfo": "Multiple contexts detected; certain elements might only be accessible after switching to a different context. Note that webview inspection in Appium Inspector is less accurate than the DevTools of Chrome or Safari. For more information, see:",
   "idAutocompletionCanBeDisabled": "The requested id selector does not have a package name prefix. This Appium session has package name autocompletion enabled, which may be the reason why no elements were found. To disable this behavior, relaunch this session with the capability 'appium:disableIdLocatorAutocompletion' set to 'true'.",
   "missingAutomationNameForStrategies": "Additional locator strategies may be available. To view them, add the capability 'appium:automationName' when launching the session. Note that this capability is mandatory in Appium 2.",
+  "snapshotMaxDepthReached": "This element is located at the maximum source depth {{selectedElementDepth}}, therefore its child elements (if any) are not shown. To reveal them, change the 'snapshotMaxDepth' setting to a value higher than {{selectedElementDepth}}. Check your driver documentation for the maximum supported value.",
```
mykola-mokhnach, line 136: should we also mention the maximum supported value is 62?

mykola-mokhnach: perhaps `greater than`?

eglitise: The maximum value of 62 refers only to the XCUITest driver, but for UiAutomator2 it seems to be 500.

mykola-mokhnach: I assume it would still be hard for users to find this value in docs themselves

mykola-mokhnach: for UIA2 the value does not really have an upper limit. I don't know exactly at which value the tree reader starts to freak out

mykola-mokhnach (approved):

PR #1607, opened 2024-08-15
feat: improve UX for attaching to existing session

### LINKED ISSUE #1585: feature request: Improved functionality on the "Attach to Session" tab
### Current Behavior

1. The latest session will be at the end of the session list.
2. It's hard to tell one session from another unless you look at the server logs to see the Session ID.
3. Sessions might not show up in the dropdown list unless you click the refresh icon.

image

### Suggested Solution

1. Show the last session at the top of the list and make it pre-selected
2. Add a timestamp to sessions
3. When you click on the list, automatically press refresh

### Additional Information

_No response_

eglitise:
This PR is a slight enhancement to the Attach to Session tab functionality:
* Do not retrieve sessions when opening app or changing server parameters
* Automatically retrieve sessions upon opening the Attach to Session tab
* Move `attachSessId` calculation from reducers file to actions file
* Do not auto-populate selected session if multiple sessions are found (unless a session was already selected)
* List discovered sessions in reverse order (newest to oldest)
* Add more details in session text (e.g. `bundleId`/`appPackage` if no `app` found; `udid` if no `deviceName` found)
  * I would also add timestamps, but the Appium server does not store this information
* Add tooltip for refresh button
* Update docs

Resolves #1585.


app/common/renderer/actions/Session.js
```diff
@@ -958,7 +949,19 @@ export function getRunningSessions() {
...
-      dispatch({type: GET_SESSIONS_DONE, sessions: res.data.value});
+      const sessions = res.data.value;
+      dispatch({type: GET_SESSIONS_DONE, sessions});
+
+      // set attachSessId if only one session found
+      if (sessions.length === 1) {
+        dispatch({type: SET_ATTACH_SESS_ID, attachSessId: sessions[0].id});
+      } else if (attachSessId) {
+        // clear attachSessId if it is no longer present in the found session list
+        const attachSessIdFound = sessions.filter((session) => session.id === attachSessId);
```
mykola-mokhnach, line 960: you may also use .find

app/common/renderer/components/Session/AttachToSession.jsx
```diff
@@ -1,45 +1,43 @@
...
 import SessionStyles from './Session.module.css';
 
-const formatCaps = (caps) => {
-  let importantCaps = [caps.app, caps.platformName, caps.deviceName];
-  if (caps.automationName) {
-    importantCaps.push(caps.automationName);
+const formatCaps = (initialCaps, serverType) => {
+  let returnedCaps = [];
+  const caps =
+    serverType === ServerTypes.lambdatest && 'capabilities' in initialCaps
```
mykola-mokhnach, line 12: this meeds a separate comment with explanation

app/common/renderer/components/Session/AttachToSession.jsx
```diff
@@ -1,45 +1,43 @@
...
+      ? initialCaps.capabilities
+      : initialCaps;
+  // add sessionName (BrowserStack and other cloud providers only)
+  returnedCaps.push(caps.sessionName);
```
mykola-mokhnach, line 16: I assume it should only be pushed if not nil

app/common/renderer/components/Session/AttachToSession.jsx
```diff
@@ -1,45 +1,43 @@
...
-const formatCapsLambdaTest = (caps) => {
-  if (caps.hasOwnProperty.call(caps, 'capabilities')) {
-    caps = caps.capabilities;
-  }
-  const deviceName = caps.desired ? caps.desired.deviceName : caps.deviceName;
-  const importantCaps = [deviceName, caps.platformName, caps.platformVersion];
-  return importantCaps.join(', ').trim();
-};
-
-const getSessionInfo = (session, serverType) => {
```
mykola-mokhnach, line 33: I would still prefer this format as it is more flexible and better readable in comparison to having to having all logic mixed into a single function. Vendor-specific stuff must be somehow separated

app/common/renderer/components/Session/AttachToSession.jsx
eglitise: Since some of the other caps may also end up being nil, I figured it's safer to just filter them out only once the whole list is assembled (on line 35)

app/common/renderer/components/Session/AttachToSession.jsx
eglitise: Created separate methods for standard Appium server and cloud provider sessions. Although it's still a bit messy because some caps need specific handling (e.g. `deviceName`), so it ends up being less complicated to just duplicate most of the code.

app/common/renderer/components/Session/AttachToSession.jsx
```diff
@@ -1,44 +1,68 @@
...
 
-const formatCaps = (caps) => {
-  let importantCaps = [caps.app, caps.platformName, caps.deviceName];
-  if (caps.automationName) {
-    importantCaps.push(caps.automationName);
-  }
-  return importantCaps.join(', ').trim();
+// Format the session string for standard Appium server connections
+const formatStandardCaps = (caps) => {
+  let returnedCaps = [];
```
mykola-mokhnach, line 11: this variable could be const
same below

app/common/renderer/components/Session/AttachToSession.jsx
```diff
@@ -1,44 +1,68 @@
...
-      return `${session.id} - ${formatCapsLambdaTest(session.capabilities)}`;
-    default:
-      return `${session.id} — ${formatCaps(session.capabilities)}`;
-  }
+  const formattedCaps =
+    serverType in [ServerTypes.remote, ServerTypes.local]
+      ? formatStandardCaps(session.capabilities)
+      : formatVendorCaps(session.capabilities, serverType);
+  // omit all null or undefined values
+  const nonEmptyCaps = _.reject(formattedCaps, _.isNil);
```
mykola-mokhnach, line 63: you may use _.compact for this purpose

app/common/renderer/components/Session/AttachToSession.jsx
```diff
@@ -1,44 +1,68 @@
...
-    case ServerTypes.browserstack:
-      return `${session.id} — ${formatCapsBrowserstack(session.capabilities)}`;
-    case ServerTypes.lambdatest:
-      return `${session.id} - ${formatCapsLambdaTest(session.capabilities)}`;
-    default:
-      return `${session.id} — ${formatCaps(session.capabilities)}`;
-  }
+  const formattedCaps =
+    serverType in [ServerTypes.remote, ServerTypes.local]
+      ? formatStandardCaps(session.capabilities)
```
mykola-mokhnach, line 60: this implementation looks better to me, but I believe there is still a way to improve the code.
Lets just walk it step-by-step.
Our input is a capabilities object, whose structure might differ from one vendor to another. We'd like to transform this object into a session description string with a predefined format, where each value is separated with a slash character.
This means we need one function that is going to transform our caps object into the description string.
We know which values we'd like to put into our resulting string. We also know that a way to fetch each value may different from one provider to another.
Lets apply a bit of OOP approach to get what we need and also avoid unneeded code repetition:

```js
class DefaultDescriptionTransformer:
  constructor(caps) {
    this._caps = caps;
  }

  _fetchSessionId() {
      ....
  }

  _fetchDeviceName() {
     ....
  }

  _ fetchApp() {
    ...
  }

  transform() {
    const suffixItems = [
      this._fetchDeviceName(),
      this._fetchApp(),
     ....
    ];
    return `${this. _fetchSessionId() - _.compact(suffixItems).join(' / ')}`;
  }

class LambdaTestDescriptionTransformer(DefaultDescriptionTransformer) {
   // Only override different implementations specific to LT
   _fetchDeviceName() {
    ....
   }
}

class BrowserStackDescriptionTransformer(DefaultDescriptionTransformer) {
   // Only override different implementations specific to BS
}

.....

function getDescriptionTransformer(caps) {
  switch (serverType) {
    case ServerTypes.browserstack:
      return new BrowserStackDescriptionTransformer(caps);
    case ServerTypes.lambdatest:
      return new LambdaTestDescriptionTransformer(caps);
    default:
      return new DefaultDescriptionTransformer(caps);
}

...

const description = getDescriptionTransformer(caps).transform();

```

eglitise: Thanks for the detailed example - I've rewritten the code to follow this approach. I haven't tested it yet though (will only be able to on Monday).

mykola-mokhnach: I believe it would be super easy to write simple unit test(s) for this implementation

app/common/renderer/components/Session/AttachToSession.jsx
```diff
@@ -1,46 +1,83 @@
...
 import React from 'react';
 
 import {ServerTypes} from '../../actions/Session';
 import SessionStyles from './Session.module.css';
 
-const formatCaps = (caps) => {
-  let importantCaps = [caps.app, caps.platformName, caps.deviceName];
-  if (caps.automationName) {
-    importantCaps.push(caps.automationName);
+class DefaultSessionDescription {
```
mykola-mokhnach, line 9: these classes could be extracted to a separate module for better readability

app/common/renderer/components/Session/AttachToSession.jsx
```diff
@@ -1,46 +1,83 @@
...
+class LambdaTestSessionDescription extends DefaultSessionDescription {
+  constructor(caps) {
+    super();
+    if ('capabilities' in caps) {
+      this._caps = caps.capabilities;
+    }
+  }
+
+  _fetchDeviceInfo() {
+    if ('desired' in this._caps) {
```
mykola-mokhnach, line 61: this expression could be simplified using a ternary operator

app/common/renderer/components/Session/AttachToSession.jsx
```diff
@@ -1,46 +1,83 @@
...
+
+  _fetchAutomationName() {
+    return this._caps.automationName;
+  }
+
+  _fetchAppInfo() {
+    return this._caps.app || this._caps.bundleId || this._caps.appPackage;
+  }
+
+  transform() {
```
mykola-mokhnach, line 40: for the given  class names I would rather rename this method to something like `calculate` or `extract`

app/common/renderer/actions/Session.js
```diff
@@ -713,11 +704,15 @@ export function getSavedSessions() {
...
 /**
- * Switch to a different tab
+ * Switch to a different Session Builder tab
  */
 export function switchTabs(key) {
-  return (dispatch) => {
+  return (dispatch, getState) => {
     dispatch({type: SWITCHED_TABS, key});
+    // if switching to Attach to Session tab, also retrieve the running sessions
+    if (key === 'attach') {
```
KazuCocoa, line 713: I don't recall what `key` values could have, but could we define possible values as consts?

KazuCocoa (commented):
Left a minor comment, otherwise lgtm. nice

app/common/renderer/components/Session/AttachToSession.jsx
eglitise: Added 3 unit tests.

app/common/renderer/components/Session/AttachToSession.jsx
eglitise: Changed to `assemble`

eglitise:
While writing unit tests, I had to import the `ServerTypes` values, but since it was not properly defined as a const, I had to make changes in several other files.

app/common/renderer/utils/attaching-to-session.js
```diff
@@ -0,0 +1,71 @@
...
+      this._fetchAutomationName(),
+      this._fetchAppInfo(),
+    ];
+    return _.compact(suffixItems).join(' / ');
+  }
+}
+
+class LambdaTestSessionDescription extends DefaultSessionDescription {
+  constructor(caps) {
+    super();
```
mykola-mokhnach, line 50: `super('capabilities' in caps ? caps.capabilities : caps)`

test/unit/utils-attaching-to-session.spec.js
```diff
@@ -0,0 +1,62 @@
+import {describe, expect, it} from 'vitest';
```
mykola-mokhnach, line 1: 👍

mykola-mokhnach (approved):

app/common/renderer/constants/session-builder.js
```diff
@@ -1,3 +1,28 @@
+export const SESSION_BUILDER_TABS = {
```
KazuCocoa, line 1: 👍

PR #1611, opened 2024-08-16
feat: replace native confirmation popups with antd popconfirms

eglitise:
This PR is a style change for the confirmation dialogs that appear when deleting a saved capability set or gesture.
* The current implementation uses the native browser alert/Electron full window popup, which blocks all interaction with the application until either of the two prompt options are selected.
* The proposed implementation switches to the `antd` `Popconfirm` component, which offers several benefits:
  * More lightweight UX
  * Consistent style with the rest of the app
  * Automatically dismissed upon any action outside the dialog
  * i18n support

Screenshot 2024-08-16 at 16 32 28

Note: antd does not auto-hide tooltips on click, so for saved capability sets, the `zIndex` prop was used to place the tooltip behind the popconfirm.

### DIFF
--- a/app/common/renderer/components/Inspector/SavedGestures.jsx
+++ b/app/common/renderer/components/Inspector/SavedGestures.jsx
@@ -1,5 +1,5 @@
 import {DeleteOutlined, EditOutlined, PlayCircleOutlined, PlusOutlined} from '@ant-design/icons';
-import {Button, Space, Table, Tooltip} from 'antd';
+import {Button, Popconfirm, Space, Table, Tooltip} from 'antd';
 import _ from 'lodash';
 import moment from 'moment';
 import React, {useEffect, useRef} from 'react';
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
Do we only want to replace a single session deletion confirmation popup in this PR or there are more?

eglitise:
These are the only two instances of `window.confirm` in the Inspector code.

eglitise:
Both changes are applied to every element within their lists, so this will replace all instances of the deletion confirmation popups.

mykola-mokhnach (approved):

mykola-mokhnach:
I mean maybe there are other types of alerts, for example informational ones, and we'd like to replace them as well

eglitise:
True, we do use `notification` in various places, but antd also offers `message`, which may be more useful in some locations. That can come in a separate PR though 🙂

PR #1634, opened 2024-08-29
chore: convert Electron modules to IPC

eglitise:
This PR changes Electron's `shell` and `clipboard` modules to be only accessible through IPC. The goal of this change is to make these modules still operational without Electron's remote module, which is removed in Electron 14.
Note that these are not all of the modules that need to be converted: IPC will be required for most of the imports in `app/electron/renderer/polyfills.js`, and the remaining ones are as follows:
* `electron-settings`: possibly requires updating to v5
* Node's built-in `path`, `fs`, `util`: the latter two are only used for `.appiumsession` files, but they may not be required if session import/export is changed to match the import/export approach used for gestures
* `i18next-fs-backend`: not fully sure if this one needs IPC. May also require changing to [i18next-electron-fs-backend](https://github.com/reZach/i18next-electron-fs-backend)

Clipboard and link opening functionality was tested in both Electron and browser versions.

### DIFF
--- a/app/common/renderer/components/ErrorBoundary/ErrorBoundary.jsx
+++ b/app/common/renderer/components/ErrorBoundary/ErrorBoundary.jsx
@@ -1,10 +1,10 @@
 import React from 'react';
 
-import {clipboard} from '../../polyfills';
+import {copyToClipboard} from '../../polyfills';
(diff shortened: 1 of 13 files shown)


jlipps (approved):
yeah, great approach!

PR #1636, opened 2024-08-30
chore: adjust xmldom wrappers for xmldom 0.9

eglitise:
`@xmldom/xmldom` `0.9.0` has a breaking change where it now requires the `mimeType` to be specified. This PR adds the MIME type, and also includes some additional refactoring for exporting only the DOMParser/XMLSerializer methods that are actually used, instead of their whole objects.
Tests were also confirmed to pass after updating to `@xmldom/xmldom` `0.9.0`.

### DIFF
--- a/app/common/renderer/utils/source-parsing.js
+++ b/app/common/renderer/utils/source-parsing.js
@@ -1,10 +1,11 @@
 import {DOMParser, XMLSerializer} from '@xmldom/xmldom';
 import _ from 'lodash';
 
-export const domParser = new DOMParser();
+const domParser = new DOMParser();
 const xmlSerializer = new XMLSerializer();
 
-export const serializeXMLToString = (xml) => xmlSerializer.serializeToString(xml);
+export const xmlToDOM = (string) => domParser.parseFromString(string, 'application/xml');
+export const domToXML = (dom) => xmlSerializer.serializeToString(dom);
 
 /**
  * Get the child nodes of a Node object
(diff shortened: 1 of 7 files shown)


app/common/renderer/utils/source-parsing.js
```diff
@@ -1,8 +1,11 @@
 import {DOMParser, XMLSerializer} from '@xmldom/xmldom';
 import _ from 'lodash';
 
-export const domParser = new DOMParser();
-export const xmlSerializer = new XMLSerializer();
+const domParser = new DOMParser();
+const xmlSerializer = new XMLSerializer();
+
+export const xmlToDOM = (string) => domParser.parseFromString(string, 'application/xml');
```
mykola-mokhnach, line 7: the new version needs to use MIME_TYPE enum, which is defined in the module. See https://github.com/appium/appium/pull/20513/commits/85c01a679f2af1b5b6a9df5b68f2586cb88801b9

eglitise: I wonder if the issue in the Appium repo was caused by the build tool. Here, tests ran fine on 0.9.0 even without the enum.

eglitise: But it is probably safer to use the built-in enums anyway

mykola-mokhnach: I'm sure this is because inspector source don't have typescript checks enabled (I would really like they had)

eglitise: Changed to use the enum, and also updated the dependency itself (might as well)

mykola-mokhnach (approved):

jlipps (approved):

PR #1646, opened 2024-09-03
chore: refactor in preparation for session file handling changes

eglitise:
This is a small refactor in preparation for adjustments to session file handling. These adjustments are already made in a separate branch, but the branch includes too many changes that could be extracted, therefore I made this PR.
Changes made:
* Adjust Electron IPC calls to include scope (basically just replace `-` with `:`)
* Rename redux event `SET_STATE_FROM_SAVED` to `SET_STATE_FROM_FILE`
* Add a dedicated constant for the types that a capability value can take
* Move file handler util methods from `other.js` to a new `file-handling.js` file


KazuCocoa (approved):

jlipps (commented):
👍

PR #1659, opened 2024-09-05
chore(deps): upgrade Electron to 31.4.0

### LINKED ISSUE #1321: feature request: Upgrade electron
### Current Behavior

appium-inspector currently uses electon 13, which reached it's End of Life more than 2 years ago.

### Suggested Solution

Upgrade electron

### Additional Information

https://endoflife.date/electron

eglitise:
This PR (finally) upgrades Electron all the way to the second latest version 31.4.0! ⚡
(latest version 32 has [some known dev-specific issues](https://github.com/electron/electron/issues/43415))

Seems like the only blocker for this was the remote module, the need for which has been eliminated with the IPC migration PRs. I started with Electron 14, and when that worked fine, I continued iterating through every major version, checking whether the app launches fine in dev mode. For Electron 31, I also checked preview and packaged app versions, and tested out basic things like session creation, element selection via screenshot, and element search.

However, every Electron version comes with some breaking changes, and some of them also affect Inspector. The most relevant of these are related to platforms: support for macOS < 10.15 and Windows < 10 has been dropped.

Another notable point is that the packaged Node version has been updated from 14.16 to 20.16, which enables various new features and unblocks some dependency upgrades. These will be handled in separate PRs.

As I mentioned above, I tested only some basic operations, therefore further testing with more complex flows would be greatly appreciated 🤞 

Closes #1321.

### DIFF
--- a/app/common/renderer/components/Inspector/Inspector.module.css
+++ b/app/common/renderer/components/Inspector/Inspector.module.css
@@ -22,8 +22,7 @@
   width: auto !important;
 }
 
-.inspector-toolbar button {
-  border-color: #cacaca;
+.inspector-toolbar :global(.ant-space-item) {
   -webkit-app-region: no-drag;
 }
 
@@ -52,7 +51,6 @@
   width: 100%;
   height: 38px;
   margin-bottom: 12px;
-  -webkit-app-region: drag;
 }
 
 .inspector-main .screenshot-container img {
--- a/app/common/renderer/components/Session/Session.module.css
+++ b/app/common/renderer/components/Session/Session.module.css
@@ -69,6 +69,10 @@
   -webkit-app-region: drag;
 }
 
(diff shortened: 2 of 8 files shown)


mykola-mokhnach (approved):

mykola-mokhnach:
We probably need to bump major pkg version with this release and mention minimum platform requirements in readme

mykola-mokhnach:
> The platform changes will be reflected in the release notes, but I will also add this to the dedicated documentation page.

👍 

> Not sure about additional versioning changes - the version schema does not follow semver anyway.

Leaving it up to you then

jlipps (approved):
nice! so long as we get a few beta testers and make a note of any potential breaking changes, this should not be an issue

KazuCocoa (approved):

eglitise:
Checked most other UI flows and didn't find any issues.

PR #1664, opened 2024-09-06
chore: build for ESM

eglitise:
Electron 28 adds support for building in ESM, and this PR simply enables this functionality. It applies only to the main process and preload script (which is empty in this case).
ESM support also unblocks updating certain dependencies that require it (such as `electron-debug`).

Functionality was confirmed to still work in the macOS and Windows packaged app, and the web version still works fine too.


app/electron/main/updater.js
```diff
@@ -1,5 +1,6 @@
 import {dialog} from 'electron';
-import {autoUpdater} from 'electron-updater';
+import pkg from 'electron-updater';
+const {autoUpdater} = pkg;
```
eglitise, line 3: This module is CJS, and this is the recommended workaround [according to the `electron-vite` docs](https://electron-vite.org/guide/dev#migrating-to-esm).

jlipps (approved):

KazuCocoa (approved):
lgtm

PR #1673, opened 2024-09-10
chore: migrate to ESLint flat config

eglitise:
This PR migrates from the ESLint legacy `eslintrc` format to the new flat config format. Some additional changes were also required here:
* The `import` module was removed as an explicit import, since it was not configured any further, and its rules are already being applied as part of `@appium/eslint-config-appium-ts`
* The `react-native` module was removed, as does not seem to be actively maintained and is unlikely to be updated for ESLint 9 compatibility (and also this codebase does not use React Native anyway)
* The configuration for the `no-unused-vars` rule was removed and instead applied to the `@typescript-eslint/no-unused-vars` rule (from `@appium/eslint-config-appium-ts`), which extends `no-unused-vars`
* The configuration for the `react/jsx-uses-vars` rule was removed, since the `@typescript-eslint/no-unused-vars` rule seems to detect JSX variables just fine
* Since `.eslintignore` is now integrated into the config file, the ignored file patterns have been changed to extend the `.gitignore` file


mykola-mokhnach (approved):
It would be interesting to know if we could do the same for other appium modules. Last time I checked it the typescript stuff was still not compatible with eslint 9

eglitise:
I think we can likely migrate to flat config everywhere, but probably not to ESLint 9.

PR #1698, opened 2024-09-17
fix: adjust calculation of element highlighter positions

### LINKED ISSUE #1680: bug: Screenshot Panel coordinates are incorrect when session is started in landscape mode or the inspector window is resized

### Current Behavior

- When the session is started with the **DUT in landscape mode**, after the screenshot is displayed, the **screenshot panel is automatically resized**, causing the coordinates of elements to become inaccurate. The highlights and interaction areas no longer align with their actual positions on the device screen. This misalignment happens immediately after the panel is resized automatically.
- In **portrait mode**, the issue does not occur until the **Appium Inspector window** is manually resized. After resizing, the coordinates in the screenshot panel become incorrect, making UI element inspection and interaction unreliable.
- Using the **Refresh Source & Screenshot** button resolves the issue in both modes. After manually refreshing, the coordinates are corrected and interactions work as expected.

Screenshot 2024-09-13 at 07 03 57

### Expected Behavior

- The screenshot panel should display accurate coordinates for UI elements, regardless of whether the DUT is in portrait or landscape mode. Any window resizing (manual or automatic) should not break the coordinate mapping.
- Whether the screenshot is in portrait or landscape, and regardless of any resizing events, the coordinates and highlighted areas should always match the actual positions on the device screen without needing to manually refresh.

### Operating System

Mac

### Appium Inspector Version

2024.8.2

### Appium Version

2.11.3

### Further Information

_No response_

eglitise:
This PR fixes the issue where element highlighters over the screenshot were sometimes drawn over incorrect coordinates. This commonly happened when starting a session directly into landscape mode, or resizing the Inspector window height (not width). The underlying cause was that the scale ratio for drawing highlighters was using old data.
The overall implementation has also been slightly simplified, and the unnecessary 500ms debounce was removed, so now the highlighter update is subject to the same 50ms debounce used for calculating the screenshot scale - it feels much more responsive.

Fixes #1680.

### DIFF
--- a/app/common/renderer/components/Inspector/Inspector.jsx
+++ b/app/common/renderer/components/Inspector/Inspector.jsx
@@ -79,7 +79,6 @@ const Inspector = (props) => {
     t,
   } = props;
 
-  const screenAndSourceEl = useRef(null);
   const screenshotContainerEl = useRef(null);
   const mjpegStreamCheckInterval = useRef(null);
 
@@ -95,12 +94,12 @@ const Inspector = (props) => {
     // If the screenshot has too much space to the right or bottom, adjust the max width
     // of its container, so the source tree always fills the remaining space.
     // This keeps everything looking tight.
-    if (!screenAndSourceEl.current) {
+    if (!screenshotContainerEl.current) {
       return;
     }
 
-    const screenshotBox = screenAndSourceEl.current.querySelector('#screenshotContainer');
-    const img = screenAndSourceEl.current.querySelector('#screenshotContainer img#screenshot');
+    const screenshotBox = screenshotContainerEl.current;
+    const img = screenshotContainerEl.current.querySelector('#screenshot');
 
     if (!img) {
       return;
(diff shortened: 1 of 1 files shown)


app/common/renderer/components/Inspector/Inspector.jsx
```diff
@@ -92,23 +90,16 @@ const Inspector = (props) => {
...
+    // If the screenshot has too much space to the right or bottom, adjust the max width
+    // of its container, so the source tree always fills the remaining space.
+    // This keeps everything looking tight.
+    if (!screenshotContainerEl.current) {
       return;
     }
 
-    const screenshotBox = screenAndSourceEl.current.querySelector('#screenshotContainer');
-    const img = screenAndSourceEl.current.querySelector('#screenshotContainer img#screenshot');
+    const screenshotBox = screenshotContainerEl.current;
```
mykola-mokhnach, line 101: should we use screenshotBox everywhere instead of screenshotContainerEl.current ?

mykola-mokhnach (approved):

jlipps (commented):
amazing, thanks for this

PR #1849, opened 2024-12-11
chore: add base for browser build

KazuCocoa:
Add a method to set `--base` for browser build. This is the same as what https://github.com/appium/appium-inspector/pull/1156 did.
Then, if we build with `PUBLIC_URL="/inspector" npm run build:browser:url`, the idea of https://github.com/appium/appium/issues/20792#issuecomment-2530668032 can do with https://github.com/appium/appium-inspector/compare/proto-plugin?expand=1#diff-99b4c701ee2fd34d1c70c79a4b072753f2f47f2436cbb97311f9cb209597b357R8-R22 as a future possible work.

- `dist-browser/index.html` with `PUBLIC_URL="/inspector"`

```
    <script type="module" crossorigin src="/inspector/assets/index-DSPcYjoE.js"></script>
    <link rel="stylesheet" crossorigin href="/inspector/assets/index-CNB9fSg_.css">
```

- default one: `build:browser`
```
    <script type="module" crossorigin src="/assets/index-DSPcYjoE.js"></script>
    <link rel="stylesheet" crossorigin href="/assets/index-CNB9fSg_.css">
```


eglitise (approved):
Not sure how straightforward would it be to add the plugin code without breaking the current builds, but at least this specific change is safe.

PR #1867, opened 2024-12-27
make inspector as appium plugin

saikrishna321:
This PR enables inspector to be published as plugin along side electron app. 

Run the command: 

Build: `npm run build:browser`
Install Plugin: `appium plugin install --source=local $(pwd)/plugins`
Enable Plugin:  `appium server -pa /wd/hub --use-plugins=appium-inspector --allow-cors`

Open URl: http://localhost:4723/inspector

Published the plugin to test it: https://www.npmjs.com/package/appium-inspector-plugin

Can pull it down once we decide to publish from appium org.

cc: @KazuCocoa @Srinivasantarget


package.json
```diff
@@ -32,7 +32,7 @@
     "dev:browser": "vite",
     "dev:electron": "electron-vite -w",
     "build": "npm run build:browser && npm run build:electron",
-    "build:browser": "vite build",
+    "build:browser": "rm -rf plugins/dist-browser && vite build && mv dist-browser plugins/",
```
KazuCocoa, line 35: Can you define a separate command for plugin use? Or 1) move the `rm` part into `clean:browser` command and call `cp` instead of `mv` to keep the original `dist-browser` for non-plugin usage

saikrishna321: @KazuCocoa Fixed!

KazuCocoa (approved):
lg for now to the branch

PR #1868, opened 2024-12-27
feat: enable appium-inspector as an appium plugin

KazuCocoa:
https://github.com/appium/appium/issues/20834

We'd like to make it available as part of regular appium plugin in Appium 3 (mostly for marketing purposes in this case) like `appium plugin install inspector`. Otherwise this plugin can be available always with local installation.

When an appium server start with this plugin, `/inspector` route will return the inspector contents.


plugins/index.js
```diff
@@ -0,0 +1,23 @@
+import { fileURLToPath } from 'node:url';
+import { dirname } from 'node:path';
+import {BasePlugin} from 'appium/plugin.js';
```
eglitise, line 3: Would this import work? Maybe `@appium/base-plugin` needs to be added to dev dependencies?

eglitise: Ah nevermind, this is still a draft. I was confused because CI had run.

plugins/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This package is https://github.com/appium/appium-inspector as an Appium plugin.
+
+```
+appium plugin install --source=local /path/to/appium-inspector/plugins
+```
+
+```
+appium plugin install --source=npm appium-inspector-plugin
+```
```
saikrishna321, line 11: @KazuCocoa It should be appium-inspector instead of appium-inspector-plugin

saikrishna321 (changes requested):

eglitise:
A few small pointers:

I wonder if `build:plugin` can be simplified to avoid the unnecessary copying? Something like:
```
vite build --base /inspector --outDir plugins/dist-browser
```
Would also suggest not modfying `clean:browser` but instead adding `clean:plugin`.
Also, I assume `plugins/dist-browser` should be added to `.gitignore`.

KazuCocoa:
`outDir` sounds good. `dist-browser` in existing `gitignore` already ignored `plugins/dist-browser` as well. (git status didn't show that)

saikrishna321:
@KazuCocoa Linter is failing for ReadMe. Will push a fix.

plugins/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "appium-inspector-plugin",
+  "version": "2024.12.1",
```
mykola-mokhnach, line 3: how this version is supposed to be bumped/synchronized with main inspector version?

plugins/package.json
```diff
@@ -0,0 +1,44 @@
...
+  },
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/appium/appium-inspector/issues"
+  },
+  "keywords": [
+    "appium"
+  ],
+  "homepage": "https://github.com/appium/appium-inspector",
+  "main": "index.js",
```
mykola-mokhnach, line 21: ESM modules require `exports` property to be defined. The `main` one is relevant for CJS

plugins/package.json
```diff
@@ -0,0 +1,44 @@
...
+  },
+  "files": [
+    "index.js",
+    "package.json",
+    "dist-browser",
+    "README.md"
+  ],
+  "dependencies": {},
+  "devDependencies": {
+    "@appium/base-plugin": "^2.2.50"
```
mykola-mokhnach, line 34: Is this a dev dependency?

plugins/index.js
```diff
@@ -0,0 +1,28 @@
...
+import {fileURLToPath} from 'node:url';
+
+import {BasePlugin} from '@appium/base-plugin';
+
+const PLUGIN_ROOT_PATH = '/inspector';
+const INDEX_HTML = 'index.html';
+const ROOT_DIR = path.join(path.dirname(fileURLToPath(import.meta.url)), 'dist-browser');
+
+class AppiumInspectorPlugin extends BasePlugin {
+  constructor(name, cliArgs) {
```
mykola-mokhnach, line 11: would it be possible to add type definitions to method interfaces?

plugins/index.js
```diff
@@ -0,0 +1,28 @@
...
+
+  static async openInspector(req, res) {
+    const reqPath = req.path === PLUGIN_ROOT_PATH ? INDEX_HTML : req.path.substring(10);
+    res.sendFile(reqPath, {root: ROOT_DIR});
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  static async updateServer(expressApp, httpServer, cliArgs) {
+    // We won't return the inspector contents for `/`
+    expressApp.all(/\/inspector.*/, AppiumInspectorPlugin.openInspector);
```
mykola-mokhnach, line 23: express5 does not support regexps. Consider avoiding using them

plugins/package.json
saikrishna321: @mykola-mokhnach Good catch! Actually we don't need this in dependencies nor devDependecies as it comes from peerDependencies 

We can use as `import {BasePlugin} from 'appium/plugin';`

plugins/package.json
saikrishna321: Currently we are manually updating the package.json. If are still sticking to manual then we need to update in both the places. 

@eglitise can add more, I have not done any inspector release before.

eglitise: We might need to add a release script to bump both versions simultaneously. Currently the main version bump is done via `npm version <version>`, which edits `package.json`, adds the version tag, then commits and pushes the changes.

KazuCocoa: Yea, I've been working slowly for this PR but thought to prepare a script to update the `version` in `plugins/package.json` a part of this pr.

plugins/package.json
saikrishna321: @mykola-mokhnach When changed to the above command, we see 

[Appium] Could not load plugin 'inspector', so it will not be available. Error in loading the plugin was: Cannot find module '/Users/saikrishna/Documents/git/appium-inspector/node_modules/appium/plugin' imported from /Users/saikrishna/Documents/git/appium-inspector/plugins/index.js
Did you mean to import "appium/plugin.js"?
[Appium] Error: Cannot find module '/Users/saikrishna/Documents/git/appium-inspector/node_modules/appium/plugin' imported from /Users/saikrishna/Documents/git/appium-inspector/plugins/index.js
Did you mean to import "appium/plugin.js"?

mykola-mokhnach: ESM modules have different import rules from CJS. Please try to add the `.js` suffix as mentioned in the above error message.

plugins/index.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  /**
+   * Handles inspector page requests
+   * @param {import('express').Request} req - Express request object
+   * @param {import('express').Response} res - Express response object
+   * @returns {Promise<void>}
+   */
+  static async openInspector(req, res) {
+    const reqPath = req.path === PLUGIN_ROOT_PATH ? INDEX_HTML : req.path.substring(10);
```
mykola-mokhnach, line 30: what is the meaning of the `10` magic number?

plugins/index.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  /**
+   * Updates the Express server configuration
+   * @param {import('express').Application} expressApp - Express application instance
+   * @param {import('http').Server} httpServer - HTTP server instance
+   * @param {Record<string, unknown>} cliArgs - Command line arguments
+   * @returns {Promise<void>}
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
mykola-mokhnach, line 41: you may simply drop these unused variables from the method signature

plugins/index.js
```diff
@@ -0,0 +1,49 @@
...
+  static async updateServer(expressApp, httpServer, cliArgs) {
+    // Handle both /inspector and /inspector/* paths
+    expressApp.all(['/inspector', '/inspector/*'], AppiumInspectorPlugin.openInspector);
+  }
+}
+
+export {AppiumInspectorPlugin};
```
mykola-mokhnach, line 48: I prefer to use `export class`

scripts/sync-plugin.mjs
```diff
@@ -0,0 +1,34 @@
+import path from 'node:path';
+import {fileURLToPath} from 'node:url';
+
+import {fs} from '@appium/support';
+
+const ROOT_PKG_JSON_PATH = path.join(
```
mykola-mokhnach, line 6: I  would prefer path.resolve

scripts/sync-plugin.mjs
```diff
@@ -0,0 +1,34 @@
+import path from 'node:path';
+import {fileURLToPath} from 'node:url';
+
+import {fs} from '@appium/support';
```
mykola-mokhnach, line 4: consider importing from `appium/support` instead

mykola-mokhnach: you could also use the standard `fs/promises` instead of `support` wrappers

plugins/index.js
```diff
@@ -0,0 +1,49 @@
+import path from 'node:path';
```
mykola-mokhnach, line 1: this file could also have .mjs extension to show it is a ESM module

scripts/sync-plugin.mjs
KazuCocoa: it makes sense to use fs/promises

KazuCocoa:
-> https://github.com/appium/appium-inspector/pull/1878

PR #1873, opened 2024-12-29
chore: adjust routes and fix root behavior

KazuCocoa:
- Return inspector contents only for `/inspector` route
- Remove lock file since this plugin itself doesn't need any deps.
    - `appium plugin install --source=local /path/to/appium-inspector/plugins` doesn't need to run in `plugins`
- Make it work with `appium --use-plugins=inspector`. `appium-` was redundant.

cc @saikrishna321


plugins/index.js
```diff
@@ -16,24 +14,15 @@ class AppiumInspectorPlugin extends BasePlugin {
...
-    res.sendFile(reqPath, {root: `${dirname(fileURLToPath(import.meta.url))}/${DIST_DIR}`});
+    res.sendFile(reqPath, {root: ROOT_DIR});
   }
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   static async updateServer(expressApp, httpServer, cliArgs) {
-    const staticFilesRouter = express.Router();
-    const PUBLIC_DICRECTORY = getPublicDirectory();
-    staticFilesRouter.use(express.static(PUBLIC_DICRECTORY));
-    expressApp.use(staticFilesRouter);
```
saikrishna321, line 27: @KazuCocoa This was added coz the assets folder was not served. Are you able to get the inspector UI without this code?

KazuCocoa: I guess you didn't set `--base /inspector` to build the base browser package. Then, inspector contents refer to `/` as the root. So, you could see requests to `/` without `/inspector` by inspector contents in appium server logs

Issue #1702, opened 2024-09-19
feature request: Use Environment variables for secrets capabilities


itkhanz:
### Current Behavior

Currently, I have different JSON capabilities for starting Appium Session using Appium Inspector. For iOS applications on real device, we need to pass **appium:xcodeOrgId** and **appium:xcodeSigningId** as a session capability. Currently this is hard-coded in all the session capabilities and being repeated across all the sessions. This is not so nice when sharing these capabilities across different teams result in leakage of such secrets. There should be some way to hide or mask these capabilities values.

### Suggested Solution

Ideally these secrets should not be visible in session capabilities, and there should be a way for Appium Inspector to read these values from environment variables.
Possible solution could look like:-

Passing these capabilities as appium:xcodeOrgId = {$XCODE_ORG_ID} where value within brace is treated as environment variable and loaded accordingly. 

Or maybe we could add another dropdown option next to capability key that has option of environment variable for the type of capability and as value of this capability then we simply pass the name of environment variable.

Or alternatively there should be a way to store these secrets in Appium Inspector itself just like how tools like Postman allow user to store secrets and then use them in tool.

### Additional Information

Additionally, a rather simple UI enhancement can be to have a checkmark next to such secrets capabilities that when checked masks the capability value.

eglitise:
I like this idea, though this information will not be hidden from the Appium server logs ([unless the server is explicitly instructed to hide it](http://appium.io/docs/en/latest/guides/log-filters/)). It would still make it safer to share such capabilities, though.

saikrishna321:
@eglitise FYI

@saikrishna321 @SrinivasanTarget will pick this.

KazuCocoa:
Btw, is the running appium server working remotely, or is the target appium server long-running process?

Actually, it depends on the usage but I think starting an appium server process with `--default-capabilities` instead of sending them over http/s is more secure in terms of setting secrets (and using `--log-filters` to mask them from the logs as well as https://github.com/appium/appium-inspector/issues/1702#issuecomment-2361593880)

PR #1877, opened 2024-12-31
Feat: Use Environment variables for secrets capabilities #1702

SrinivasanTarget:
Addressing #1702 

Environment variable tab:
image

Usage: `"appium:testing": "${asdf}",`

image

We will address export and import of the secrets in a different PR.

### DIFF
--- a/app/common/renderer/actions/Inspector.js
+++ b/app/common/renderer/actions/Inspector.js
@@ -116,6 +116,9 @@ export const TOGGLE_SHOW_ATTRIBUTES = 'TOGGLE_SHOW_ATTRIBUTES';
 export const TOGGLE_REFRESHING_STATE = 'TOGGLE_REFRESHING_STATE';
 
 export const SET_GESTURE_UPLOAD_ERROR = 'SET_GESTURE_UPLOAD_ERROR';
+export const SET_ENVIRONMENT_VARIABLES = 'SET_ENVIRONMENT_VARIABLES';
+export const ADD_ENVIRONMENT_VARIABLE = 'ADD_ENVIRONMENT_VARIABLE';
+export const DELETE_ENVIRONMENT_VARIABLE = 'DELETE_ENVIRONMENT_VARIABLE';
 
 const KEEP_ALIVE_PING_INTERVAL = 20 * 1000;
 const NO_NEW_COMMAND_LIMIT = 24 * 60 * 60 * 1000; // Set timeout to 24 hours
@@ -916,6 +919,24 @@ export function setAwaitingMjpegStream(isAwaiting) {
 export function setGestureUploadErrors(errors) {
   return (dispatch) => {
     dispatch({type: SET_GESTURE_UPLOAD_ERROR, errors});
+  };
+}
+
+export function setEnvironmentVariables(envVars) {
+  return (dispatch) => {
+    dispatch({type: SET_ENVIRONMENT_VARIABLES, envVars});
+  };
+}
+
+export function addEnvironmentVariable(key, value) {
+  return (dispatch) => {
+    dispatch({type: ADD_ENVIRONMENT_VARIABLE, key, value});
+  };
+}
+
+export function deleteEnvironmentVariable(key) {
+  return (dispatch) => {
+    dispatch({type: DELETE_ENVIRONMENT_VARIABLE, key});
   };
 }
 
--- a/app/common/renderer/actions/Session.js
+++ b/app/common/renderer/actions/Session.js
@@ -1,5 +1,6 @@
 import {notification} from 'antd';
 import {includes, isPlainObject, isUndefined, toPairs, union, values, without} from 'lodash';
+import {interpolateEnvironmentVariables} from '../utils/env-utils';
 import moment from 'moment';
 import {Web2Driver} from 'web2driver';
 
@@ -229,7 +230,12 @@ export function newSession(caps, attachSessId = null) {
 
     dispatch({type: NEW_SESSION_REQUESTED, caps});
 
+    // Get environment variables from state
+    const environmentVariables = getState().inspector.environmentVariables || [];
+    
+    // Get capabilities and interpolate environment variables
     let desiredCapabilities = caps ? getCapsObject(caps) : {};
+    desiredCapabilities = interpolateEnvironmentVariables(desiredCapabilities, environmentVariables);
     let host, port, username, accessKey, https, path, token;
     desiredCapabilities = addCustomCaps(desiredCapabilities);
 
--- a/app/common/renderer/components/Inspector/EnvironmentVariables.jsx
+++ b/app/common/renderer/components/Inspector/EnvironmentVariables.jsx
@@ -1,0 +1,93 @@
+import {PlusOutlined, DeleteOutlined} from '@ant-design/icons';
+import {Button, Input, Space, Table, Tooltip} from 'antd';
+import {useState} from 'react';
+import {useDispatch, useSelector} from 'react-redux';
+import {setEnvironmentVariables, addEnvironmentVariable, deleteEnvironmentVariable} from '../../actions/Inspector';
+import styles from './EnvironmentVariables.module.css';
+
+const EnvironmentVariables = ({t}) => {
+  const dispatch = useDispatch();
+  const envVars = useSelector(state => state.inspector.environmentVariables || []);
+  const [newVar, setNewVar] = useState({key: '', value: ''});
+
+  const columns = [
+    {
+      title: t('Variable Name'),
+      dataIndex: 'key',
+      key: 'key',
+      width: '40%',
+    },
+    {
+      title: t('Value'),
+      dataIndex: 'value',
+      key: 'value',
+      width: '40%',
+      render: (text) => <Input.Password value={text} readOnly />,
+    },
+    {
+      title: t('Actions'),
+      key: 'action',
+      width: '20%',
+      render: (_, record) => (
+        <Tooltip title={t('Delete Variable')}>
+          <Button
+            type="text"
+            danger
+            icon={<DeleteOutlined />}
+            onClick={() => {
+              dispatch(deleteEnvironmentVariable(record.key));
+            }}
+          />
+        </Tooltip>
+      ),
+    },
+  ];
+
+  const handleAddVariable = () => {
+    if (!newVar.key || !newVar.value) return;
+    
+    // Check for duplicate keys
+    if (envVars.some((v) => v.key === newVar.key)) {
+      return;
+    }
+
+    dispatch(addEnvironmentVariable(newVar.key, newVar.value));
+    setNewVar({key: '', value: ''});
+  };
+
+  return (
+    <div className={styles.container}>
+      <div className={styles.addForm}>
+        <Space.Compact style={{width: '100%'}}>
+          <Input
+            placeholder={t('Variable Name')}
+            value={newVar.key}
+            onChange={(e) => setNewVar({...newVar, key: e.target.value})}
+          />
+          <Input.Password
+            placeholder={t('Value')}
+            value={newVar.value}
+            onChange={(e) => setNewVar({...newVar, value: e.target.value})}
+          />
+          <Button
+            type="primary"
+            icon={<PlusOutlined />}
+            onClick={handleAddVariable}
+            disabled={!newVar.key || !newVar.value}
+          >
+            {t('Add')}
+          </Button>
+        </Space.Compact>
+      </div>
+      <Table
+        columns={columns}
+        dataSource={envVars}
+        pagination={false}
+        rowKey="key"
+        size="small"
+      />
+    </div>
+  );
+};
+
+export default EnvironmentVariables;
--- a/app/common/renderer/components/Inspector/Inspector.jsx
+++ b/app/common/renderer/components/Inspector/Inspector.jsx
@@ -15,6 +15,7 @@ import {
 import {Button, Card, Modal, Space, Spin, Switch, Tabs, Tooltip} from 'antd';
 import {debounce} from 'lodash';
 import {useEffect, useRef, useState} from 'react';
+import styles from './Inspector.module.css';
 import {useNavigate} from 'react-router';
 
 import {BUTTON} from '../../constants/antd-types';
@@ -30,7 +31,6 @@ import {downloadFile} from '../../utils/file-handling';
 import Commands from './Commands.jsx';
 import GestureEditor from './GestureEditor.jsx';
 import HeaderButtons from './HeaderButtons.jsx';
-import InspectorStyles from './Inspector.module.css';
 import Recorder from './Recorder.jsx';
 import SavedGestures from './SavedGestures.jsx';
 import Screenshot from './Screenshot.jsx';
@@ -221,7 +221,7 @@ const Inspector = (props) => {
   }, [showKeepAlivePrompt]);
 
   const screenShotControls = (
-    <div className={InspectorStyles['screenshot-controls']}>
+    <div className={styles['screenshot-controls']}>
       <Space size="middle">
         <Tooltip
           title={t(showCentroids ? 'Hide Element Handles' : 'Show Element Handles')}
@@ -263,10 +263,10 @@ const Inspector = (props) => {
   );
 
   const main = (
-    <div className={InspectorStyles['inspector-main']}>
+      <div className={styles['inspector-main']}>
       <div
         id="screenshotContainer"
-        className={InspectorStyles['screenshot-container']}
+        className={styles['screenshot-container']}
         ref={screenshotContainerEl}
       >
         {screenShotControls}
@@ -274,11 +274,11 @@ const Inspector = (props) => {
         {screenshotError && t('couldNotObtainScreenshot', {screenshotError})}
         {!showScreenshot && (
           <Spin size="large" spinning={true}>
-            <div className={InspectorStyles.screenshotBox} />
+            <div className={styles.screenshotBox} />
           </Spin>
         )}
       </div>
-      <div id="sourceTreeContainer" className={InspectorStyles['interaction-tab-container']}>
+      <div id="sourceTreeContainer" className={styles['interaction-tab-container']}>
         <Tabs
           activeKey={selectedInspectorTab}
           size="small"
@@ -331,7 +331,7 @@ const Inspector = (props) => {
                   </div>
                   <div
                     id="selectedElementContainer"
-                    className={`${InspectorStyles['interaction-tab-container']} ${InspectorStyles['element-detail-container']} action-col`}
+                    className={`${styles['interaction-tab-container']} ${styles['element-detail-container']} action-col`}
                   >
                     <Card
                       title={
@@ -339,7 +339,7 @@ const Inspector = (props) => {
                           <TagOutlined /> {t('selectedElement')}
                         </span>
                       }
-                      className={InspectorStyles['selected-element-card']}
+                      className={styles['selected-element-card']}
                     >
                       {selectedElement.path && <SelectedElement {...props} />}
                       {!selectedElement.path && <i>{t('selectElementInSource')}</i>}
@@ -359,7 +359,7 @@ const Inspector = (props) => {
                       <ThunderboltOutlined /> {t('Execute Commands')}
                     </span>
                   }
-                  className={InspectorStyles['interaction-tab-card']}
+                  className={styles['interaction-tab-card']}
                 >
                   <Commands {...props} />
                 </Card>
@@ -376,7 +376,7 @@ const Inspector = (props) => {
                       <HighlightOutlined /> {t('Gesture Builder')}
                     </span>
                   }
-                  className={InspectorStyles['interaction-tab-card']}
+                  className={styles['interaction-tab-card']}
                 >
                   <GestureEditor {...props} />
                 </Card>
@@ -387,7 +387,7 @@ const Inspector = (props) => {
                       <HighlightOutlined /> {t('Saved Gestures')}
                     </span>
                   }
-                  className={InspectorStyles['interaction-tab-card']}
+                  className={styles['interaction-tab-card']}
                 >
                   <SavedGestures {...props} />
                 </Card>
@@ -410,7 +410,7 @@ const Inspector = (props) => {
                       <InfoCircleOutlined /> {t('Session Information')}
                     </span>
                   }
-                  className={InspectorStyles['interaction-tab-card']}
+                  className={styles['interaction-tab-card']}
                 >
                   <SessionInfo {...props} />
                 </Card>
@@ -423,7 +423,7 @@ const Inspector = (props) => {
   );
 
   return (
-    <div className={InspectorStyles['inspector-container']}>
+    <div className={styles['inspector-container']}>
       <HeaderButtons quitCurrentSession={quitCurrentSession} {...props} />
       {main}
       <Modal
--- a/app/common/renderer/components/Session/Inspector.module.css
+++ b/app/common/renderer/components/Session/Inspector.module.css
@@ -1,0 +1,734 @@
+.inspector-container {
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+}
+
+.inspector-toolbar {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: center;
+  width: 100%;
+  height: 50px;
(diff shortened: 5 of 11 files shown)


app/common/renderer/utils/env-utils.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * Recursively interpolate environment variables in an object
+ * @param {*} obj - The object to interpolate
+ * @param {Array} envVars - Array of environment variables [{key: string, value: string}]
+ * @returns {*} - The interpolated object
+ */
+export function interpolateEnvironmentVariables(obj, envVars) {
```
mykola-mokhnach, line 7: what is the purpose of this method?

mykola-mokhnach:
I am not sure if it makes sense to make all env variables secret. I assume there might be some which values we still would like to inspect in plain text

app/common/renderer/utils/env-utils.js
SrinivasanTarget: this method helps in replacing the ${asdf} to actual value before sending to create session.

SrinivasanTarget:
> I am not sure if it makes sense to make all env variables secret. I assume there might be some which values we still would like to inspect in plain text

Currently both environment variables and secrets are same and in frontend we just mask it whereas it remains plaintext in appium logs unless user explicitly filters it in logs.

app/common/renderer/utils/env-utils.js
mykola-mokhnach: it would make sense to cover this method with a unit test

app/common/renderer/components/Inspector/EnvironmentVariables.jsx
```diff
@@ -0,0 +1,93 @@
+import {PlusOutlined, DeleteOutlined} from '@ant-design/icons';
+import {Button, Input, Space, Table, Tooltip} from 'antd';
+import {useState} from 'react';
+import {useDispatch, useSelector} from 'react-redux';
+import {setEnvironmentVariables, addEnvironmentVariable, deleteEnvironmentVariable} from '../../actions/Inspector';
```
eglitise, line 5: These methods can be retrieved from `EnvironmentVariables.props` instead of being explicitly imported. Check other component files for reference.

app/common/renderer/components/Inspector/EnvironmentVariables.jsx
```diff
@@ -0,0 +1,93 @@
...
+import styles from './EnvironmentVariables.module.css';
```
eglitise, line 6: Not sure why this is a separate file instead of reusing `Inspector.module.css`

app/common/renderer/components/Inspector/EnvironmentVariables.jsx
```diff
@@ -0,0 +1,93 @@
+import {PlusOutlined, DeleteOutlined} from '@ant-design/icons';
+import {Button, Input, Space, Table, Tooltip} from 'antd';
+import {useState} from 'react';
+import {useDispatch, useSelector} from 'react-redux';
```
eglitise, line 4: I'm not fully sure about the suitability of these methods since the rest of the codebase still uses the `connect` API

app/common/renderer/components/Inspector/EnvironmentVariables.jsx
```diff
@@ -0,0 +1,93 @@
...
+        <Tooltip title={t('Delete Variable')}>
+          <Button
+            type="text"
+            danger
+            icon={<DeleteOutlined />}
+            onClick={() => {
+              dispatch(deleteEnvironmentVariable(record.key));
+            }}
+          />
+        </Tooltip>
```
eglitise, line 41: Would suggest reusing the popconfirm implementation used for sessions and gestures

app/common/renderer/components/Session/Inspector.module.css
```diff
@@ -0,0 +1,734 @@
+.inspector-container {
```
eglitise, line 1: Not sure why this file was duplicated from the `Inspector` directory

app/common/renderer/components/Session/Session.jsx
```diff
@@ -11,6 +12,9 @@ import {
   SERVER_TYPES,
   SESSION_BUILDER_TABS,
 } from '../../constants/session-builder';
+import {
+  INSPECTOR_TABS,
+} from '../../constants/session-inspector';
```
eglitise, line 17: This tab is in the session builder so the `SESSION_BUILDER_TABS` constant is the one that should be changed, not `INSPECTOR_TABS`

app/common/renderer/components/Session/Session.jsx
```diff
@@ -157,6 +163,22 @@ const Session = (props) => {
               className: SessionStyles.scrollingTab,
               children: <AttachToSession {...props} />,
             },
+            {
```
eglitise, line 166: Please make this consistent with the implementation of other tabs (using `SESSION_BUILDER_TABS` & `SessionStyles`)

app/common/renderer/reducers/Inspector.jsx
```diff
@@ -162,6 +196,7 @@ export default function inspector(state = INITIAL_STATE, action) {
     case QUIT_SESSION_DONE:
       return {
         ...INITIAL_STATE,
+        environmentVariables: state.environmentVariables,
```
eglitise, line 199: Since we don't want env vars to be cleared here, I think it's better to store them in persistent settings (see `app/common/shared/setting-defs.js`)

eglitise:
Also I don't really see the need of renaming `reducers/Inspector.js` to `reducers/Inspector.jsx`. The Redux website does not use this extension in their examples, and it would look out of place among its adjacent `.js` files.

app/common/renderer/components/Inspector/EnvironmentVariables.jsx
```diff
@@ -0,0 +1,103 @@
...
+const mapStateToProps = (state) => ({
+  envVars: state.inspector.environmentVariables || [],
+});
+
+const mapDispatchToProps = (dispatch) => ({
+  addVariable: (key, value) => dispatch(addEnvironmentVariable(key, value)),
+  deleteVariable: (key) => dispatch(deleteEnvironmentVariable(key)),
+});
+
+export default connect(mapStateToProps, mapDispatchToProps)(EnvironmentVariables);
```
eglitise, line 103: None of this code should be necessary - the `connect` API is already being called in `app/common/renderer/containers`. Check adjacent component files for comparison.

app/common/renderer/components/Inspector/EnvironmentVariables.jsx
```diff
@@ -0,0 +1,103 @@
...
+    },
+    {
+      title: t('Actions'),
+      key: 'action',
+      width: '20%',
+      render: (_, record) => (
+        <Tooltip zIndex={3} title={t('Delete Variable')}>
+          <Popconfirm
+            zIndex={4}
+            title={t('Are you sure you want to delete this variable?')}
```
eglitise, line 38: Can reuse `t('confirmDeletion')`

app/common/renderer/components/Inspector/EnvironmentVariables.jsx
```diff
@@ -0,0 +1,103 @@
...
+            placement="topRight"
+            okText={t('OK')}
+            cancelText={t('Cancel')}
+            onConfirm={() => deleteVariable(record.key)}
+          >
+            <Button type="text" danger icon={<DeleteOutlined />} />
```
eglitise, line 44: Does it need to have the `type="text"` style? IMO it's better to be consistent with other uses of the delete button

app/common/renderer/components/Session/Session.jsx
```diff
@@ -157,6 +161,22 @@ const Session = (props) => {
...
+              label: t('Environment Variables'),
+              key: SESSION_BUILDER_TABS.ENV_VARS,
+              children: (
```
eglitise, line 167: Can the `<Card>` part also be included in the `EnvironmentVariables` component, for consistency?
Also, again in terms of consistency, this is missing `className: SessionStyles.scrollingTab`

app/common/renderer/reducers/Inspector.jsx
```diff
@@ -87,13 +91,18 @@ import {
 } from '../actions/Inspector';
 import {SCREENSHOT_INTERACTION_MODE} from '../constants/screenshot';
 import {APP_MODE, INSPECTOR_TABS, NATIVE_APP} from '../constants/session-inspector';
+import {getSetting} from '../polyfills';
 
 const DEFAULT_FRAMEWORK = 'java';
 
+// Load initial environment variables from persistent settings
+const envVars = (await getSetting(ENVIRONMENT_VARIABLES)) || [];
```
eglitise, line 99: I don't think this is necessary if the env vars are already being retrieved in `await loadEnvironmentVariables()`. Compare against how saved sessions or visible providers are retrieved.

app/common/renderer/reducers/Inspector.jsx
```diff
@@ -133,8 +142,40 @@ const INITIAL_STATE = {
 
 let nextState;
 
+// Helper function to interpolate environment variables in a string
+const interpolateEnvVars = (str, envVars) => {
```
eglitise, line 146: Duplicate of `interpolateEnvironmentVariables()`

app/common/renderer/reducers/Inspector.jsx
```diff
@@ -133,8 +142,40 @@ const INITIAL_STATE = {
...
+  }
+  return str.replace(/\${([^}]+)}/g, (match, key) => {
+    const envVar = envVars.find((v) => v.key === key);
+    return envVar ? envVar.value : match;
+  });
+};
+
 export default function inspector(state = INITIAL_STATE, action) {
   switch (action.type) {
+    case SET_ENVIRONMENT_VARIABLES:
```
eglitise, line 158: All of this code should be in the `Session` reducer

app/common/renderer/components/Inspector/EnvironmentVariables.jsx
```diff
@@ -0,0 +1,103 @@
+import {DeleteOutlined, PlusOutlined} from '@ant-design/icons';
```
eglitise, line 1: This entire file should be in the `Session` directory, as that corresponds to Session Builder. The `Inspector` directory corresponds to the Session Inspector. Though I realise the directory names are not obvious.

app/common/renderer/components/Inspector/Inspector.jsx
```diff
@@ -30,7 +30,7 @@ import {downloadFile} from '../../utils/file-handling';
 import Commands from './Commands.jsx';
 import GestureEditor from './GestureEditor.jsx';
 import HeaderButtons from './HeaderButtons.jsx';
-import InspectorStyles from './Inspector.module.css';
```
eglitise, line 33: All of the changes in this file are renaming and can be omitted in this PR

app/common/renderer/components/Inspector/Inspector.module.css
```diff
@@ -732,3 +732,13 @@
   color: red;
   margin: 5px;
 }
+
+.container {
```
eglitise, line 736: This code should be in `Session.module.css` as mentioned in other comments. Also the style names are quite generic and should be linked more closely to the env vars tab

app/common/renderer/components/Session/Session.jsx
```diff
@@ -40,6 +41,7 @@ const Session = (props) => {
     savedSessions,
     newSessionLoading,
     attachSessId,
+    loadEnvironmentVariables,
```
eglitise, line 44: This line is not needed since the method is already retrieved on line 78

app/common/renderer/actions/Inspector.js
```diff
@@ -357,6 +364,9 @@ export function quitSession(reason, killedByUser = true) {
     const applyAction = applyClientMethod({methodName: 'quit'});
     await applyAction(dispatch, getState);
     dispatch({type: QUIT_SESSION_DONE});
+    // Reload environment variables from persistent settings after session ends
+    const loadEnvAction = loadEnvironmentVariables();
+    await loadEnvAction(dispatch);
```
eglitise, line 369: Not sure if this code is necessary

app/common/renderer/actions/Inspector.js
```diff
@@ -919,6 +929,38 @@ export function setGestureUploadErrors(errors) {
   };
 }
 
+export function setEnvironmentVariables(envVars) {
```
eglitise, line 932: As mentioned in other comments, all of these functions should be in the `Session.js` file

app/common/renderer/components/Session/Session.jsx
eglitise: I'm also not entirely convinced about wrapping the whole UI component in a `Card`, as, again, this looks very inconsistent with the other tabs

app/common/renderer/reducers/Inspector.jsx
```diff
@@ -1,14 +1,17 @@
 import {omit} from 'lodash';
 
+import {ENVIRONMENT_VARIABLES} from '../../shared/setting-defs';
```
eglitise, line 3: This file is still renamed to have the `.jsx` extension which is not necessary

test/unit/utils-env.spec.js
```diff
@@ -0,0 +1,79 @@
+import {describe, expect, it} from 'vitest';
```
eglitise, line 1: 👍

eglitise:
Also, maybe it's better to use a `Map` for storing the env vars, instead of an array of objects? It should simplify/speed up the duplicate key/value checks.

SrinivasanTarget:
@eglitise Can you please help review this as we have other PR's lined up?

eglitise:
@SrinivasanTarget I will check this once I have some time, likely this weekend. I don't see how this should block you from creating new PRs though.

eglitise:
One thing I can point out right now though - the approach of storing these vars in persistent settings. I thought about it some more and came to the conclusion that my suggestion about this was wrong. I'm not sure why these vars should be retained after quitting the session, and we can just reapply them when starting the next session. That lead me to think that the best way to store them would be in the saved session object, alongside the server and capability details.
What do you think @SrinivasanTarget @saikrishna321 ?

SrinivasanTarget:
> @SrinivasanTarget I will check this once I have some time, likely this weekend. I don't see how this should block you from creating new PRs though.

No worries take your time. I wanted to avoid conflicts as i will touching the same files.

SrinivasanTarget:
> One thing I can point out right now though - the approach of storing these vars in persistent settings. I thought about it some more and came to the conclusion that my suggestion about this was wrong. I'm not sure why these vars should be retained after quitting the session, and we can just reapply them when starting the next session. That lead me to think that the best way to store them would be in the saved session object, alongside the server and capability details.
> What do you think @SrinivasanTarget @saikrishna321 ?

@eglitise I would prefer to keep it persistent so that between multiple session creations users doesnt have to re-enter env variables again and save it. Also what if server (inspector as plugin) is hosted at a central place in an org and environment variables (aka secrets). do correct me if im wrong.

eglitise:
@SrinivasanTarget the user would not need to re-enter the env vars because they would be stored in the saved session object. They would only need to select the entry in the Saved Capability Sets tab (which should be renamed, but not in this PR), and the Inspector would apply the stored env vars the same way as it currently applies server and capability details. This approach would also allow specifying different sets of env vars for different session objects.

app/common/renderer/components/Session/Session.jsx
```diff
@@ -157,6 +163,19 @@ const Session = (props) => {
...
+            {
+              label: t('Environment Variables'),
+              key: SESSION_BUILDER_TABS.ENV_VARS,
+              className: SessionStyles.scrollingTab,
+              children: (
+                <EnvironmentVariables
+                  t={t}
+                  envVars={environmentVariables}
+                  addVariable={addEnvironmentVariable}
+                  deleteVariable={deleteEnvironmentVariable}
```
eglitise, line 175: You don't need to explicitly pass these methods, they should be accessible in `EnvironmentVariables` by default

app/common/renderer/components/Session/Session.jsx
```diff
@@ -82,6 +87,7 @@ const Session = (props) => {
         await setVisibleProviders();
         await setSavedServerParams();
         await setLocalServerParams();
+        await loadEnvironmentVariables();
```
eglitise, line 90: This line isn't needed since the env vars are only relevant during an active session

app/common/renderer/actions/Inspector.js
eglitise: As mentioned above, I don't think this code is necessary because we only care about env vars in the context of an active session. Quitting a session won't delete them from tthe persistent settings in any case.

app/common/renderer/components/Inspector/Inspector.jsx
eglitise: This is still not resolved

app/common/renderer/reducers/Inspector.js
```diff
@@ -659,6 +665,29 @@ export default function inspector(state = INITIAL_STATE, action) {
...
+      };
+
+    case DELETE_ENVIRONMENT_VARIABLE:
+      return {
+        ...state,
+        environmentVariables: (state.environmentVariables || []).filter(
+          (envVar) => envVar.key !== action.key,
+        ),
+      };
+
```
eglitise, line 690: This code can be removed since the Session reducer handles this functionality. The only use of this code is in `loadEnvironmentVariables` in `actions/Inspector.js`, but as I commented under that code, that method can also be removed.

app/common/renderer/reducers/Session.js
```diff
@@ -75,12 +78,34 @@ const INITIAL_STATE = {
   isValidatingCapsJson: false,
   isAddingCloudProvider: false,
   addVendorPrefixes: true,
+  environmentVariables: [],
```
eglitise, line 81: The default value here can be changed to `new Map()`, so you can avoid unnecessary type conversion in the set/add/delete actions

app/common/renderer/actions/Session.js
```diff
@@ -1190,3 +1215,35 @@ export function initFromQueryString(loadNewSession) {
     }
   };
 }
+
+export function setEnvironmentVariables(envVars) {
```
eglitise, line 1219: This method isn't used anywhere

app/common/renderer/actions/Session.js
```diff
@@ -1190,3 +1215,35 @@ export function initFromQueryString(loadNewSession) {
...
+  return async (dispatch) => {
+    await setSetting(ENVIRONMENT_VARIABLES, envVars);
+    dispatch({type: SET_ENVIRONMENT_VARIABLES, envVars});
+  };
+}
+
+export function loadEnvironmentVariables() {
```
eglitise, line 1226: This should probably be renamed to `setEnvironmentVariables` for consistency with `setVisibleProviders`. I was initially confused by its call in `components/Session/Session.jsx`.

app/common/renderer/actions/Session.js
```diff
@@ -1190,3 +1215,35 @@ export function initFromQueryString(loadNewSession) {
...
+  return async (dispatch) => {
+    const envVars = (await getSetting(ENVIRONMENT_VARIABLES)) || [];
+    dispatch({type: SET_ENVIRONMENT_VARIABLES, envVars});
+  };
+}
+
+export function addEnvironmentVariable(key, value) {
+  return async (dispatch, getState) => {
+    const currentEnvVars = getState().inspector.environmentVariables || [];
```
eglitise, line 1235: I don't see why these are retrieved from the inspector reducer (they aren't relevant anymore at that point)
```suggestion
    const currentEnvVars = getState().session.environmentVariables || [];
```

app/common/renderer/actions/Session.js
```diff
@@ -1190,3 +1215,35 @@ export function initFromQueryString(loadNewSession) {
...
+    const newEnvVars = [...currentEnvVars, {key, value}];
+    await setSetting(ENVIRONMENT_VARIABLES, newEnvVars);
+    dispatch({type: ADD_ENVIRONMENT_VARIABLE, key, value});
+  };
+}
+
+export function deleteEnvironmentVariable(key) {
+  return async (dispatch, getState) => {
+    const currentEnvVars = getState().inspector.environmentVariables || [];
```
eglitise, line 1244: Same here
```suggestion
    const currentEnvVars = getState().session.environmentVariables;
```

app/common/renderer/actions/Session.js
```diff
@@ -229,7 +236,25 @@ export function newSession(caps, attachSessId = null) {
 
     dispatch({type: NEW_SESSION_REQUESTED, caps});
 
+    // Get environment variables from state
+    const environmentVariables = session.environmentVariables || [];
```
eglitise, line 240: Handling for default value is not needed since it is guaranteed to be set

app/common/renderer/components/Session/EnvironmentVariables.jsx
```diff
@@ -0,0 +1,101 @@
...
+          pagination={false}
+          rowKey="key"
+          size="small"
+        />
+      </div>
+    </Card>
+  );
+};
+
+export default EnvironmentVariables;
```
eglitise, line 101: I'm not really a fan of the `Card` view, how about this?
I've also adjusted the table for better handling of more rows (added starting point for scroll + sticky header)
```suggestion
import {DeleteOutlined, PlusOutlined} from '@ant-design/icons';
import {Button, Col, Form, Input, Popconfirm, Row, Table, Tooltip} from 'antd';
import {useState} from 'react';

const EnvironmentVariables = ({t, envVars, addVariable, deleteVariable}) => {
  const [newVar, setNewVar] = useState({key: '', value: ''});

  const tableData = Array.from(envVars, ([key, value]) => ({key, value}));

  const columns = [
    {
      title: t('Variable Name'),
      dataIndex: 'key',
      key: 'key',
      width: '45%',
    },
    {
      title: t('Value'),
      dataIndex: 'value',
      key: 'value',
      width: '45%',
      render: (text) => <Input.Password value={text} readOnly />,
    },
    {
      title: t('Actions'),
      key: 'action',
      width: '10%',
      render: (_, record) => (
        <Tooltip zIndex={3} title={t('Delete')}>
          <Popconfirm
            zIndex={4}
            title={t('confirmDeletion')}
            placement="topRight"
            okText={t('OK')}
            cancelText={t('Cancel')}
            onConfirm={() => deleteVariable(record.key)}
          >
            <Button icon={<DeleteOutlined />} />
          </Popconfirm>
        </Tooltip>
      ),
    },
  ];

  const handleAddVariable = () => {
    if (!newVar.key || !newVar.value) {
      return;
    }

    // Check for duplicate keys is no longer needed since Map handles this automatically
    addVariable(newVar.key, newVar.value);
    setNewVar({key: '', value: ''});
  };

  return (
    <>
      <Row gutter={8}>
        <Col span={11}>
          <Form.Item>
            <Input
              placeholder={t('Variable Name')}
              value={newVar.key}
              onChange={(e) => setNewVar({...newVar, key: e.target.value})}
            />
          </Form.Item>
        </Col>
        <Col span={11}>
          <Form.Item>
            <Input.Password
              placeholder={t('Value')}
              value={newVar.value}
              onChange={(e) => setNewVar({...newVar, value: e.target.value})}
            />
          </Form.Item>
        </Col>
        <Col span={2}>
          <Form.Item>
            <Button
              block
              type="primary"
              icon={<PlusOutlined />}
              onClick={handleAddVariable}
              disabled={!newVar.key || !newVar.value}
            >
              {t('Add')}
            </Button>
          </Form.Item>
        </Col>
      </Row>
      <Table
        columns={columns}
        dataSource={tableData}
        pagination={false}
        sticky={true}
        rowKey="key"
        scroll={{y: 'calc(100vh - 34em)'}}
        size="small"
      />
    </>
  );
};

export default EnvironmentVariables;
```

app/common/renderer/components/Session/Session.module.css
```diff
@@ -316,3 +316,19 @@
   color: #ff4d4f;
   font-size: 12px;
 }
+
```
eglitise, line 319: If you agree with my suggestion for `components/Session/EnvironmentVariables.jsx`, all of these changes can be removed

PKRITHQA (approved):

eglitise:
I am closing this as abandoned, but feel free to resubmit an updated version of this feature.

PR #1878, opened 2025-01-01
feat: enable appium-inspector as an appium plugin

KazuCocoa:
Complete partially in https://github.com/appium/appium/issues/20834

Final PR instead of https://github.com/appium/appium-inspector/pull/1868, which was ordinally just for a prototype.

After this PR merge, we should run `npm publish` as README's `release` section in this PR if needed for now. I'll also create a follow-up PR to make it automated as part of the existing workflow https://github.com/appium/appium-inspector/blob/main/.github/workflows/package.yml to push the npm package as well.

- make this inspector's web-based contents via `/inspector` endpoint
- available via `npm`
     - `@appium/inspector-plugin` or `appium-inspector-plugin`, which is better naming...? <- this will be in the follow-up PR
         - https://www.npmjs.com/package/appium-inspector-plugin is the current prototype place by @saikrishna321


plugins/index.mjs
```diff
@@ -0,0 +1,44 @@
+import path from 'node:path';
+import {fileURLToPath} from 'node:url';
+
+import {BasePlugin} from '@appium/base-plugin';
```
saikrishna321, line 4: We need to change this to 

`import {BasePlugin} from 'appium/plugin.js';`

cc: @mykola-mokhnach

saikrishna321 (changes requested):

plugins/index.mjs
saikrishna321: Everything LGTM.

package.json
```diff
@@ -34,16 +34,19 @@
...
     "preview:electron": "electron-vite preview",
     "pack:electron": "electron-builder build --publish never",
     "clean": "npm run clean:electron && npm run clean:browser && npm run clean:npm",
     "clean:electron": "rimraf dist/ && rimraf node_modules/.vite/ && rimraf node_modules/.vite-electron-renderer/",
     "clean:browser": "rimraf dist-browser/ && rimraf node_modules/.vite/",
+    "clean:plugin": "rimraf plugins/dist-browser",
     "clean:npm": "rimraf package-lock.json && rimraf node_modules && npm install",
     "build:docs": "appium-docs build",
     "dev:docs": "appium-docs build --serve",
+    "plugin:sync:version": "node ./scripts/sync-plugin.mjs && prettier ./plugins/package.json -w",
```
eglitise, line 49: Is the `prettier` command necessary here? We already have a script for that on line 25.

plugins/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Install the plugin using one of the following methods:
+
+```bash
+# Install from local directory
+appium plugin install --source=local /path/to/appium-inspector/plugins
+
+# Install from npm
+appium plugin install --source=npm appium-inspector-plugin
+```
```
eglitise, line 20: I think it's fine to only include the npm method here (`appium plugin install --source=npm appium-inspector-plugin`). I highly doubt anyone will bother cloning the whole repo, when it's possible to skip that step and directly install the plugin. The local installation command should still be included in the Development section since it is actually relevant there.

scripts/sync-plugin.mjs
```diff
@@ -0,0 +1,45 @@
+import path from 'node:path';
+import {fileURLToPath} from 'node:url';
+
+import fs from 'fs/promises';
```
eglitise, line 4: Would `import { promises as fs } from 'node:fs';` work here, for consistency?

eglitise:
> @appium/inspector-plugin or appium-inspector-plugin, which is better naming...?

`@appium/inspector-plugin` implies the plugin code lives under the Appium monorepo, which is not the case, so my vote goes to `appium-inspector-plugin`.

After this PR is merged, I will also update the main Inspector docs to mention this functionality.

eglitise:
Tried to test out the plugin functionality:
* `npm` version: could not load it after installation: `Cannot find package 'appium' imported from /Users/edgarseglitis/.appium/node_modules/appium-inspector-plugin/index.js`
* Local version: seems to open, but unfortunately it does not load the English translations, which means that many places use the raw translation keys, making it difficult to navigate.

package.json
KazuCocoa: That was to add a new line at the end to avoid prettier error but we could simply add a new line. Modified.

plugins/README.md
KazuCocoa: Yes, i think we can remove the local installtion after updating https://www.npmjs.com/package/appium-inspector-plugin. Currently I mainly expect to use local installation as this PR.

scripts/sync-plugin.mjs
```diff
@@ -0,0 +1,50 @@
+import path from 'node:path';
+import {fileURLToPath} from 'node:url';
+
+import fs from 'node:fs/promises';
+
+const ROOT_PKG_JSON_PATH = path.resolve(
+  path.dirname(fileURLToPath(import.meta.url)),
```
mykola-mokhnach, line 7: we might extract the root module path into a separate constant

scripts/sync-plugin.mjs
```diff
@@ -0,0 +1,50 @@
...
+ * Return JSON parsed contents from the given path.
+ * @param {string} path
+ * @returns {Promise<object>}
+ */
+async function readJsonContent(jsonPath) {
+  return await JSON.parse(await fs.readFile(jsonPath, 'utf8'));
+}
+
+async function main() {
+  const rootJsonContent = await readJsonContent(ROOT_PKG_JSON_PATH);
```
mykola-mokhnach, line 28: could also be done in parallel:
```
const [rootJsonContent, pluginJsonContent] = Promise.all([
  ROOT_PKG_JSON_PATH,
  PLUGIN_PKG_JSON_PATH,
].map(readJsonContent));
```

scripts/sync-plugin.mjs
```diff
@@ -0,0 +1,50 @@
...
+  const pluginJsonContent = await readJsonContent(PLUGIN_PKG_JSON_PATH);
+
+  // To update ever version release
+  pluginJsonContent.version = rootJsonContent.version;
```
mykola-mokhnach, line 32: the copy might be done in a for loop

package.json
```diff
@@ -34,16 +34,19 @@
     "build": "npm run build:browser && npm run build:electron",
     "build:browser": "vite build",
     "build:browser:url": "vite build --base $PUBLIC_URL",
+    "build:plugin": "vite build --base /inspector --outDir ../../plugins/dist-browser",
```
mykola-mokhnach, line 37: should it be a part of the common `build` script?

package.json
```diff
@@ -34,16 +34,19 @@
...
     "build:electron": "electron-vite build",
     "preview:browser": "npm run build:browser && vite preview",
     "preview:electron": "electron-vite preview",
     "pack:electron": "electron-builder build --publish never",
     "clean": "npm run clean:electron && npm run clean:browser && npm run clean:npm",
     "clean:electron": "rimraf dist/ && rimraf node_modules/.vite/ && rimraf node_modules/.vite-electron-renderer/",
     "clean:browser": "rimraf dist-browser/ && rimraf node_modules/.vite/",
+    "clean:plugin": "rimraf plugins/dist-browser",
```
mykola-mokhnach, line 45: same as above

package.json
```diff
@@ -34,16 +34,19 @@
...
     "clean:npm": "rimraf package-lock.json && rimraf node_modules && npm install",
     "build:docs": "appium-docs build",
     "dev:docs": "appium-docs build --serve",
+    "plugin:sync:version": "node ./scripts/sync-plugin.mjs",
```
mykola-mokhnach, line 49: same as above

plugins/README.md
```diff
@@ -0,0 +1,57 @@
...
+3. Update the plugin content with `npm run build:plugin` in `/path/to/appium-inspector`
+4. Start Appium with `appium --use-plugins=inspector --allow-cors`
+
+This plugin only needs the `appium` server as a `peerDependencies`.
+
+## Release
+
+(TODO: add this release steps in .github/workflows/package.yml later as another PR)
+
+1. Run `npm run plugin:sync:version` to sync the version with the root project.json
```
mykola-mokhnach, line 52: I would say this script should be executed automatically as soon as we bump the version of the main project via `npm version`

plugins/README.md
KazuCocoa: I remvoed via npm method for now as I haven't tested it yet. It might need more things to be adjusted.

KazuCocoa:
> Local version: seems to open, but unfortunately it does not load the English translations, which means that many places use the raw translation keys, making it difficult to navigate.

How could I check it?
e.g. on my local with the `Development` steps:
Screenshot 2025-01-02 at 12 29 31 AM

plugins/README.md
KazuCocoa: Yes. It will be a follow-up PR to treat npm publishment stuff in my current plan.

package.json
KazuCocoa: I'd like to include this in `build` after completing npm publish work as well, so yes but not yet in this PR.

plugins/index.mjs
KazuCocoa: Done with the latest.

plugins/README.md
saikrishna321: `https://www.npmjs.com/package/appium-inspector-plugin` I published this to test it out under my personnel account. I suggest once this is PR merged I can bump it up and test it again. Later will delete from npm, so we can publish from appium creds

eglitise:
> How could I check it?

When I activate the plugin and open the inspector page, Appium logs all the HTTP requests, which is where I am observing the 404 response:

```
[HTTP] --> GET /inspector {}
[HTTP] <-- GET /inspector 200 18 ms - 334
[HTTP] --> GET /inspector/assets/index-CNB9fSg_.css {}
[HTTP] --> GET /inspector/assets/index-B5YrvsL3.js {}
[HTTP] <-- GET /inspector/assets/index-CNB9fSg_.css 200 10 ms - 691539
[HTTP] <-- GET /inspector/assets/index-B5YrvsL3.js 200 2 ms - 6732856
[HTTP] --> GET /locales/en/translation.json {}
[HTTP] No route found for /locales/en/translation.json
[HTTP] <-- GET /locales/en/translation.json 404 5 ms - 211
```

KazuCocoa:
I see. Then, maybe somewhere may have hard code to `/locales` where is not modified with `--base` option in vite.
https://github.com/appium/appium-inspector/blob/d4c60996e57c3af1b6d2c3bd25a225ddb7029ca9/app/web/polyfills.js#L5-L8 ?

eglitise:
It is confusing that the locales load fine for you, though 🤔 maybe the translations are cached? If I open a new private window/incognito, the cache is cleared and I get the 404 again.

eglitise:
This seems to do the trick and works correctly in both web and plugin builds:
```js
// Adjust locales path depending on Vite base (web vs plugin)
const viteBase = import.meta.env.BASE_URL;
const vitePath = viteBase.endsWith('/') ? viteBase : `${viteBase}/`;

const localesPath =
  process.env.NODE_ENV === 'development'
    ? '/locales' // 'public' folder contents are served at '/'
    : `..${vitePath}locales`; // from 'dist-browser/assets/'
```

KazuCocoa:
Yea, my env could be the browser cache reason.

app/web/polyfills.js
```diff
@@ -2,10 +2,14 @@ import i18NextBackend from 'i18next-chained-backend';
 import HttpApi from 'i18next-http-backend';
 import LocalStorageBackend from 'i18next-localstorage-backend';
 
+// Adjust locales path depending on Vite base (web vs plugin)
+const viteBase = import.meta.env.BASE_URL;
+const vitePath = viteBase.endsWith('/') ? viteBase : `${viteBase}/`;
```
mykola-mokhnach, line 7: could also be `${_.trimEnd(viteBase, '/')}/`

scripts/sync-plugin.mjs
```diff
@@ -0,0 +1,48 @@
...
+  'homepage',
+];
+
+/**
+ * Return JSON parsed contents from the given path.
+ * @param {string} path
+ * @returns {Promise<object>}
+ */
+async function readJsonContent(jsonPath) {
+  return await JSON.parse(await fs.readFile(jsonPath, 'utf8'));
```
mykola-mokhnach, line 28: JSON.parse does not return a promise, so await there is not needed

KazuCocoa: oo, that's my bad in copy & paste somewhere

plugins/README.md
KazuCocoa: Can you invite https://www.npmjs.com/~kazucocoa to the npm package? @saikrishna321 Then, I'd like to add the package in https://www.npmjs.com/settings/appium/packages for future releases.

mykola-mokhnach (approved):

eglitise (approved):

PR #1882, opened 2025-01-05
chore(deps): bump ESlint to v9 and update other lint-related packages

eglitise:
Additional code adjustments were also necessary to tune the lint-related errors and warnings.

The one change I'm unsure of is leaving the one remaining `XPath.select` call in `utils/locator-generation.js`, because otherwise it would fail the `should keep going if xpath.select throws an exception` test. I couldn't quite figure out how to adjust Vitest to properly handle this case.


KazuCocoa (approved):

app/common/renderer/utils/locator-generation.js
```diff
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import XPath from 'xpath';
+import XPath, {select as xpathSelect} from 'xpath';
```
mykola-mokhnach, line 2: you could import it as `import * as xpath from 'xpath'`. and then perform `xpath.select`.

If it is necessary to stub these methods then the easiest method would be to create a local wrapper and stub it in unit tests instead of the third-party one, similarly to https://github.com/appium/appium-ios-simulator/pull/443/commits/28b066e4cc2ce615b93a47512630e7ea0e1e4afa

eslint.config.mjs
```diff
@@ -1,30 +1,24 @@
...
 import {includeIgnoreFile} from '@eslint/compat';
-import {FlatCompat} from '@eslint/eslintrc';
-import js from '@eslint/js';
 import reactPlugin from 'eslint-plugin-react';
 import simpleImportSortPlugin from 'eslint-plugin-simple-import-sort';
 import globals from 'globals';
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 const gitignorePath = path.resolve(__dirname, '.gitignore');
```
mykola-mokhnach, line 12: gitignore rule is already present in the appium's config

mykola-mokhnach (approved):

app/common/renderer/utils/locator-generation.js
eglitise: I did try `import * as xpath from 'xpath'` but that didn't help.
The problem here is that the test is stubbing `xpath.select` but not actually calling it directly - instead it's calling `getOptimalXPath`, which then calls `getUniqueXPath` -> `determineXpathUniqueness` -> `xpath.select`.

eslint.config.mjs
eglitise: True, but it seems like it doesn't do anything, since the `@appium/eslint-config-appium-ts` directory does not have a `.gitignore` file. It will probably need to be moved to the Appium monorepo-level `eslint.config.mjs` file.

PR #1915, opened 2025-01-27
docs: build docs for appium-inspector@2024.12

eglitise:
As with previous releases, updating the docs through a separate PR instead of CI.


mykola-mokhnach:
The "Files changed" tab freezes Safari %)

PR #1927, opened 2025-02-02
feat: add support for Appium 3 session retrieval

eglitise:
This PR adds support for retrieving and attaching to existing sessions in Appium 3, which will use the `/appium/sessions` endpoint (see https://github.com/appium/appium/issues/20880). If such sessions are found, they will also have the session ID (shown in their dropdown entry) replaced with the session creation timestamp.

Compatibility with Appium 1-2 and Selenium Grid sessions remains unchanged, although all 3 endpoints are now queried in parallel for better performance.

### DIFF
--- a/app/common/renderer/actions/Session.js
+++ b/app/common/renderer/actions/Session.js
@@ -966,27 +966,28 @@ export function getRunningSessions() {
 }
 
 async function fetchAllSessions(baseUrl, authToken) {
-  async function fetchAppiumServerSessions() {
-    const url = `${baseUrl}sessions`;
+  const appiumSessionsEndpoint = `${baseUrl}appium/sessions`; // Appium 3+
+  const oldAppiumSessionsEndpoint = `${baseUrl}sessions`; // Appium 1-2
+  const seleniumSessionsEndpoint = `${baseUrl}status`;
+
+  async function fetchSessionsFromEndpoint(url) {
     try {
       const res = await fetchSessionInformation({url, authToken});
-      return res.data.value ?? [];
+      return url === seleniumSessionsEndpoint ?
+        formatSeleniumGridSessions(res) :
+        res.data.value ?? [];
     } catch {
       return [];
     }
   }
 
-  async function fetchSeleniumGridSessions() {
-    const url = `${baseUrl}status`;
-    try {
-      const res = await fetchSessionInformation({url, authToken});
-      return formatSeleniumGridSessions(res);
-    } catch {
-      return [];
-    }
-  }
+  const [appiumSessions, oldAppiumSessions, seleniumSessions] = await Promise.all([
+    fetchSessionsFromEndpoint(appiumSessionsEndpoint),
+    fetchSessionsFromEndpoint(oldAppiumSessionsEndpoint),
+    fetchSessionsFromEndpoint(seleniumSessionsEndpoint),
+  ]);
 
-  return [...(await fetchAppiumServerSessions()), ...(await fetchSeleniumGridSessions())];
+  return [...appiumSessions, ...oldAppiumSessions, ...seleniumSessions];
 }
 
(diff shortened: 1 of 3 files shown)


KazuCocoa (approved):

app/common/renderer/utils/attaching-to-session.js
```diff
@@ -65,8 +65,15 @@ const getSessionDescription = (caps, serverType) => {
   }
 };
 
-export const getSessionInfo = (session, serverType) =>
-  `${session.id} — ${getSessionDescription(session.capabilities, serverType).assemble()}`;
+export const getSessionInfo = (session, serverType) => {
+  let identifier = session.id;
+  if ('created' in session && !_.isUndefined(session.created)) {
+    // For Appium 3+ sessions, replace session ID with timestamp
+    identifier = new Date(session.created).toJSON();
```
mykola-mokhnach, line 72: not sure if this would work if multiple sessions were created in parallel at the same time

app/common/renderer/actions/Session.js
```diff
@@ -966,27 +966,28 @@ export function getRunningSessions() {
...
-  async function fetchAppiumServerSessions() {
-    const url = `${baseUrl}sessions`;
-    try {
-      const res = await fetchSessionInformation({url, authToken});
-      return res.data.value ?? [];
-    } catch {
-      return [];
-    }
-  }
+  const appiumSessionsEndpoint = `${baseUrl}appium/sessions`; // Appium 3+
```
mykola-mokhnach, line 969: is it guaranteed that baseUrl always ends with `/`?

app/common/renderer/actions/Session.js
```diff
@@ -966,27 +966,28 @@ export function getRunningSessions() {
...
+  const oldAppiumSessionsEndpoint = `${baseUrl}sessions`; // Appium 1-2
+  const seleniumSessionsEndpoint = `${baseUrl}status`;
 
-  async function fetchSeleniumGridSessions() {
-    const url = `${baseUrl}status`;
+  async function fetchSessionsFromEndpoint(url) {
     try {
       const res = await fetchSessionInformation({url, authToken});
-      return formatSeleniumGridSessions(res);
+      return url === seleniumSessionsEndpoint
```
mykola-mokhnach, line 976: what is the expected type of this variable? is it a list?

app/common/renderer/utils/attaching-to-session.js
eglitise: The timestamp has a milisecond precision, so it's quite unlikely, but even if this does happen, the list would be rendered just fine, as there are no requirements for the identifier to be unique.

app/common/renderer/actions/Session.js
eglitise: [Yes, this is checked on line 950.](https://github.com/appium/appium-inspector/blob/main/app/common/renderer/actions/Session.js#L950)

app/common/renderer/utils/attaching-to-session.js
mykola-mokhnach: maybe it worth to make an explicit comment about the uniqueness of the identifier

app/common/renderer/actions/Session.js
eglitise: Yes, either empty or populated with session objects.

mykola-mokhnach: isn't it confusing to name it in singular then?

eglitise: Not sure I understand. The method is called `fetchSessionsFromEndpoint`, meaning there can be multiple sessions, but only one endpoint (per invocation of the method).

mykola-mokhnach: nvm I was confused by the diff. All good

PR #1942, opened 2025-02-16
chore: refactor session detail storage & retrieval

eglitise:
This is primarily a refactoring PR for the Session Information screen, but also adjusts the data format used when storing the initial session details, which is also used in the Session Information screen:
* Refactor Session Information
  * Extract the `getSession` call into a separate method for consistency
  * Avoid magic numbers when calculating session length
  * From user perspective, the server host, path, port, and session ID fields have been removed, but the session URL field has always included this information anyway
* Assemble and store the full server URL along with the server/session details
* Store headers in server details (currently only used for connecting to a TV Labs server)
* Store a new boolean var `isUsingMjpegMode` instead of implicitly converting `mjpegScreenshotUrl` to boolean


app/common/renderer/actions/Session.js
```diff
@@ -218,20 +218,20 @@ function _addVendorPrefixes(caps, dispatch, getState) {
...
  */
-export function newSession(caps, attachSessId = null) {
+export function newSession(originalCaps, attachSessId = null) {
   return async (dispatch, getState) => {
     let session = getState().session;
 
     // first add vendor prefixes to caps if requested
     if (!attachSessId && session.addVendorPrefixes) {
-      caps = _addVendorPrefixes(caps, dispatch, getState);
+      originalCaps = _addVendorPrefixes(originalCaps, dispatch, getState);
```
mykola-mokhnach, line 227: would it be possible to avoid overriding method arguments?

app/common/renderer/actions/Session.js
```diff
@@ -218,20 +218,20 @@ function _addVendorPrefixes(caps, dispatch, getState) {
...
 
-    let desiredCapabilities = caps ? getCapsObject(caps) : {};
-    let host, port, username, accessKey, https, path, token, headers;
-    desiredCapabilities = addCustomCaps(desiredCapabilities);
+    let sessionCaps = originalCaps ? getCapsObject(originalCaps) : {};
+    let host, port, username, accessKey, https, path, headers;
+    sessionCaps = addCustomCaps(sessionCaps);
 
     switch (session.serverType) {
       case SERVER_TYPES.LOCAL:
```
mykola-mokhnach, line 237: I believe in the future it makes sense to modularize this switch, so we keep all vendor-specific capability transformers (as well as other vendor-specific stuff) in separate modules, for example, `sauce.js` or `perfecto.js`, which must implement one common base interface (defined by us)

eglitise: I fully agree, but that would also make this PR a lot bigger, so I would prefer to have a separate PR for that

mykola-mokhnach (approved):

app/common/renderer/actions/Session.js
mykola-mokhnach: ofc, not in the same PR

app/common/renderer/actions/Session.js
eglitise: fixed

KazuCocoa (approved):

PR #1966, opened 2025-02-26
feat: Isolate vendor logic for session creation

mykola-mokhnach:
It's just a prototype. Any ideas for improvements are very welcome

### DIFF
--- a/app/common/renderer/actions/vendor/base.js
+++ b/app/common/renderer/actions/vendor/base.js
@@ -1,0 +1,26 @@
+export class BaseVendor {
+  constructor(server) {
+    this._server = server;
+  }
+
+  /**
+   * ! It is OK for this method to mutate sessionCaps
+   *
+   * @param {Record<string, any>} sessionCaps
+   * @returns {Promise<VendorProperties>}
+   */
+  async apply (/* sessionCaps **/) {
+    throw new Error(`The apply() method must be implemented for the ${this.constructor.name}`);
+  }
+}
+
+/**
+ * @typedef {Object} VendorProperties
+ * @property {string} [host='127.0.0.1'] Server host name
+ * @property {number} [port=4723] Server port
+ * @property {string} [username] Optional auth username for HTTP basic auth.
+ * @property {string} [accessKey] Optional auth password for HTTP basic auth.
+ * @property {boolean} [https=false] Whether to use https protocol while conecting to the server
+ * @property {string} [path='/'] Server pathname
+ * @property {Record<string, string>} [headers] Optional server headers
+ */
--- a/app/common/renderer/actions/vendor/pcluody.js
+++ b/app/common/renderer/actions/vendor/pcluody.js
@@ -1,0 +1,34 @@
+import i18n from '../../i18next';
+import { BaseVendor } from './base';
+
+export class PcloudyVendor extends BaseVendor {
+  /**
+   * @override
+   */
+  async apply(sessionCaps) {
+    const pcloudy = this._server.pcloudy;
+    const host = pcloudy.hostname;
+    const port = pcloudy.port = 443;
+    const path = pcloudy.path = '/objectspy/wd/hub';
+    const username = pcloudy.username || process.env.PCLOUDY_USERNAME;
+    const accessKey = pcloudy.accessKey || process.env.PCLOUDY_ACCESS_KEY;
+    if (!username || !accessKey) {
+      throw new Error(i18n.t('pcloudyCredentialsRequired'));
+    }
+    sessionCaps['pcloudy:options'] = {
+      ...(sessionCaps['pcloudy:options'] ?? {}),
+      source: 'appiumdesktop',
+      pCloudy_Username: username,
+      pCloudy_ApiKey: accessKey,
+    };
+    const https = pcloudy.ssl = true;
+    return {
+      path,
+      host,
+      port,
+      username,
+      accessKey,
+      https,
+    };
+  }
+}
--- a/app/common/renderer/actions/Session.js
+++ b/app/common/renderer/actions/Session.js
@@ -1,6 +1,5 @@
 import {notification} from 'antd';
 import _ from 'lodash';
-import moment from 'moment';
 import {Web2Driver} from 'web2driver';
 
 import {
@@ -19,6 +18,7 @@ import {downloadFile, parseSessionFileContents} from '../utils/file-handling';
 import {log} from '../utils/logger';
 import {addVendorPrefixes} from '../utils/other';
 import {quitSession, setSessionDetails} from './Inspector';
+import {VENDOR_MAP} from './vendor/map.js';
 
 export const NEW_SESSION_REQUESTED = 'NEW_SESSION_REQUESTED';
 export const NEW_SESSION_LOADING = 'NEW_SESSION_LOADING';
@@ -97,8 +97,6 @@ let isFirstRun = true; // we only want to auto start a session on a first run
 export const DEFAULT_SERVER_PATH = '/';
 export const DEFAULT_SERVER_HOST = '127.0.0.1';
 export const DEFAULT_SERVER_PORT = 4723;
-
-const SAUCE_OPTIONS_CAP = 'sauce:options';
 
 const JSON_TYPES = ['object', 'number', 'boolean'];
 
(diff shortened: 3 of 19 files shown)


eglitise:
Would recommend moving the `vendors` directory to `/lib` instead of keeping it in `/actions`.

app/common/renderer/actions/vendor/pcluody.js
```diff
@@ -0,0 +1,34 @@
+import i18n from '../../i18next';
```
eglitise, line 1: Typo in file name

app/common/renderer/actions/vendor/base.js
```diff
@@ -0,0 +1,26 @@
...
+    this._server = server;
+  }
+
+  /**
+   * ! It is OK for this method to mutate sessionCaps
+   *
+   * @param {Record<string, any>} sessionCaps
+   * @returns {Promise<VendorProperties>}
+   */
+  async apply(/* sessionCaps **/) {
```
eglitise, line 12: Not sure why this is async

eglitise:
Some more suggestions:
* Use a base class method for setting the common properties (host, port, path, https). I guess the latter two aren't needed for the local server type, but that one also isn't actually used and might be removed (but not necessarily in this PR).
* Use a base class method for validating the presence of required access credentials (username, accessKey, token, etc.)
  * This is a good opportunity to also parameterise the relevant translation strings for extensibility - maybe use a string just for the vendor name, then remove the vendor name from the access credential strings (e.g. Sauce Username -> Username), and finally use a common error message (something like `{{vendorName}} requires parameters: {{vendorParams}}`)
* Most vendors modify `sessionCaps`, so it makes sense to have a common method for this action too. As for the `<vendorPrefix>:options` capability name, we could set only `vendorPrefix` in the vendor class, and then use it to create the full capability name in the base class.

app/common/renderer/actions/vendor/base.js
mykola-mokhnach: to enable custom logic support for vendors where they need to invoke asynchronous functions

mykola-mokhnach:
> Would recommend moving the `vendors` directory to `/lib` instead of keeping it in `/actions`.

moved

app/common/renderer/actions/vendor/pcluody.js
mykola-mokhnach: changed

mykola-mokhnach:
> Some more suggestions:
> 
> * Use a base class method for setting the common properties (host, port, path, https). I guess the latter two aren't needed for the local server type, but that one also isn't actually used and might be removed (but not necessarily in this PR).

Could you please provide an example? I'm not quite sure how such method should look like

> * Use a base class method for validating the presence of required access credentials (username, accessKey, token, etc.)

Not sure if this should be validated as all vital properties, like host or port, have default values assigned if they are undefined

>   * This is a good opportunity to also parameterise the relevant translation strings for extensibility - maybe use a string just for the vendor name, then remove the vendor name from the access credential strings (e.g. Sauce Username -> Username), and finally use a common error message (something like `{{vendorName}} requires parameters: {{vendorParams}}`)

Not sure if it would be possible to do that. I would rather leave it to vendors to select which messages they want to show

> * Most vendors modify `sessionCaps`, so it makes sense to have a common method for this action too. As for the `<vendorPrefix>:options` capability name, we could set only `vendorPrefix` in the vendor class, and then use it to create the full capability name in the base class.

I've mentioned it in comments, that it is ok to mutate sessionCaps object passed as an argument to the `apply` method. Could you please provide an example of the above?
Also, not all vendors use the `<vendorPrefix>:options` capability. Some of them, like lambda, even support multiple prefixes. Not sure how we could unify that.

> * The URL validation used for some vendors can also be moved to base class and applied to all vendors

Not sure if this is a good idea. I intentionally kept the logic isolated, so each vendor could easily customize/update it whenever needed without affecting others. I find minor code duplication for such scenario to be lesser evil than potential possibility to break multiple vendor logics due to minor unifications.

eglitise:
@mykola-mokhnach [I've created my own branch based on these changes.](https://github.com/appium/appium-inspector/compare/main...eglitise:vendormap-example?expand=1) It is a partial example and therefore does not include files for all vendors (as well as typedefs), but it has all the base class methods I mentioned above, and should hopefully address your questions. I tried to structure the code in a way that also allows bypassing these base methods if the vendor requires additional customization.

mykola-mokhnach:
> @mykola-mokhnach [I've created my own branch based on these changes.](https://github.com/appium/appium-inspector/compare/main...eglitise:vendormap-example?expand=1) It is a partial example and therefore does not include files for all vendors (as well as typedefs), but it has all the base class methods I mentioned above, and should hopefully address your questions. I tried to structure the code in a way that also allows bypassing these base methods if the vendor requires additional customization.

Thanks. I've added some of the above suggestions. Not sure about all of them in context of different vendors. Feel free to check the updated PR

jlipps (approved):
This LGTM. It's a lot cleaner for sure. Do we need to ping vendors to make sure that they have a chance to test these changes with their clouds to ensure they still work? That is, unless you have accounts with each of them and have been able to test.

app/common/renderer/lib/vendor/base.js
```diff
@@ -0,0 +1,69 @@
...
+    throw new Error(`The apply() method must be implemented for the ${this.constructor.name}`);
+  }
+
+  /**
+   *
+   * @param {string} name
+   * @param {any} value
+   * @param {boolean} [merge=true]
+   */
+  _updateSessionCap(name, value, merge = true) {
```
eglitise, line 32: Currently the `merge` parameter is not changed for any vendor. I assume it was added for potential future use?

mykola-mokhnach: correct

app/common/renderer/actions/vendor/pcluody.js
eglitise: Unfortunately the typo is still there (`pcluody`)

eglitise:
I still maintain that more functionality could be moved to the base class like in my example. The goal should be to make it as easy as possible for new vendors to add support for their own platform, therefore the vendor-specific files should only include a minimal amount of shared code.
* The `return` call with all the vendor properties: while different vendors return different properties, I don't see why we can't just always return all of them - the unset ones will remain `undefined`, same as currently. The vendor would only need to set them.
* The assignment of `host`/`port`/`path`/`ssl` values in `session.server.<vendor>`: these variables aren't relevant for any actual functionality and are only stored in Redux. The vendor shouldn't need to worry about setting them.
* Username/access key validation: I don't believe there is a need for each vendor to define their own custom error message, as currently there are only minor variations between them, yet each requires its own translation. Like I mentioned above, we can define the vendor-specific parameter names in the translation file (and encourage vendors to add new ones if necessary), and then assemble the error message in the base class.

mykola-mokhnach:
> I still maintain that more functionality could be moved to the base class like in my example. The goal should be to make it as easy as possible for new vendors to add support for their own platform, therefore the vendor-specific files should only include a minimal amount of shared code.
> 
> * The `return` call with all the vendor properties: while different vendors return different properties, I don't see why we can't just always return all of them - the unset ones will remain `undefined`, same as currently. The vendor would only need to set them.
> * The assignment of `host`/`port`/`path`/`ssl` values in `session.server.<vendor>`: these variables aren't relevant for any actual functionality and are only stored in Redux. The vendor shouldn't need to worry about setting them.
> * Username/access key validation: I don't believe there is a need for each vendor to define their own custom error message, as currently there are only minor variations between them, yet each requires its own translation. Like I mentioned above, we can define the vendor-specific parameter names in the translation file (and encourage vendors to add new ones if necessary), and then assemble the error message in the base class.

What if we merge this PR as is and then you could add further improvements on top of it, or you'd rather prefer a separate change?

eglitise:
> What if we merge this PR as is and then you could add further improvements on top of it, or you'd rather prefer a separate change?

Sure, I can make another PR.

app/common/renderer/lib/vendor/base.js
eglitise: Actually, looking at the current vendor implementations again, it should already be applied for some vendors - specifically, Perfecto, Bitbar, and pCloudy, as they directly overwrite their `options` capability object, instead of individual capabilities inside that object.

mykola-mokhnach: I saw that, although I assume they still would prefer merge in the end. It's safe and is only used if users actually provide some options in advance

eglitise: There is one case where this may not safe: if the user (intentionally or not) provides their own options, but the vendor does not have handling for unexpected parameters and throws an error. It's unlikely, but still worth handling.

mykola-mokhnach: Set merge to false for the three above vendors to make sure we don't change the original logic

eglitise (approved):

PR #1973, opened 2025-03-04
chore: refactor lib directory

eglitise:
This is a relatively small refactor of the `lib` directory files, in preparation for integrating web2driver.

* Move `webview-helpers.js` to the `utils` folder
* Add a new test for webview parsing functionality
* Move `appium-client.js` to `appium/inspector-client.js` & rename its class
* Group all client frameworks using a `map.js` file, similarly to #1966
* Add a constant for all framework names (values intentionally match the keys previously used in `index.js`, to ensure backwards compatibility)
* For each framework, set its visible name & highlight.js language as explicit static properties
* Rename various client framework-related variables for consistency & clarity


app/common/renderer/components/Inspector/Recorder.jsx
```diff
@@ -1,30 +1,32 @@
...
-  const {showBoilerplate, recordedActions, actionFramework, t} = props;
+  const {showBoilerplate, recordedActions, clientFramework, t} = props;
 
   const code = (raw = true) => {
     const {serverDetails, sessionCaps} = props;
     const {serverUrl, serverUrlParts} = serverDetails;
 
-    let framework = new frameworks[actionFramework](serverUrl, serverUrlParts, sessionCaps);
+    const ClientFrameworkClass = CLIENT_FRAMEWORK_MAP[clientFramework];
+    let framework = new ClientFrameworkClass(serverUrl, serverUrlParts, sessionCaps);
```
mykola-mokhnach, line 19: const?

app/common/renderer/lib/appium/inspector-driver.js
```diff
@@ -16,7 +16,16 @@ const IOS_TOP_CONTROLS_SELECTOR =
...
 let _instance = null;
 
-export default class AppiumClient {
+/**
+ * Wrapper class providing access to the currently active Appium driver methods,
+ * with additional Inspector-specific handling
+ */
+export default class InspectorDriver {
+  static instance(driver) {
+    _instance ??= new this(driver);
```
mykola-mokhnach, line 25: I prefer to use the [_.memoize](https://github.com/appium/appium/pull/21075/files#diff-4cbd581b1c401dcc27804541efdeb2d986402d1dc879f06b8c18f443adbbbcdaR46) wrapper for this purpose if we need lazy initialization.

mykola-mokhnach: I also assume such implementation might have a potential bug, since we probably only expect the same instance to be returned for the same driver argument value, but a different one if `driver` differs

jlipps (approved):
agree w/ @mykola-mokhnach's comment about memoize, otherwise LGTM

KazuCocoa (approved):
lg overall

app/common/renderer/lib/client-frameworks/dotnet-nunit.js
```diff
@@ -1,11 +1,10 @@
...
-import Framework from './framework';
+import CommonClientFramework from './common.js';
 
-class DotNetNUnitFramework extends Framework {
-  get language() {
-    return 'csharp';
-  }
+export default class DotNetNUnitFramework extends CommonClientFramework {
+  static readableName = '.NET - NUnit';
+  static highlightLang = 'csharp';
```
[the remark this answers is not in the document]
eglitise: The `readableName` values are used to create the framework dropdown options ([here](https://github.com/appium/appium-inspector/pull/1973/files#diff-7c3b74aafe25f5dd52bafb5e86971b5265873a09e3d7921d846e60956736677bR55) and [here](https://github.com/appium/appium-inspector/pull/1973/files#diff-2b86cb011d499114b76071ccf69f5b6bb42d69dc2b7a79fb9fc16059088cedf9R37)), and instantiating each class just to retrieve this value sounds a bit wasteful. Would it be better to define them as consts outside the class? Alternatively, the map for these values can be defined as a separate constant.
This isn't an issue for `highlightLang` because that's only requested for one framework at a time, and the required class instance already exists at that point - but it would also be reasonable to store these values similarly to `readableName`.

app/common/renderer/lib/appium/inspector-driver.js
eglitise: Looked a bit into `_.memoize` and I'm not quite sure how to use it in this case. It expects a function, but we are trying to memoize the `new` operator along with the class.
The current approach with nullish coalescing assignment does work because it will correctly create a new instance only for the first `InspectorDriver.instance` call, but a precondition for a having different `driver` is the need to quit the current `driver` session - [which is when `_instance` is cleared](https://github.com/appium/appium-inspector/pull/1973/files#diff-5b14b4189625b97bb0c88ca73cafe8d5aa17a74b9044f31fb4939f359e8b7e47R54).

PR #1990, opened 2025-03-15
chore: add '|| ^3.0.0-beta.0' to plugin

KazuCocoa:
to avoid:
```
✖ Checking if 'appium-inspector-plugin' is compatible
Error: ✖ 'inspector' cannot be installed because the server version it requires (^2.0.0) does not meet the currently installed one (3.0.0-beta.0). Please install a compatible server version first.
```
in the install command. So far, we don't have breaking changes related to this plugin.

### DIFF
--- a/plugins/package.json
+++ b/plugins/package.json
@@ -27,7 +27,7 @@
     "./package.json": "./package.json"
   },
   "peerDependencies": {
-    "appium": "^2.0.0"
+    "appium": "^2.0.0 || ^3.0.0-beta.0"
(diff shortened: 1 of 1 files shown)


eglitise (approved):

eglitise:
Should the `@appium/base-plugin` dependency also be bumped to allow 3.0.0 or later?

KazuCocoa:
I expect that bumping the version will be later with other plugins/drivers do that

PR #1992, opened 2025-03-15
chore: refactor vendormap for clarity

eglitise:
Follow-up to #1966. The main change is 2 new methods for clarity: `_checkInputPropertyPresence()` and `_saveProperties()`.

* `_checkInputPropertyPresence()` provides a common interface for validating the presence of values in all input fields exposed to the user (that do not have a default value).
  * It allows the vendor to specify both the names and values of these properties, ensuring vendor-specific terminology in the error message, should one or more properties be absent
  * The vendor name is also included in the error message. It is specified in their vendor file, and is intentionally not translated
  * The translation file has also been updated - the full vendor-specific error messages have been removed, and instead only the vendor-specific properties have been added, along with a common error message template. This reduces the overall translation efforts, and in the future will also allow removing translation keys for the property labels (e.g. `LambdaTest Username` can be replaced with `Username` - the vendor name is already obvious from the selected tab)
  * Some fields that were missing this validation now have it (e.g. pCloudy hostname)
* `_saveProperties()` provides a common interface for saving all properties as class instance variables, allowing to use a common return call and removing the need to have a return call for each vendor


app/common/renderer/lib/vendor/base.js
```diff
@@ -51,19 +72,61 @@ export class BaseVendor {
...
     return webdriverUrl;
   }
+
+  /**
+   * Locally save all class properties required to access the vendor server
+   *
+   * @param {Object} vendor
+   * @param {VendorProperties} vendorProperties
+   */
+  _saveProperties(vendor, {host, path, port, https, username, accessKey, headers}) {
```
KazuCocoa, line 86: It's not the main issue (and not a very strong opinion), but we could probably remove the private prefix `_` for functions that will be called by vendor modules. I guess they might be confused if they can use `_` ones. (we already have multiple vendor implementations, so this should not be confusing  today though)

KazuCocoa (approved):

app/common/renderer/lib/vendor/base.js
```diff
@@ -11,59 +11,122 @@ export class BaseVendor {
...
   }
 
   /**
+   * Validate the presence of one or more properties that the user can enter in the Inspector GUI
    *
-   * @param {string} name
-   * @param {any} value
-   * @param {boolean} [merge=true]
+   * @param {string} vendorName
+   * @param {List<InputProperty>} propertyList
```
mykola-mokhnach, line 45: InputProperty[]

app/common/renderer/lib/vendor/base.js
```diff
@@ -11,59 +11,122 @@ export class BaseVendor {
...
+ * @property {string} [username] Optional username for authenticating to vendor service
+ * @property {string} [accessKey] Optional password/access key for authenticating to vendor service
+ * @property {boolean} [https=false] Whether to use https protocol while connecting to the server
  * @property {string} [path='/'] Server pathname
  * @property {Record<string, string>} [headers] Optional server headers
  */
+
+/**
+ * @typedef {Object} InputProperty
+ * @property {string} [name] Property name (used in error messages if no value is provided)
```
mykola-mokhnach, line 130: I assume name is required and cannot be undefined (e.g. square brackets are not needed)

app/common/renderer/lib/vendor/base.js
```diff
@@ -11,59 +11,122 @@ export class BaseVendor {
...
+ * @property {string} [val] Property value
```
mykola-mokhnach, line 131: same as above

app/common/renderer/lib/vendor/base.js
```diff
@@ -11,59 +11,122 @@ export class BaseVendor {
...
-    if (merge && _.isPlainObject(previousValue) && _.isPlainObject(value)) {
-      this._sessionCaps[name] = {
-        ...previousValue,
-        ...value,
-      };
-    } else {
-      this._sessionCaps[name] = value;
+  checkInputPropertyPresence(vendorName, propertyList) {
+    const missingProps = [];
+    for (const prop of propertyList) {
```
mykola-mokhnach, line 49: could be `for (const [name, val] of _.toPairs(propertyList))`. _.toPairs is equivalent to Object.entries

app/common/renderer/lib/vendor/bitbar.js
```diff
@@ -4,30 +4,26 @@ export class BitbarVendor extends BaseVendor {
   /**
    * @override
    */
-  async apply() {
+  async setProperties() {
```
mykola-mokhnach, line 7: I find it confusing that there are two methods saveProperties and setProperties. Perhaps they could have more descriptive names

app/common/renderer/lib/vendor/base.js
```diff
@@ -11,59 +11,122 @@ export class BaseVendor {
...
   }
 
   /**
+   * Check validity of the WebDriver URL, which may be required by the vendor
    *
    * @param {string} url
    * @returns {URL}
    */
-  _validateUrl(url) {
+  validateUrl(url) {
```
mykola-mokhnach, line 70: not sure why it is needed to make this method public
I usually try to make all class method private unless they must be used externally. the reason for that is that all private methods could be safely modified in the future whenever changing of public ones might be breaking and should be confirmed across all class consumers

eglitise: I guess this can go either way, I can understand Kazu's viewpoint that new vendors wanting to add their own integration may be confused about having to use private methods. But it is also true that the methods are only expected to be used in the vendor classes.

app/common/renderer/lib/vendor/base.js
eglitise: `propertyList` is a list of objects, so `toPairs` returns `[[0, {name: <name>, val: <val>}], [1, {name: <name>, val: <val>}], ...]`

app/common/renderer/lib/vendor/base.js
eglitise: I've changed them back to private for now

app/common/renderer/lib/vendor/base.js
mykola-mokhnach: sorry for the confusion I mean `for (const [name, val] of propertyList)`. I somehow assumed propertyList is an object

app/common/renderer/lib/vendor/base.js
mykola-mokhnach: if we had typescript then protected modified would be the best solution. unfortunately vanilla JS does not make any diff between private and protected if a method has an underscore prefix

app/common/renderer/lib/vendor/base.js
eglitise: Changed to `for (const {name, val} of propertyList)` which seems to work

mykola-mokhnach (approved):

PR #2018, opened 2025-03-29
feat: adjust Commands tab command list

eglitise:
This is a sizeable adjustment to the list of commands shown in the Commands tab, as required by #2016. Many commands have been removed due to their removal in Appium 3 (though all of them have execute method replacements), and some missing commands have been added. Some command categories have also been merged for simplicity. Finally, the recorder code for all client frameworks has also been adjusted accordingly.

List of changes:
* Remove 27 Commands tab commands that will be removed in Appium 3. Each is replaced with the following `executeMethod` command:
  * `background` -> `mobile: backgroundApp` (XCUI/UiA2/Espresso)
  * `fingerPrint` -> `mobile: fingerprint` (UiA2/Espresso)
  * `getCurrentActivity` -> `mobile: getCurrentActivity` (UiA2/Espresso)
  * `getCurrentPackage` -> `mobile: getCurrentPackage` (UiA2/Espresso)
  * `getClipboard` -> `mobile: getClipboard` (XCUI/UiA2/Espresso)
  * `getPerformanceData` -> `mobile: getPerformanceData` (UiA2/Espresso)
  * `getPerformanceDataTypes` -> `mobile: getPerformanceDataTypes` (UiA2/Espresso)
  * `getStrings` -> `mobile: getAppStrings` (XCUI/UiA2/Espresso)
  * `gsmCall` -> `mobile: gsmCall` (UiA2/Espresso)
  * `gsmSignal` -> `mobile: gsmSignal` (UiA2/Espresso)
  * `gsmVoice` -> `mobile: gsmVoice` (UiA2/Espresso)
  * `isLocked` -> `mobile: isLocked` (XCUI/UiA2/Espresso)
  * `lock` -> `mobile: lock` (XCUI/UiA2/Espresso)
  * `longPressKeyCode` -> `mobile: pressKey` (UiA2/Espresso)
  * `openNotifications` -> `mobile: openNotifications` (UiA2) / `mobile: statusBar` (UiA2/Espresso)
  * `pressKeyCode` -> `mobile: pressKey` (UiA2/Espresso)
  * `sendSms` -> `mobile: sendSms` (UiA2/Espresso)
  * `setClipboard` -> `mobile: setClipboard` (XCUI/UiA2/Espresso)
  * `shake` -> `mobile: shake` (XCUI)
  * `startActivity` -> `mobile: startActivity` (UiA2/Espresso)
  * `toggleAirplaneMode` -> `mobile: setConnectivity` (UiA2/Espresso)
[...]

PR #2020, opened 2025-03-30
chore: update ruby commands

KazuCocoa:
Helps https://github.com/appium/appium-inspector/pull/2018 a bit

### DIFF
--- a/app/common/renderer/lib/client-frameworks/ruby.js
+++ b/app/common/renderer/lib/client-frameworks/ruby.js
@@ -165,15 +165,15 @@ driver.quit`;
   // Device Interaction
 
   codeFor_getWindowRect() {
-    return `window_rect = driver.get_window_rect`;
+    return `window_rect = driver.window_rect`;
   }
 
   codeFor_takeScreenshot() {
-    return `screenshot = driver.take_screenshot`;
+    return `screenshot = driver.screenshot_as :base64`;
   }
 
   codeFor_isKeyboardShown() {
-    return `is_keyboard_shown = ${this.codeFor_executeScriptNoArgs('mobile: isKeyboardShown')}`;
+    return `is_keyboard_shown = driver.keyboard_shown?`;
   }
 
   codeFor_getOrientation() {
@@ -243,7 +243,7 @@ driver.quit`;
   }
 
   codeFor_getUrl() {
-    return `current_url = driver.get_current_url`;
+    return `current_url = driver.current_url`;
   }
 
   codeFor_back() {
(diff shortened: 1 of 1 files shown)


app/common/renderer/lib/client-frameworks/ruby.js
```diff
@@ -243,42 +243,42 @@ driver.quit`;
...
   }
 
   codeFor_getWindowHandles() {
-    return `window_handles = driver.get_window_handles`;
+    return `window_handles = driver.window_handles`;
   }
 
   codeFor_createWindow(varNameIgnore, varIndexIgnore, type) {
-    return `new_window = driver.new_window '${type}'`;
+    return `new_window = driver.switch_to.window :${type}`;
```
eglitise, line 282: If this is the Selenium implementation, I don't think it will actually return the new window object. That's the reason why I've set this method as not supported for most languages.

eglitise:
I was basing the commands off [this file](https://github.com/appium/ruby_lib_core/blob/master/lib/appium_lib_core/common/command.rb), but I guess the client actually uses the Selenium implementations?

app/common/renderer/lib/client-frameworks/ruby.js
KazuCocoa: I see. I have updated proper new_window and switch to a window 
 https://github.com/appium/appium-inspector/pull/2020/commits/e1af9535c6028db53027ba71735a9435fdad4d66

Current example simply raises no method error.

KazuCocoa:
The file is available endpoints list, but they are not method names. selenium's ones and https://github.com/appium/ruby_lib_core/blob/master/lib/appium_lib_core/common/base/driver.rb are methods usually called as `driver.something`

app/common/renderer/lib/client-frameworks/ruby.js
eglitise: According to the [New Window spec](https://w3c.github.io/webdriver/#new-window), it should return an object, but I'm not fully sure if Selenium actually does this. They do reference this [in one location](https://github.com/SeleniumHQ/selenium/blob/trunk/rb/lib/selenium/webdriver/remote/bridge.rb#L188), but I couldn't find any tests that check the actual return value.

KazuCocoa: Here is the `driver.switch_to.new_window` call in Ruby https://github.com/SeleniumHQ/selenium/blob/06fe6815171ae4e388e98a17b2477e95b8682d9b/rb/lib/selenium/webdriver/common/target_locator.rb#L54-L73
Your referenced method is called in https://github.com/SeleniumHQ/selenium/blob/06fe6815171ae4e388e98a17b2477e95b8682d9b/rb/lib/selenium/webdriver/common/target_locator.rb#L58

It looks like selenium clients set the focus window to the newly opened tab/window internally

KazuCocoa: https://github.com/appium/appium-inspector/pull/2020/commits/e36baf779e683f3e885c343e20807010cc218a6d or simply return not supported as no primitive method to call only the `/window/new` endpoint as this code generation? I'm ok for either.

app/common/renderer/lib/client-frameworks/ruby.js
```diff
@@ -243,42 +243,42 @@ driver.quit`;
...
 
   codeFor_getWindowHandles() {
-    return `window_handles = driver.get_window_handles`;
+    return `window_handles = driver.window_handles`;
   }
 
-  codeFor_createWindow(varNameIgnore, varIndexIgnore, type) {
-    return `new_window = driver.new_window '${type}'`;
+  codeFor_createWindow() {
+    return `# Not supported: 'driver.switch_to.new_window' opens a new window but it also switches to the window.`;
```
eglitise, line 282: Let's align this with the other frameworks and just use `Not supported: createWindow`

KazuCocoa: https://github.com/appium/appium-inspector/pull/2020/commits/cf48604d67422d8061c3ae9816221dfcdcd02c57

PR #2049, opened 2025-04-17
feat: implement tab close functionality with i18n support

losingle:
- Add right-click close functionality for tabs
- Add 'Close tab' translations for multiple languages including:
  - English
  - Chinese (Simplified & Traditional)
  - Japanese
  - Korean
  - French
  - Spanish
  - German
  - Russian


app/common/public/locales/en/translation.json
```diff
@@ -291,5 +291,6 @@
   "AccessToken": "AccessToken",
   "Token": "Token",
   "Key": "Key",
-  "Secret": "Secret"
+  "Secret": "Secret",
+  "Close tab": "Close tab"
```
mykola-mokhnach, line 295: It is enough to only add English resources. localized resources are synchronized automatically to crowdin as soon as the PR is merged and should not be changed manually.

losingle: I only kept English resources

app/common/renderer/components/Session/Session.jsx
```diff
@@ -44,8 +45,9 @@ const Session = (props) => {
   } = props;
 
   const navigate = useNavigate();
+  const [activeTab, setActiveTab] = useState(SESSION_BUILDER_TABS.CAPS_BUILDER);
```
eglitise, line 48: I don't quite see the reason in adding this, along with `handleTabChange` and `handleSwitchTabs`. Switching session builder tabs is also not related to the server type tab closing functionality.

app/common/renderer/components/Session/Session.jsx
```diff
@@ -24,11 +24,12 @@ import SessionStyles from './Session.module.css';
 
 const Session = (props) => {
   const {
-    tabKey,
     switchTabs,
     serverType,
+    setServerType,
```
eglitise, line 29: No such function exists. You probably wanted to use `changeServerType`

app/common/renderer/components/Session/Session.jsx
```diff
@@ -56,20 +58,54 @@ const Session = (props) => {
...
       navigate('/inspector', {replace: true});
     }
   };
 
+  const getContextMenu = (tabKey) => ({
+    items: [
+      {
+        key: 'closeTab',
+        label: t('Close tab'),
+        onClick: () => handleCloseTab(tabKey),
```
eglitise, line 82: Unfortunately this action will also trigger `handleSelectServerTab`, which will try to select the same tab that was just removed

eglitise (commented):
First of all, thank you for the PR!
It looks like you are mixing up two different sets of tabs - one is used for switching between server types (Appium server & any cloud providers), and the other is used for session builder tabs (session builder, saved sets, attach to session). I would suggest only including changes for the former in this PR.
I do think the added functionality is useful, but a) it changes the regular right-click behavior that applies elsewhere in the app, and b) it may be less hacky to implement [antd's own tab close button functionality](https://4x.ant.design/components/tabs/#components-tabs-demo-editable-card).

Mtomik86 (approved):

eglitise:
I am closing this as abandoned, but feel free to resubmit an updated version of this feature.
Arguably the tab-based UI isn't particularly great as a whole, since I don't expect most users to be regularly switching between tabs, so it'd be more beneficial to do a larger redesign.

PR #2103, opened 2025-05-28
fix: attach to session tab not making requests

regan-karlewicz:
## Bug

When navigating to the `Attach to Session...` tab in the Sessions view, I had noticed that the requests for discovering sessions were not being made when any cloud vendor was set.

## Solution

- Use new vendor map when getting server details for `getRunningSessions`
- Use `headers` in the request if they exist on the vendor
- Call `showError` if session discovery query cannot be made to give user feedback

### DIFF
--- a/app/common/renderer/actions/Session.js
+++ b/app/common/renderer/actions/Session.js
@@ -654,7 +654,6 @@ function retreiveVendorProperties({server, serverType, sessionCaps}) {
     log.info(`Using ${VendorClass.name}`);
     try {
       const vendor = new VendorClass(server, sessionCaps);
-      // ({host, port, username, accessKey, https, path, headers} = await vendor.apply());
       return vendor.apply();
     } catch (e) {
       showError(e);
--- a/app/common/renderer/actions/Session.js
+++ b/app/common/renderer/actions/Session.js
@@ -709,8 +709,9 @@ export function getRunningSessions() {
       return;
     }
 
+    const protocol = https ? 'https' : 'http';
     const adjPath = path.endsWith('/') ? path : `${path}/`;
-    const baseUrl = `http${https ? 's' : ''}://${host}:${port}${adjPath}`;
+    const baseUrl = `${protocol}://${host}:${port}${adjPath}`;
     const sessions = await fetchAllSessions(baseUrl, headers);
     dispatch({type: GET_SESSIONS_DONE, sessions});
 
--- a/app/common/renderer/actions/Session.js
+++ b/app/common/renderer/actions/Session.js
@@ -226,7 +226,10 @@ export function newSession(originalCaps, attachSessId = null) {
     let sessionCaps = prefixedCaps ? getCapsObject(prefixedCaps) : {};
     sessionCaps = addCustomCaps(sessionCaps);
 
-    let {host, port, username, accessKey, https, path, headers} = await retreiveVendorProperties({...session, sessionCaps});
+    let {host, port, username, accessKey, https, path, headers} = await retreiveVendorProperties({
+      ...session,
+      sessionCaps,
+    });
 
     // if the server path is '' (or any other kind of falsy) set it to default
     path = path || DEFAULT_SERVER_PATH;
@@ -658,9 +661,7 @@ function retreiveVendorProperties({server, serverType, sessionCaps}) {
       return false;
     }
   } else {
-    log.info(
-      `No vendor mapping is defined for the server type '${serverType}'. Using defaults`,
-    );
+    log.info(`No vendor mapping is defined for the server type '${serverType}'. Using defaults`);
 
     return {};
   }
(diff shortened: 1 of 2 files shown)


app/common/renderer/actions/Session.js
```diff
@@ -656,6 +639,33 @@ export function saveSessionAsFile() {
   };
 }
 
+/**
+ * @returns {Promise<VendorProperties>}
+ */
+function retreiveVendorProperties({server, serverType, sessionCaps}) {
```
eglitise, line 645: Typo in method name, also missing the `async` keyword

app/common/renderer/actions/Session.js
```diff
@@ -225,28 +225,11 @@ export function newSession(originalCaps, attachSessId = null) {
...
-        showError(e);
-        return false;
-      }
-    } else {
-      log.info(
-        `No vendor mapping is defined for the server type '${session.serverType}'. Using defaults`,
-      );
-    }
+
+    let {host, port, username, accessKey, https, path, headers} = await retreiveVendorProperties({
```
eglitise, line 229: This call does not distinguish the possible return values `false` and `{}`. In the former case it should return from the `newSession` function to match the existing behavior.

app/common/renderer/actions/Session.js
```diff
@@ -656,6 +639,33 @@ export function saveSessionAsFile() {
...
+  // - Implement a new class inherited from VendorBase in app/common/renderer/lib/vendor/<vendor_name>.js
+  // - Add the newly created class to the VENDOR_MAP defined in app/common/renderer/lib/vendor/map.js
+  //
+  const VendorClass = VENDOR_MAP[serverType];
+
+  if (VendorClass) {
+    log.info(`Using ${VendorClass.name}`);
+    try {
+      const vendor = new VendorClass(server, sessionCaps);
+      return vendor.apply();
```
eglitise, line 657: Missing `await`

app/common/renderer/actions/Session.js
```diff
@@ -664,19 +674,32 @@ export function getRunningSessions() {
...
+      server,
+      serverType,
+      sessionCaps: {},
+    });
+
+    if (username && accessKey) {
+      const authToken = btoa(`${username}:${accessKey}`);
+
+      headers = {
+        ...(headers || {}),
```
eglitise, line 687: This can probably be simplified to just `...headers,`

app/common/renderer/actions/Session.js
```diff
@@ -664,19 +674,32 @@ export function getRunningSessions() {
...
-      hostname = hostname || DEFAULT_SERVER_HOST;
+      host = host || DEFAULT_SERVER_HOST;
       port = port || DEFAULT_SERVER_PORT;
       path = path || DEFAULT_SERVER_PATH;
     }
 
     // no need to get sessions if we don't have complete server info
-    if (!hostname || !port || !path) {
+    if (!host || !port || !path) {
+      showError(new Error('Missing server info, could not query sessions'), {secs: 5});
```
eglitise, line 702: User-facing messages should preferably be using translation keys. The actual string should then be added in [the English translation file.](https://github.com/appium/appium-inspector/blob/main/app/common/public/locales/en/translation.json)
Also, 5 seconds is the default value and can therefore be omitted here.

app/common/renderer/actions/Session.js
```diff
@@ -225,28 +225,11 @@ export function newSession(originalCaps, attachSessId = null) {
...
-        return false;
-      }
-    } else {
-      log.info(
-        `No vendor mapping is defined for the server type '${session.serverType}'. Using defaults`,
-      );
-    }
+
+    let {host, port, username, accessKey, https, path, headers} = await retreiveVendorProperties({
+      ...session,
```
eglitise, line 230: I would rather explicitly pass `session.server` and `session.serverType`, since `session` has a lot of other properties that aren't needed in the target method. This also makes it consistent with the method call in `getRunningSessions`

app/common/renderer/actions/Session.js
```diff
@@ -656,6 +639,33 @@ export function saveSessionAsFile() {
   };
 }
 
+/**
+ * @returns {Promise<VendorProperties>}
```
eglitise, line 643: `false` and `{}` are also valid return values

eglitise (changes requested):
Thanks for spotting and fixing this 👍
Left a couple minor comments

regan-karlewicz:
@eglitise Thanks for the feedback, I've addressed your comments!

app/common/renderer/lib/vendor/tvlabs.js
regan-karlewicz, line 1: Allows overrides on vendor config when using `initFromQueryString` for the TV Labs vendor. @eglitise please let me know if you'd like me to move this to a separate PR.

app/common/renderer/actions/Session.js
```diff
@@ -664,19 +685,32 @@ export function getRunningSessions() {
     const avoidServerTypes = ['sauce'];
     const state = getState().session;
     const {server, serverType, attachSessId} = state;
-    let {hostname, port, path, ssl, username, accessKey} = server[serverType];
-    const authToken = username && accessKey ? btoa(`${username}:${accessKey}`) : null;
+    let {path, host, port, username, accessKey, https, headers} = await retrieveVendorProperties({
```
mykola-mokhnach, line 688: this will throw an error if the method returns false

app/common/renderer/lib/vendor/tvlabs.js
eglitise: I think it's fine to keep it

PR #2123, opened 2025-06-12
refactor: upgrade antd to v5

wilhus:
Notable changes according to https://ant.design/docs/react/migration-v5:
- Default border radius of base components has increased.
- Less files are removed and less variables are no longer exported.
- IE browser is no longer supported.
- PageHeader component was moved into `@ant-design/pro-components`.

Summary of changes in this PR:
- Migrated styling in `main.less` to `main.css` since less variables are no longer supported by antd. Some styling seemed unnecessary but kept it for now.
- Using `ConfigProvider` for configuration support.
- Using `App` as an application wrapper.
- Using `Layout` for overall layout of the app.

See screenshots below to compare the old version with the new one.

Also see examples below of warning messages that should be handled.

Please try it out and let me know if something is not working as intended!

After doing this upgrade we should:
- Handle new warnings.
- Refactor notification static method to hooks usage.
- Add dark mode.

### Session builder

| Old | New |
| - | - |
| session-builder-old | session-builder-new |

### Session inspector

| Old | New |
| - | - |
| session-inspector-old | session-inspector-new |

## Warnings
These warnings will need to be handled but are not critical right now.

- `[antd: notification] Static function can not consume context like dynamic theme. Please use App component instead.`
- `[antd: Button.Group] Button.Group is deprecated. Please use Space.Compact instead.`
- `[rc-collapse] children will be removed in next major version. Please use items instead.`


eglitise:
Thank you so much! I had looked into `antd v5` multiple times, but had assumed that all of the CSS/Less changes would have to be converted to the Component Token interface, which would require tons of effort. Glad to see it's a lot simpler than that!

After some brief testing (dev & preview, Electron & browser), everything still seems to work, and the layout appears to be the same as well! I did notice that icons and toggles are a bit smaller than in `antd v4`, but that's something that can easily be adjusted later.

Some additional notes:
* CI failures seem to be related to `package-lock` changes. Maybe it's your npm/Node version? I'm running Node `22.12` + `npm 10.9` on Windows, and a simple `npm i` did _not_ result in the removal of optional dependencies, unlike what seems to have happened here.
* Could you please also remove `antd` from the Renovate blacklist, in `renovate.json`?

app/common/renderer/Root.jsx
```diff
@@ -14,17 +15,35 @@ ipcRenderer.on('appium-language-changed', (event, message) => {
...
-        <Routes>
-          <Route path="/" element={<SessionPage />} />
-          <Route path="/session" element={<SessionPage />} />
-          <Route path="/inspector" element={<InspectorPage />} />
-        </Routes>
-      </Suspense>
-    </MemoryRouter>
+    <ConfigProvider theme={getTheme()}>
+      <App>
+        <Layout>
```
eglitise, line 34: I assume these two, plus their CSS adjustments in `main.css`, are required to ensure that the layout does not change from the current one?

app/common/renderer/assets/stylesheets/main.css
```diff
@@ -0,0 +1,82 @@
+@import 'antd/dist/reset.css';
+@import 'highlight.js/styles/intellij-light.css';
+
+// Global styles
```
eglitise: I'm getting a warning on this line, maybe `/* Global styles */` could work?

eglitise: Arguably the comment was needed in the Less version to differentiate the changes from the antd variables, but since they have been removed in this CSS file, this comment may as well also be removed

app/common/renderer/components/Inspector/GestureEditor.jsx
```diff
@@ -9,14 +9,14 @@ import {
   RightCircleOutlined,
   UpCircleOutlined,
 } from '@ant-design/icons';
+import {PageHeader} from '@ant-design/pro-components';
```
eglitise, line 12: I was also looking into converting the current use of PageHeader into antd standard components, but it was not trivial, so I'm okay with adding this import for now

app/common/renderer/Root.jsx
wilhus: Yes! The adjustments in `main.css` make sure that they use the full height. The components themselves are needed for a few reasons:
- App (https://ant.design/components/app) provides context that is needed for the notification hooks usage among others.
- Layout (https://ant.design/components/layout) applies proper background color based on the theme.

app/common/renderer/assets/stylesheets/main.css
wilhus: Removed now!

app/common/renderer/components/Inspector/GestureEditor.jsx
wilhus: 👍

app/common/renderer/assets/stylesheets/main.css
```diff
@@ -0,0 +1,81 @@
...
+.list-group-item.active {
+  background-color: #662d91 !important;
+}
+
+.ant-spin-nested-loading,
+.ant-spin-container {
+  height: 100%;
+}
+
+.ant-notification {
```
eglitise: Since the `width` property no longer affects this class, the notification is no longer centered. I think this can be fixed by setting `margin-right` to `33%`. The `-webkit-app-region` property also doesn't seem to do anything and can be removed.

mykola-mokhnach (approved):

app/common/renderer/assets/stylesheets/main.css
wilhus: Setting the `margin-right` to `33%` would work but it won't be truly centered since the distance would be based off the right side only. I guess it can be fixed by trying different margins or finding another way to make it centered. If it's of importance that the notification is centered I can try out some things, otherwise it could remain as the default of right side. Wdyt?

I can remove the `-webkit-app-region` property. There also seems to be a few other properties that don't do a lot or anything at all. My intention was to make as few changes as possible but maybe I should remove them since it's being refactored now anyways?

eglitise: Good point, then I think it should be fine to move it to the default location on the right side. In that case this entire `.ant-notification` block can be removed, but any other properties can be handled later.

wilhus: Removed the notification block

eglitise (approved):
Thank you!

PR #2127, opened 2025-06-15
refactor: button group to space compact

wilhus:
Refactor according to console warning since upgrading to antd v5 (https://github.com/appium/appium-inspector/pull/2123): 

`[antd: Button.Group] Button.Group is deprecated. Please use Space.Compact instead.`

No changes to UI as far as I could tell.


eglitise (approved):
👍

PR #2130, opened 2025-06-15
refactor: notification to hooks usage

wilhus:
Refactor according to console warning since upgrading to antd v5 (https://github.com/appium/appium-inspector/pull/2123): 

`[antd: notification] Static function can not consume context like dynamic theme. Please use App component instead.`

Now using `notification` from `App.useApp()`. It's recommended by antd to use hooks instead of the static notification method (https://ant.design/components/notification). Notifications are triggered by dispatching an event which are picked up by the `Notification` component listener.

### DIFF
--- a/app/common/renderer/Root.jsx
+++ b/app/common/renderer/Root.jsx
@@ -3,6 +3,7 @@ import {Suspense} from 'react';
 import {Provider} from 'react-redux';
 import {MemoryRouter, Route, Routes} from 'react-router';
 
+import Notification from './components/Notification';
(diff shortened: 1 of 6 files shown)


eglitise:
Will review this in the near future - but for now I noticed that there is one remaining import of `notification` in `renderer/components/Inspector/GestureEditor.jsx`.
Also is there a need for the event dispatcher/listener? The antd docs for `App` suggest that the `notification` returned by `App.useApp();` can be used as a direct replacement for the regular `notification` import.

PR #2138, opened 2025-06-23
feat: dark theme

### LINKED ISSUE #642: need dark theme for mac. too much glare
## Appium or Appium Desktop?

## The problem

need dark theme for mac. too much glare

## Environment

- I am running Appium Desktop version _v2022.9.1_.
- I am on (pick one):
    - [x] Mac
    - [ ] Windows
    - [ ] Linux

wilhus:
Added dark theme to main window and splash screen. There is a button to toggle the theme at the top right corner of the session builder.

Also fixed button icon size.

Please try it out and let me know what you think!

### Session builder
session-builder-dark

### Session inspector
session-inspector-dark

### Cloud providers
cloud-providers-dark

Closes #642


eglitise:
Thanks for picking up this long-requested feature!
I haven't tested this out yet, but one of the first things I noticed were the dark mode cloud provider logos. If each provider also uses the dark mode version added here, then that's fine, but if not, I would suggest avoiding any color adjustments, as this may potentially cause issues (for example, my company has explicitly defined branding guidelines for its logo colors, which may also be the case here)

app/common/renderer/components/Session/CloudProviders.jsx
```diff
@@ -31,62 +45,110 @@ import ServerTabTestingbot from './ServerTabTestingbot.jsx';
...
+  if (!logos) {
+    return null;
+  }
+
+  const logo = isDarkTheme && logos.dark ? logos.dark : logos.light;
+
+  return <img src={logo} />;
+};
+
+const providerTabs = {
```
eglitise: Could this object be merged with `logoMapping`? Both use the same keys, after all.

app/common/renderer/components/Session/ToggleTheme.jsx
```diff
@@ -0,0 +1,16 @@
...
+import {Button, Tooltip} from 'antd';
+
+import {useTheme} from '../../hooks/use-theme';
+
+const ToggleTheme = ({t}) => {
+  const {toggleTheme} = useTheme();
+
+  return (
+    <Tooltip title={t('Toggle Theme')}>
+      <Button id="btnToggleTheme" icon={<BulbOutlined />} onClick={() => toggleTheme()} />
```
eglitise: Would it be possible to make this a 3-option toggle (light/dark/system), with each using its own icon? You can refer to [the Appium docs](https://appium.io/docs/en/latest/) for one possible icon example. [The Material for MkDocs documentation](https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#color-palette-toggle) also lists a few other combinations, though the React icons pack will likely not have these exact ones.

app/common/renderer/components/Inspector/Inspector.module.css
```diff
@@ -11,9 +11,6 @@
   justify-content: center;
   width: 100%;
   height: 50px;
-  background: linear-gradient(180deg, #e2e2e2, #e0e0e0, 10%, #d6d6d6, 50%, #cacaca);
-  border-top: 1px solid #f5f4f5;
-  border-bottom: 1px solid #a5a4a5;
```
eglitise, line 16: I'm okay with removing the custom gradient, but I do think there should be some division between the header and the rest of the app window. Keeping just the bottom border also doesn't look too great, so I would suggest having some sort of background still.

app/common/renderer/components/Session/Session.module.css
```diff
@@ -37,6 +37,12 @@
 .cloudProviderModal :global(.ant-modal-body button img) {
   height: 100%;
 }
+
+.sessionHeader {
+  display: flex;
```
eglitise, line 46: This specific line breaks navigation between providers, if the user has selected more providers than can fit in the width of the window.

app/common/renderer/components/Session/CloudProviders.jsx
wilhus: Good idea! They are merged now

app/common/renderer/components/Session/ToggleTheme.jsx
wilhus: Changed it into a 3-option toggle. Let me know what you think. The icon size is a bit small but it's being fixed in a separate PR.

app/common/renderer/components/Inspector/Inspector.module.css
wilhus: I added a divider. Does it look good? Otherwise I can try out some backgrounds

app/common/renderer/components/Session/Session.module.css
wilhus: Nice catch! I added a max width to it

app/common/renderer/providers/ThemeProvider.jsx
```diff
@@ -0,0 +1,66 @@
...
+import {getSetting, setSetting} from '../polyfills';
+import {loadHighlightTheme} from '../utils/highlight-theme';
+
+export const ThemeContext = createContext(null);
+
+export const ThemeProvider = ({children}) => {
+  const [preferredTheme, setPreferredTheme] = useState('system');
+
+  const systemPrefersDarkTheme = window.matchMedia('(prefers-color-scheme: dark)').matches;
+  const isDarkTheme = preferredTheme !== 'light' && systemPrefersDarkTheme;
```
eglitise: `isDarkTheme` will be `false` if `preferredTheme == 'dark'` and `systemPrefersDarkTheme == false`.
```suggestion
  const isDarkTheme = preferredTheme == 'dark' || (preferredTheme == 'system' && systemPrefersDarkTheme);
```

app/common/shared/setting-defs.js
```diff
@@ -13,6 +14,7 @@ export const VISIBLE_PROVIDERS = 'VISIBLE_PROVIDERS';
 
 export const DEFAULT_SETTINGS = {
   [PREFERRED_LANGUAGE]: fallbackLng,
+  [PREFERRED_THEME]: '',
```
eglitise: I think this can be set to `system`. Yes, it will be a behavior change, but arguably this should also be the default behavior

app/electron/main/windows.js
```diff
@@ -49,12 +51,16 @@ function buildSessionWindow() {
   });
 }
 
-export function setupMainWindow() {
-  const splashWindow = buildSplashWindow();
+export async function setupMainWindow() {
+  const savedTheme = await settings.get(PREFERRED_THEME);
+  const shouldUseDarkColors = savedTheme ? savedTheme === 'dark' : nativeTheme.shouldUseDarkColors;
```
eglitise: This probably also needs handling for `savedTheme == system`

app/common/renderer/providers/ThemeProvider.jsx
wilhus: Good catch! Fixed now

app/common/shared/setting-defs.js
wilhus: I agree, changed now!

app/electron/main/windows.js
wilhus: Of course! Fixed now

app/common/renderer/components/Session/Session.module.css
```diff
@@ -144,10 +159,16 @@
   padding-left: 12px;
 }
 
+.savedCaps,
 .formattedCaps,
 .formattedCapsBody {
   height: 100%;
 }
+
+.savedCaps :global(.ant-card-body) {
```
eglitise, line 168: This probably also needs `height: 100%`, otherwise if there's more than 7 saved sets, scrolling the list causes the entire table + JSON representation to also scroll

app/common/renderer/components/Session/ToggleTheme.jsx
eglitise: Dropdown works fine but I have a few suggestions:
1. Change the icon into a button to make it more obvious
2. Change the on-hover expand behavior to trigger on-click (and add an on-hover Tooltip to indicate the purpose of this button)
3. Increase size of icons in dropdown entries
4. The system icon doesn't suggest that it is related to themes, so something like `BgColorsOutlined` might work better

app/common/renderer/components/Inspector/Inspector.module.css
eglitise: Yep, looks fine 👍

eglitise:
After some brief testing, most of the elements look quite nice in dark mode! It is maybe a bit _too_ dark though - is it possible to make the background a few shades lighter instead of fully black?
I did also find a few items that don't look too great:
1. Coordinates overlay in the screenshot's Coordinate Mode: 
image
2. Selected element in element search: 
image

app/common/renderer/components/Inspector/Inspector.module.css
```diff
@@ -4,16 +4,17 @@
   height: 100%;
 }
 
+.inspector-container :global(.ant-divider-horizontal) {
+  margin: 0px;
```
eglitise: This unfortunately also affects the Divider used in the gesture editor, which doesn't look too great:
image

eglitise:
Another minor thing (not directly related to this PR) about the Badge colors - currently each Badge element is explicitly set to `blue`, but this can also be set in the theme provider for easier adjustment in dark theme (if needed):
```js
    components: {
      Badge: {
        colorError: '#1677ff',
```
(`blue` is a different shade of blue so it won't work)

app/common/renderer/components/Inspector/Inspector.module.css
wilhus: Should be fixed now!

app/common/renderer/components/Session/Session.module.css
wilhus: Nice catch, fixed now!

app/common/renderer/components/Session/ToggleTheme.jsx
wilhus: I've made changes now based on these suggestions!

app/electron/main/windows.js
```diff
@@ -49,12 +51,17 @@ function buildSessionWindow() {
...
+  const isDarkTheme =
+    preferredTheme === 'dark' || (preferredTheme === 'system' && nativeTheme.shouldUseDarkColors);
+  const backgroundColor = isDarkTheme ? '#000' : '#fff';
+
+  const splashWindow = buildSplashWindow(backgroundColor);
   splashWindow[pathLoadMethod](splashPath);
   splashWindow.show();
 
-  mainWindow = buildSessionWindow();
+  mainWindow = buildSessionWindow(backgroundColor);
```
eglitise, line 64: I just checked this on Windows - unfortunately this does not change the color of the app window's title bar:
image
What _does_ change it is the `nativeTheme.themeSource` setting:
```js
  if (isDarkTheme) {
    nativeTheme.themeSource = 'dark';
  } else {
    nativeTheme.themeSource = 'light';
  }
```
...but in order for the user to be able to also change it (without relaunching the app), IPC is most likely required ([at least according to Electron's tutorial](https://www.electronjs.org/docs/latest/tutorial/dark-mode)).

app/electron/main/windows.js
```diff
@@ -18,27 +18,29 @@ const pathLoadMethod = isDev ? 'loadURL' : 'loadFile';
...
-function buildSessionWindow() {
+function buildSessionWindow(backgroundColor) {
   return new BrowserWindow({
     show: false,
     width: 1100,
     height: 710,
     minWidth: 890,
     minHeight: 710,
     titleBarStyle: 'hiddenInset',
+    backgroundColor,
```
eglitise, line 43: This value seems to be only needed for the splash window, and the main window looks the same without it

app/common/renderer/components/Inspector/Inspector.module.css
```diff
@@ -443,20 +453,14 @@
   min-height: 24px;
   max-height: 150px;
   overflow: auto;
-  border-style: solid;
-  border-width: 1px;
```
eglitise: A border is definitely needed here, especially in light mode:
image
It is true though that keeping these lines in dark mode results in the border being white, which looks a bit out of place, so I would recommend simply adjusting the color

app/common/renderer/components/Inspector/Inspector.module.css
eglitise: Spacing looks good now, but the timeline icons are smaller:
image
My guess is it's related to the font size change for `anticon`.

app/common/renderer/components/Inspector/HeaderButtons.jsx
```diff
@@ -233,12 +233,13 @@ const HeaderButtons = (props) => {
 
   return (
     <div className={InspectorStyles['inspector-toolbar']}>
-      <Space size="middle">
+      <Space size="middle" className={InspectorStyles['inspector-toolbar-controls']}>
```
eglitise: This style does not exist

app/common/renderer/providers/ThemeProvider.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+  const updatePreferredTheme = async (theme) => {
+    setPreferredTheme(theme);
+    await setSetting(PREFERRED_THEME, theme);
+  };
+
+  const themeConfig = {
+    algorithm: isDarkTheme ? theme.darkAlgorithm : theme.defaultAlgorithm,
+    token: {
+      colorBgLayout: isDarkTheme ? '#121212' : '#f5f5f5',
```
eglitise: I still think the dark background could still be a bit lighter, how about `#191919`?

eglitise:
The docs will also need to be updated, but I can handle that in a separate PR.

app/common/renderer/components/Inspector/Inspector.module.css
wilhus: Should be fixed now!

app/electron/main/windows.js
wilhus: I've added an IPC listener for this now. I don't have a Windows machine, could you check if it works as intended now?

app/electron/main/windows.js
wilhus: Removed it!

app/common/renderer/components/Inspector/Inspector.module.css
wilhus: Added a border with adjusted color!

app/common/renderer/components/Inspector/HeaderButtons.jsx
wilhus: Removed it!

app/common/renderer/providers/ThemeProvider.jsx
wilhus: Try it out now! Material design recommended `#121212` (https://m2.material.io/design/color/dark-theme.html#properties), but `#191919` also looks nice!

app/electron/main/windows.js
eglitise: Thanks, it works as expected when switching the theme, but if I relaunch the app, the title bar remains white even in dark mode. I think `nativeTheme.themeSource` needs to also be initially set to the proper value, ~~probably in `setupMainWindow`~~ by calling `setTheme` also in `initTheme`, not just in `updateTheme`.

eglitise:
Just noticed two minor things probably related to antd v5 (again, not due to this PR) - but since you are already changing CSS, would you mind changing these in `Inspector.module.css`?
* For `.inspector-main :global(.ant-card-head)`: add `margin-bottom: 0px;`
  * When selecting an element and scrolling down the attributes list, the list currently overlaps the border with the header
* For `.inspector-main :global(.ant-card-body)`: change the height offset from `48px` to `54px`
  * When opening e.g. Actions tab and expanding a few Collapses, the bottom of the main area slightly extends outside of the bottom border of the card

app/electron/main/windows.js
wilhus: Please try it now again!

eglitise (approved):
Thanks so much! I will test the Windows-specific change later today, but the rest looks great!

app/electron/main/windows.js
eglitise: My mistake - seems like without specifying `backgroundColor` for the main window, when launching the app in dark mode, there is a split-second moment where the light mode background is shown, before changing to the dark version. Once `backgroundColor` is set, the light background is never shown, making for a more consistent UX. Could you please add it back?

eglitise: Thanks, it's fixed now 🙂

PR #2147, opened 2025-06-26
fix: components and icons size

wilhus:
Fixes button icon size and increases the size of Badge and Switch components to match the pre-antd v5 size.


app/common/renderer/Root.jsx
```diff
@@ -22,6 +22,14 @@ const getTheme = () => ({
     fontSize: 12,
   },
   components: {
+    Badge: {
+      indicatorHeight: 20,
```
eglitise, line 26: I think this also needs `textFontSize: 12`

app/common/renderer/assets/stylesheets/main.css
```diff
@@ -64,3 +64,26 @@ body::-webkit-scrollbar-corner {
...
+  font-size: 12px !important;
+}
+
+.ant-switch-inner-checked .anticon {
+  margin-right: 2px !important;
+}
+
+.ant-switch-inner-unchecked .anticon {
+  margin-left: 2px !important;
+}
```
eglitise: I can't see much of a difference with and without these lines 🤔 there's only a small change for the switch that toggles showing/hiding element highlighters, but it's so minor that it can also be omitted

app/common/renderer/components/Session/Session.jsx
```diff
@@ -143,7 +143,7 @@ const Session = (props) => {
               label: (
                 <span>
                   {t('Saved Capability Sets')}{' '}
-                  <Badge count={savedSessions.length} offset={[0, -3]} />
+                  <Badge color="blue" count={savedSessions.length} offset={[0, -3]} />
```
eglitise, line 146: There is also a second Badge element in `components/Inspector/LocatedElements.jsx`

app/common/renderer/Root.jsx
wilhus: Indeed! Added now

app/common/renderer/assets/stylesheets/main.css
wilhus: Yeah they are removed now

app/common/renderer/components/Session/Session.jsx
wilhus: Ah, added the color to it as well

eglitise (approved):
👍

PR #2186, opened 2025-07-17
feat: update app icon

eglitise:
This PR is a visual refresh of the application icon, with a cleaner design and removal of the mobile phone element, since Appium also supports other platforms. The changes also introduce slightly different variations for macOS, Windows, as well as the splash screen/docs welcome page.

For a comparison - from left to right: current icon, new macOS icon, new Windows icon, new splash screen/docs welcome icon.
icons

(I started with the macOS icon in App Composer and later adjusted it for Windows and splash/docs)


KazuCocoa (approved):
Looks good to me

Issue #766, opened 2023-04-10
Ideas for UI/UX improvements


fleytman:
The inspector's appium window is divided into 3 parts. On the left is a screenshot of the device, in the middle the App Source, and on the right the Selected element. ~The problem occurs when you start scrolling down App Source, because the whole window is scrolled. Because of this, I constantly have to scroll up to see what is highlighted in the screenshot and down to examine the structure of the xml. An illogical situation with the Selected element section. I propose to divide these sections so that scrolling works for each of them separately.~ **UPD.** fixed in 2023.3.1+ version

<details>
  <summary>Screenshot</summary>
<img width="1283" alt="Снимок экрана 2023-04-10 в 19 15 14" src="https://user-images.githubusercontent.com/1923645/231009789-076ca59d-eff5-40b7-94c0-0352dde5c644.png">
</details>

Also inconvenient that the size of these sections can not change itself, you can only change the size of the window itself. I think that there are a few solutions here
1. Add delimiters, which will allow you to change the size of sections.
2. To add a possibility to make the screenshot smaller
3. Add the ability to collapse a section
4. Move the right section down, which gives more space for the middle section
5. Minimize information in the middle section, add buttons for collapsing, full-text search

For inspiration, let's see how Layout Inspector looks like in Android studio

## Environment

- I am running Appium Inspector Version: 2023.2.1
- I am on (pick one): Mac

eglitise:
The App Source scrolling issue is fixed in version 2023.3.1.
I do like the suggestions. Not sure how easy it would be to implement them though...

fleytman:
I didn't realize from the list of changes that I had to update to 2023.3.1 and the newer 2023.4.1 wasn't there yet. Tried it, it works - it makes it very easy to work and takes the frustration out of using it.

Updated the original post. The other improvements I suggested, I'm sure can improve the user experience too.

jlipps:
OK, I updated the title here to reflect the enhancement requests

idebenone:
@jlipps Any noise on revamping the UI? 👀
I would love to get my hands dirty!

jlipps:
It would be great to have you get to know the project @idebenone! Maybe @eglitise has some ideas for a good start?

idebenone:
@eglitise I'm all ears!

eglitise:
@idebenone there's quite a lot of room for improvements and enhancements in this project.
[...]

PR #2196, opened 2025-07-21
feat: allow resizing/collapsing source/selected element panels

eglitise:
This PR adds support for resizing and collapsing of the source and selected element panels shown in the Session Inspector, as suggested in #766.
I've also set a fixed width for the fixed columns in the selected element tables (they leave too much blank space until the next column, which looks even worse if the source panel is collapsed), as well as increased the width of the send keys input field.
Here is a brief demo:

https://github.com/user-attachments/assets/4a2eed22-2d3f-406c-a197-c8d25689712c


KazuCocoa (approved):
Looks great 🚀

PR #2200, opened 2025-07-23
feat: show lines between source tree nodes

eglitise:
Small PR to add a useful feature in the antd Tree component - showing lines between sibling nodes in the source tree. I've also included a minor CSS fix for the tree node expand/collapse icon, since it was vertically misaligned.

For a comparison, this is how a tree might look currently:
tree-current

This is how the same tree looks after these changes:
tree-new


KazuCocoa (approved):
oh, that's great

PR #2235, opened 2025-08-11
feat: search app source

### LINKED ISSUE #1004: feature request: Search option at the App Source section

wilhus:
Adds support for searching the app source in the session inspector.

Screenshots

Closes #1004


app/common/renderer/components/Inspector/Source.jsx
```diff
@@ -75,6 +105,38 @@ const Source = (props) => {
...
   const treeData = sourceJSON && recursive(sourceJSON);
 
+  const flatten = (elemObj) => [elemObj, ...(elemObj.children?.flatMap(flatten) || [])];
+
+  const flatTreeData = sourceJSON && flatten(sourceJSON);
+
+  const elementMatchesSearch = (element, searchValue) =>
+    (element.tagName + Object.entries(element.attributes).map(([name, value]) => name + value))
+      .toLowerCase()
+      .includes(searchValue.toLowerCase());
```
eglitise, line 119: I would prefer if the search was case sensitive, but I'm open to changing my mind. Ideally I guess the user would have a choice, but that should probably stay outside the scope of this PR.

app/common/renderer/components/Inspector/Inspector.module.css
```diff
@@ -704,3 +703,33 @@
...
+  max-width: 300px;
+}
+
+.tree-search-input :global(.ant-input-group-addon) {
+  min-width: 46px;
+}
+
+.tree-search-highlight {
+  background-color: #ffeb3b;
+  padding: 1px 2px;
```
eglitise: I don't think the padding is needed, the highlighted background is enough to identify the element, plus having the padding ever so slightly increases the width of the element node.

eglitise:
I noticed that if I search for a very ubiquitous term and then delete the query, this leaves most (if not all) the element paths expanded. Since there's currently no dedicated button for collapsing the whole tree, maybe it's worth doing so upon clearing the search field? (though maybe except if the user has already adjusted the expanded nodes).
In the current implementation, if I search for something, then collapse the highlighted node, and then delete the query, the previously highlighted node is expanded again.

app/common/renderer/components/Inspector/Source.jsx
wilhus: I don't mind either. It felt more natural to have it case insensitive but can be easily changed

app/common/renderer/components/Inspector/Inspector.module.css
wilhus: Removed!

app/common/renderer/components/Inspector/Source.jsx
eglitise: Alright, I think we can leave it as case insensitive for now, and change it later if needed

app/common/renderer/components/Inspector/Inspector.module.css
```diff
@@ -704,3 +703,32 @@
   color: red;
   margin: 5px;
 }
+
+.tree-wrapper {
```
eglitise, line 707: It's not required, but maybe it would be easier to use antd `Space` instead of custom CSS?

app/common/renderer/components/Inspector/Inspector.module.css
```diff
@@ -704,3 +703,32 @@
...
+  gap: 8px;
+}
+
+.tree-search-input {
+  max-width: 300px;
+}
+
+.tree-search-input :global(.ant-input-group-addon) {
+  min-width: 46px;
+}
```
eglitise, line 721: Not sure if this is needed, it adds too much padding if there's <10 items found.

app/common/renderer/components/Inspector/Inspector.module.css
```diff
@@ -704,3 +703,32 @@
...
+
+.tree-wrapper {
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  gap: 8px;
+}
+
+.tree-search-input {
+  max-width: 300px;
```
eglitise: Minor thing, but if we're aligning with the selected element actions, this could be changed to 400px, similarly to `elementKeyInputActions`

wilhus: It's now wider and centered!

app/common/renderer/components/Inspector/Inspector.module.css
wilhus: Yeah it can look a bit weird when there's few matches but it was added because the width was changing so much depending on how many items that are found

app/common/renderer/components/Inspector/Inspector.module.css
wilhus: I tried but it seems to be messing a lot with the Tree, it doesn't seem to fill out the space properly

app/common/renderer/components/Inspector/Inspector.module.css
eglitise: I guess I find the width change to be expected behavior, but it's not a big issue.

eglitise (approved):

KazuCocoa (approved):

PR #2249, opened 2025-08-18
feat: enable/disable capabilities

### LINKED ISSUE #2036: feature request: Add a switch to enable/disable capability in capability builder for ease of reuse
### Current Behavior

While creating an Appium session, user needs to define the capability in capability builder section of Appium inspector.
At times, we need to experiment with new capabilities and/or change values of the existing capabilities. To achieve this, we need to delete the capability for every change and add/edit new capability from scratch again which is often annoying and time consuming. 

### Suggested Solution

I would like to have some sort of switch next to delete icon in front of the capability in the same row which can be turned on/off to enable/disable certain capability when starting session.

This way I do not have to manually enter the same capabilities again if i want to reuse some of the capabilities and I can simply enable them.

### Additional Information

For example, if i want to start a session with appium:language set to "en", i can add this capability and leave the switch to true to start a session wiith this capability.
Next time, when i want to not use this capability, i can move the switch to Off (disable state) which will remove this capability from the JSON representation and start session without it.

wilhus:
Added a checkbox to include/exclude capabilities from the session. The excluded capabilities will be hidden from the JSON representation but not the raw JSON. Feels like it makes the most sense but let me know if it should be changed.

SCR-20250818-icqh

Closes #2036

### DIFF
--- a/app/common/renderer/components/Session/CapabilityEditor.jsx
+++ b/app/common/renderer/components/Session/CapabilityEditor.jsx
@@ -127,7 +127,7 @@ const CapabilityEditor = (props) => {
                   </Select>
                 </Form.Item>
               </Col>
-              <Col md={8} lg={9} xl={9} xxl={10}>
+              <Col flex="1">
                 <Form.Item>
                   <Tooltip title={whitespaceMsg(cap.value, t)} open={whitespaces.test(cap.value)}>
                     <CapabilityControl
@@ -140,9 +140,9 @@ const CapabilityEditor = (props) => {
                   </Tooltip>
                 </Form.Item>
               </Col>
-              <Col md={4} lg={3} xl={3} xxl={2}>
+              <Col>
                 <Form.Item>
-                  <Space className={SessionStyles.capActions}>
+                  <Space>
                     <Tooltip title={t('Enable')} placement="right">
                       <Checkbox
                         disabled={isEditingDesiredCaps}
--- a/app/common/renderer/components/Session/Session.module.css
+++ b/app/common/renderer/components/Session/Session.module.css
@@ -280,10 +280,6 @@
   margin-left: 8px;
 }
 
-.capActions {
-  float: right;
-}
-
 .inputDataCenter {
   margin-left: 8px;
 }


app/common/public/locales/en/translation.json
```diff
@@ -303,5 +303,6 @@
   "yourProjectId": "Your Project ID",
   "webmateProjectId": "webmate Project ID",
   "specifyWebmateHostExplicitly": "Specify webmate host explicitly",
-  "webmateHost": "webmate Host"
+  "webmateHost": "webmate Host",
+  "Include": "Include"
```
eglitise: I think `Enable` would be clearer

app/common/renderer/actions/Session.js
```diff
@@ -828,6 +830,7 @@ export function saveRawDesiredCaps() {
         })(),
         name,
         value,
+        enabled: caps[name] ? caps[name].enabled : true,
```
eglitise:
```suggestion
        enabled: caps[name]?.enabled ?? true,
```

app/common/renderer/reducers/Session.js
```diff
@@ -129,7 +130,10 @@ export default function session(state = INITIAL_STATE, action) {
         ...state,
         server: action.server,
         serverType: action.serverType,
-        caps: action.caps,
+        caps: action.caps.map((cap) => ({
+          ...cap,
+          enabled: cap.enabled !== undefined ? cap.enabled : true,
```
eglitise:
```suggestion
          enabled: cap.enabled ?? true,
```

app/common/renderer/components/Session/CapabilityEditor.jsx
```diff
@@ -129,17 +141,24 @@ const CapabilityEditor = (props) => {
                 </Form.Item>
               </Col>
               <Col flex="40px">
```
eglitise: This and other column widths should be adjusted, otherwise certain window widths cause the CSS to break:
image

eglitise:
Functionality LGTM 👍

app/common/public/locales/en/translation.json
wilhus: I agree! Changed

app/common/renderer/actions/Session.js
wilhus: 👍

app/common/renderer/reducers/Session.js
wilhus: 👍

app/common/renderer/components/Session/CapabilityEditor.jsx
wilhus: Should be fixed now!

eglitise: Now it's the opposite problem, where there's too much empty space left of the checkbox on large window sizes:
image
This issue was the reason why I had previously set a fixed width for the delete button, and `flex` for the capability type, but maybe there's a better solution.

app/common/renderer/components/Session/CapabilityEditor.jsx
```diff
@@ -84,7 +96,7 @@ const CapabilityEditor = (props) => {
         <Form className={SessionStyles.newSessionForm}>
           {caps.map((cap, index) => (
             <Row gutter={8} key={index}>
-              <Col md={7} lg={8} xl={8} xxl={9}>
+              <Col md={7} lg={7} xl={7} xxl={7}>
```
eglitise: If setting the same width for all sizes, it's better to just replace them all with a `span`, e.g. `<Col span={7}>`

wilhus: 👌

app/common/renderer/components/Session/CapabilityEditor.jsx
wilhus: I tried another approach with `flex`, maybe it's better now?

eglitise: Seems to be fixed now, thanks!

eglitise (approved):

Issue #2257, opened 2025-08-20
feature request: UI/UX enhancements for getLogs


itkhanz:
### Current Behavior

In Appium Inspector, when user wants to see the device logs e.g. for Android, getLogs: logcat
then the logs are displayed in the popup, however I find it a bit difficult to look for relevant logs because of following issues.

<img width="809" height="845" alt="Image" src="https://github.com/user-attachments/assets/6a7a2cb2-c72a-40f6-9cdb-5622633dd231" />

### Suggested Solution

**Increase the popup width**
popup width is fixed, even though the actual Appium inspector is maximized so it should be possible to increase the width of popup so message can be displayed fully or atleast utilize the max avaialable width. I also noticed that the horizontal scrollbar is also not visible or it is not sticky causing it to not useful because it only appears if you scroll all the way to the bottom.

**Add search field to filter the logs with specific text**
Currently large amount of logs are shown, and the slider to navigate is shown in the main appium inspector window rather than the popup causing confusion. It would be helpful to have a search logs field at the top which user can use to filter the logs.

**Export logs**
An option to export logs to text file would be helpful if as a user we want to share the logs with someone.

PR #2317, opened 2025-09-17
feat: UI/UX enhancements for command result

### LINKED ISSUE #2256: feature request: Add a copy to clipboard button for getUrl and getTitle
### Current Behavior

In Appium Inspector, when user switches to web view context, and calls the following commands in Web tab:-

- getTitle
- getUrl

then the page title, and Url is displayed. But it is a bit annoying for having to manually copy the URL and Title and then remove the starting and trailing "" from the actual value.
The fact that user cannot copy the text by selecting and right clicking the value adds to misery, because it shows inspect element dialog in MAC, and hence only option to copy the value is using keyboard shortcut Ctrl/Cmd+C

Image

### Suggested Solution

There should be a 'copy to clipboard' button within the popup which should copy the displayed value to the clipboard, and automatically trim the leading and trailing ""

### Additional Information

There is an icon for 'Copy Attributes to Clipboard' in the selected element tab in source window, which may fit the need for this button as well.

wilhus:
This is a suggestion for UI/UX enhancements of command result by rendering with a table. It's possible to copy values by clicking the table cell, and the columns can be sorted/filtered. Please try it out and let me know if the idea is good and how it can be improved.

| Old | New |
| - | - |
| SCR-20250917-ozvb | SCR-20250917-pbks |
|  SCR-20250917-paep | SCR-20250917-pbqi |
|  SCR-20250917-pavk | SCR-20250917-pbtq |

Closes #2256, #2257

### DIFF
--- a/app/common/renderer/components/SessionInspector/CommandsTab/CommandResultTable.jsx
+++ b/app/common/renderer/components/SessionInspector/CommandsTab/CommandResultTable.jsx
@@ -18,7 +18,7 @@ const formatValueWithTimestamp = (value, key = null) =>
 
 const CommandResultTableCell = ({value, dataIndex}) => {
   const [isCopied, setIsCopied] = useState(false);
-  const displayText = String(formatValueWithTimestamp(value, dataIndex));
+  const displayText = String(formatValueWithTimestamp(value, dataIndex)).trim();
 
   const handleCopy = async () => {
     await copyToClipboard(displayText);


eglitise:
Thanks for this - I haven't tested this out yet, but overall I like the idea. I do have some concerns though:
* I don't think this formatted view should be visible _by default_. My suggestion would be to make it accessible through a toggle (e.g. in the bottom left corner of the modal), while the default view should still show the raw returned value - it is critical to have for users that call the command through their automation code, since they need to know the exact message format.
  * However, since the raw view doesn't set a limit on the modal height, which you have solved here using the table, I would suggest also wrapping the raw view with a card/table/other component.
* I would also recommend adding a button to copy the entire raw result to the clipboard. This can be located next to the formatting toggle.

app/common/renderer/components/SessionInspector/SessionInspector.jsx
```diff
@@ -340,10 +341,9 @@ const Inspector = (props) => {
         open={!!visibleCommandResult}
         onOk={() => setVisibleCommandResult(null)}
         onCancel={() => setVisibleCommandResult(null)}
+        width={900}
```
eglitise, line 344: This width exceeds the default width of the Inspector window. Responsive width could be a useful approach here: https://ant.design/components/modal#modal-demo-width
In the ideal case, the modal width would also shrink even further if the contents don't require all of it, but I'm not sure how possible this is. It's definitely more of a nice-to-have feature.

app/common/renderer/components/SessionInspector/CommandsTab/Commands.module.css
```diff
@@ -13,3 +13,14 @@
 .argContainer {
   padding: 4px 4px 0 0;
 }
+
+.command-result-table :global(.ant-table-tbody) {
```
eglitise, line 17: For consistency, please use camelCase for CSS class names

app/common/renderer/components/SessionInspector/CommandsTab/CommandResultTable.jsx
```diff
@@ -0,0 +1,150 @@
+import {Table, Tooltip} from 'antd';
+import {useState} from 'react';
+
+import {copyToClipboard} from '../../../polyfills.js';
+import styles from './Commands.module.css';
+
+const isTimestampKey = (key) => key === 'timestamp';
```
eglitise, line 7: I'm not sure how useful this conversion is, since it will only work for the `timestamp` key, but the result may include timestamp values under another key. Rendering the raw value is perfectly fine, in my opinion.

app/common/renderer/components/SessionInspector/CommandsTab/CommandResultTable.jsx
```diff
@@ -0,0 +1,150 @@
...
+  const displayText = String(formatValueWithTimestamp(value, dataIndex)).trim();
+
+  const handleCopy = async () => {
+    await copyToClipboard(displayText);
+    setIsCopied(true);
+    setTimeout(() => setIsCopied(false), 2000);
+  };
+
+  return (
+    <Tooltip placement="topLeft" title={isCopied ? 'Copied' : 'Click to copy'}>
```
eglitise, line 30: These texts will need to have i18n support, so I would suggest simply reusing the `Copied!` translation key from `SelectedElement.jsx`. `Click to copy` can be omitted if the cursor is changed accordingly.

app/common/renderer/components/SessionInspector/CommandsTab/CommandResultTable.jsx
```diff
@@ -0,0 +1,150 @@
...
+      if (typeof data[0] === 'object') {
+        result = handleArrayOfObjects(data);
+      } else {
+        result = handleArrayOfPrimitives(data);
+      }
+    } else if (typeof data === 'object') {
+      const keys = Object.keys(data);
+      if (keys.length === 1) {
+        return createTableResult(data[keys[0]]);
+      }
```
eglitise, line 106: This will result in the top-level key being hidden from the user

app/common/renderer/components/SessionInspector/CommandsTab/CommandResultTable.jsx
```diff
@@ -0,0 +1,150 @@
...
+    const parsedResult = JSON.parse(result);
+
+    return createTableResult(parsedResult);
+  };
+
+  const tableData = formatCommandResultForTable(result);
+
+  if (!tableData) {
+    return <div>No data to display</div>;
+  }
```
eglitise, line 134: This probably won't be needed if the formatting option is disabled for primitive return values

app/common/renderer/components/SessionInspector/CommandsTab/CommandResultTable.jsx
```diff
@@ -0,0 +1,150 @@
...
+
+  return (
+    <Table
+      dataSource={tableData.dataSource}
+      columns={tableData.columns}
+      pagination={false}
+      size="small"
+      scroll={{y: 400, x: 'max-content'}}
```
eglitise, line 142: I wonder if the `y` value can be made responsive similarly to the width 🤔 I looked into this a few weeks ago, but couldn't find an answer right away.

PR #2328, opened 2025-09-22
feat: Improve Appium Inspector stability by handling USB/Wifi disconnection gracefully

mohangowdak:
When using Appium Inspector, we observed an issue where the inspector stops working if the USB/Wifi connection is lost after a session has started. In order to continue inspecting and interacting with other pages, users previously had to quit the session, re-enter all capabilities, and start a new session. This process was both time-consuming and inconvenient.

This PR introduces a fix to handle session loss more seamlessly. Now, whenever a debugging disconnection occurs, the old session is automatically quit, and the page source is refreshed. This allows users to continue their workflow without needing to exit the inspector window or manually reconfigure the session.

With this improvement, users can enjoy a smoother and more efficient experience when working with Appium Inspector.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,7 +32,7 @@
         "react-i18next": "16.0.0",
         "react-icons": "5.5.0",
         "react-redux": "9.2.0",
-        "react-router": "7.9.3",
+        "react-router": "7.9.4",
         "webdriver": "9.20.0",
         "xpath": "0.0.34"
       },
@@ -14934,9 +14934,9 @@
       }
     },
     "node_modules/react-router": {
-      "version": "7.9.3",
-      "resolved": "https://registry.npmjs.org/react-router/-/react-router-7.9.3.tgz",
-      "integrity": "sha512-4o2iWCFIwhI/eYAIL43+cjORXYn/aRQPgtFRRZb3VzoyQ5Uej0Bmqj7437L97N9NJW4wnicSwLOLS+yCXfAPgg==",
+      "version": "7.9.4",
+      "resolved": "https://registry.npmjs.org/react-router/-/react-router-7.9.4.tgz",
+      "integrity": "sha512-SD3G8HKviFHg9xj7dNODUKDFgpG4xqD5nhyd0mYoB5iISepuZAvzSr8ywxgxKJ52yRzf/HWtVHc9AWwoTbljvA==",
       "license": "MIT",
       "dependencies": {
         "cookie": "^1.0.1",
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "react-i18next": "16.0.0",
     "react-icons": "5.5.0",
     "react-redux": "9.2.0",
-    "react-router": "7.9.3",
+    "react-router": "7.9.4",
     "webdriver": "9.20.0",
     "xpath": "0.0.34"
   },


KazuCocoa:
Thank you for the explanation.

I personally like not to hide too much like this, re-creating a new session behind the error, because this re-created session might have a different state before the error.

So, I'm wondering if it makes sense to add an option to enable this kind of behavior, or add a button to do this reconnection behavior explicitly

app/common/public/locales/ar/translation.json
```diff
@@ -314,5 +314,9 @@
   "Fireflink DeviceFarm Domain": "Fireflink DeviceFarm Domain",
   "Fireflink DeviceFarm Access Key": "Fireflink DeviceFarm Access Key",
   "Fireflink DeviceFarm License ID": "Fireflink DeviceFarm License ID",
-  "Fireflink DeviceFarm Project Name": "Fireflink DeviceFarm Project Name"
+  "Fireflink DeviceFarm Project Name": "Fireflink DeviceFarm Project Name",
```
KazuCocoa: You don't need to change language resources except for `en`. Our localization flow will cover other languages. 
https://github.com/appium/appium-inspector?tab=readme-ov-file#localization

app/web/polyfills.js
```diff
@@ -33,17 +33,38 @@ const browserUtils = {
...
+}
+
 class BrowserSettings {
   has(key) {
     return this.get(key) !== null;
   }
 
   set(key, val) {
-    return localStorage.setItem(key, JSON.stringify(val));
+    if (isLocalStorageAvailable()) {
```
KazuCocoa: Do you need these changes? If not, could you keep the existing one to simplify the condition

app/common/renderer/actions/SessionInspector.js
```diff
@@ -327,9 +328,35 @@ export function applyClientMethod(params) {
       window.dispatchEvent(new Event('resize'));
       return commandRes;
     } catch (error) {
-      log.error(error);
-      showError(error, {methodName: params.methodName, secs: 10});
-      dispatch({type: METHOD_CALL_DONE});
+      if (!getState().inspector.autoSessionRestart) {
```
KazuCocoa: I assume here needs to detect what kind of error should try out this idea.

For example, when you do open a url command for a session which doesn't support the endpoint. Then the session gets an error. Such kind of command error should not get restarted

app/common/public/locales/ar/translation.json
mohangowdak: Got it, thanks for pointing that out! I’ve removed the changes from the other language folders and kept only the updates in en.

app/web/polyfills.js
mohangowdak: Good point, I’ve reverted it back to the existing code to keep the condition simpler.

app/common/renderer/actions/SessionInspector.js
mohangowdak: Thanks for the suggestion! I’ve added a condition to handle these cases now, checking for:
- error.message === "Cannot read properties of undefined (reading 'apply')"
- error.includes "process is not running (probably crashed)"

app/common/renderer/actions/SessionInspector.js
```diff
@@ -327,9 +329,40 @@ export function applyClientMethod(params) {
       window.dispatchEvent(new Event('resize'));
       return commandRes;
     } catch (error) {
+      const {inspector} = getState();
+      const isWebDriverError = error
+        .toString()
+        .includes('process is not running (probably crashed)');
```
KazuCocoa: The message would depend on the implementation on the driver side, so this might be difficult to maintain. Let me think about this a bit

mohangowdak: I understand your concern. However, in the current implementation, Appium Inspector error messages is primarily dependent on the Appium driver error messages, so it might make sense to rely on the message at this stage. At the same time, if you suggest any other possibilities or alternatives that we could implement in parallel, I’d be glad to explore them. I’m seeking your suggestion on this.

mohangowdak: Hi @KazuCocoa ,
In the recent push, I added a custom error message to help identify whether the session has crashed. please review it from your side and I’m happy to discuss any improvements or alternatives you might suggest.

app/common/renderer/components/SessionInspector/SessionInspector.jsx
```diff
@@ -222,6 +226,19 @@ const Inspector = (props) => {
...
+          placement="topRight"
+        >
+          <Switch
+            checked={autoSessionRestart}
+            checkedChildren={<CheckCircleOutlined />}
+            unCheckedChildren={<CloseCircleOutlined />}
+            defaultChecked={autoSessionRestart}
+            onChange={() => toggleAutoSessionRestart()}
+          />
+        </Tooltip>
```
eglitise: This toggle is not in the best location. It has no relation to the app screenshot, and affects the session as a whole, so it's better to move it to the top header. For consistency I would also suggest changing this from a switch to a toggleable button, like the recorder.

app/common/renderer/components/SessionBuilder/CapabilityBuilderTab/CapabilityEditor.jsx
```diff
@@ -173,6 +175,15 @@ const CapabilityEditor = (props) => {
...
               </Form.Item>
+
+              <Form.Item>
+                <Checkbox
+                  checked={autoSessionRestart}
+                  onChange={(e) => setAutoSessionRestart(e.target.checked)}
+                >
+                  {t('AutomaticRestartSession')}
+                </Checkbox>
+              </Form.Item>
```
eglitise: I don't think this checkbox is needed at all, since the functionality can already be toggled during the session itself (where it makes the most sense).

app/common/renderer/actions/SessionInspector.js
```diff
@@ -328,8 +330,40 @@ export function applyClientMethod(params) {
...
+        await quitSes(dispatch, getState);
+        await newSes(dispatch, getState);
+        await getPageSrc(dispatch, getState);
+        await storeSessionSet(dispatch, getState);
+        await getSavedClientFrame(dispatch);
+        runKeepAliveLp(dispatch, getState);
+        setSesTime(dispatch);
+        dispatch({type: SET_AUTO_SESSION_RESTART, autoSessionRestart: true});
+        dispatch({type: METHOD_CALL_DONE});
+      }
```
eglitise: Suggest extracting all of this block into a dedicated method, something like this
```js
if (getState().inspector.autoSessionRestart) {
  return restartSession();
}
showError(error, {methodName: params.methodName, secs: 10});
dispatch({type: METHOD_CALL_DONE});
```
Then inside `restartSession()` you can do all the checks for specific error messages and so on.

app/common/renderer/actions/SessionInspector.js
```diff
@@ -328,8 +330,40 @@ export function applyClientMethod(params) {
...
+      const sesState = await inspectorDriver.getSessionState();
+      const isWebDriverError = !!sesState && sesState === 'Session Expired';
+      if (!(inspector.autoSessionRestart && (isWebDriverError || isApplyError))) {
+        showError(error, {methodName: params.methodName, secs: 10});
+        dispatch({type: METHOD_CALL_DONE});
+      } else {
+        showError(error, {methodName: params.methodName, secs: 3});
+        notification.info({
+          message: i18n.t('Information'),
+          description: i18n.t('RestartSessionMessage'),
```
eglitise: A generic 'Information' title doesn't serve much purpose, so it's better to simplify this to
```suggestion
          message: i18n.t('RestartSessionMessage'),
```

app/common/public/locales/en/translation.json
```diff
@@ -314,5 +314,10 @@
   "Fireflink DeviceFarm Domain": "Fireflink DeviceFarm Domain",
   "Fireflink DeviceFarm Access Key": "Fireflink DeviceFarm Access Key",
   "Fireflink DeviceFarm License ID": "Fireflink DeviceFarm License ID",
-  "Fireflink DeviceFarm Project Name": "Fireflink DeviceFarm Project Name"
+  "Fireflink DeviceFarm Project Name": "Fireflink DeviceFarm Project Name",
+  "AutomaticRestartSession": "Automatically, the session will restart when the connection is lost",
+  "RestartSessionMessage": "Please wait while the page refreshes. The session was restarted because the connection was lost.",
```
eglitise:
```suggestion
  "RestartSessionMessage": "Device connection lost - restarting the session...",
```

app/common/renderer/actions/SessionInspector.js
```diff
@@ -1096,3 +1132,17 @@ export function toggleShowAttributes() {
     dispatch({type: TOGGLE_SHOW_ATTRIBUTES});
   };
 }
+
+export function getAutoSessionState() {
```
eglitise: The method is called `getAutoSessionState` but it dispatches `SET_AUTO_SESSION_RESTART`

eglitise:
I would also prefer if you would undo the changes to documentation images, because the documentation itself would also need to be updated, and I can do that in a separate PR later.

app/common/renderer/components/SessionInspector/SessionInspector.jsx
mohangowdak: Thank you for the thoughtful feedback! I’ve updated the implementation and moved the toggle to the top header. I’ve also changed it to a toggleable button as recorder.

app/common/renderer/components/SessionBuilder/CapabilityBuilderTab/CapabilityEditor.jsx
mohangowdak: Thanks for pointing that out! I’ve removed the checkbox in session builder page since the functionality is already handled within the session inspector page.

app/common/renderer/actions/SessionInspector.js
mohangowdak: Thank you for the suggestion! I’ve created a dedicated restartSession function and moved all the restart session logic into it as suggested.

app/common/public/locales/en/translation.json
mohangowdak: I’ve updated as you suggested

app/common/renderer/actions/SessionInspector.js
mohangowdak: I’ve updated as you suggested

app/common/renderer/actions/SessionInspector.js
mohangowdak: This function was originally used to carry the state from the session-builder page(checkbox) to the session-inspector page. Since the checkbox has been removed from the session-builder page, this function is no longer needed, so I’ve removed it.

app/common/renderer/components/SessionInspector/SessionInspector.jsx
mohangowdak: image

app/common/renderer/components/SessionInspector/Header/HeaderButtons.jsx
```diff
@@ -171,52 +174,79 @@ const HeaderButtons = (props) => {
...
-          <Button
-            id="btnStartRefreshing"
-            icon={<PlayCircleOutlined />}
-            onClick={toggleRefreshingState}
-          />
-        </Tooltip>
-      )}
-      {isUsingMjpegMode && isSourceRefreshOn && (
-        <Tooltip title={t('Pause Refreshing Source')}>
+    <>
```
eglitise: You don't need to modify all of the general controls. Just add a `sessionReloadButton` constant, similarly to `quitSessionButton`.

app/common/renderer/actions/SessionInspector.js
```diff
@@ -328,12 +330,47 @@ export function applyClientMethod(params) {
...
       showError(error, {methodName: params.methodName, secs: 10});
       dispatch({type: METHOD_CALL_DONE});
+      if (inspector.autoSessionRestart && (isWebDriverError || isApplyError)) {
+        notification.info({
+          message: i18n.t('RestartSessionMessage'),
+          duration: 10,
+        });
+        const restartSes = restartSession();
+        await restartSes(dispatch, getState);
+      }
```
eglitise: You can still move most of this code into the `restartSession` method, so that if the user has disabled auto session restart, the driver states and errors are not retrieved. Refer to my previous example for how the catch block can look:
```js
} catch (error) {
  log.error(error);
  if (getState().inspector.autoSessionRestart) {
    return restartSession();
  }
  showError(error, {methodName: params.methodName, secs: 10});
  dispatch({type: METHOD_CALL_DONE});
}
```

app/common/public/locales/en/translation.json
```diff
@@ -314,5 +314,8 @@
   "Fireflink DeviceFarm Domain": "Fireflink DeviceFarm Domain",
   "Fireflink DeviceFarm Access Key": "Fireflink DeviceFarm Access Key",
   "Fireflink DeviceFarm License ID": "Fireflink DeviceFarm License ID",
-  "Fireflink DeviceFarm Project Name": "Fireflink DeviceFarm Project Name"
+  "Fireflink DeviceFarm Project Name": "Fireflink DeviceFarm Project Name",
+  "RestartSessionMessage": "Device connection lost - restarting the session...",
+  "EnableRestartSession": "Enable Auto Restart Session",
+  "DisableRestartSession": "Disable Auto Restart Session"
```
eglitise: You can use a single `Tooltip` component with the text `Toggle Session Reload Upon Device Disconnect`

app/common/renderer/components/SessionInspector/Header/HeaderButtons.jsx
mohangowdak: Sure, as you suggested, I have created a separate sessionReloadButton constant, similar to quitSessionButton.

app/common/public/locales/en/translation.json
mohangowdak: Thanks for the suggestion. I’ve updated the code to use a single Tooltip component with the text Toggle Session Reload Upon Device Disconnect.

app/common/renderer/actions/SessionInspector.js
mohangowdak: Gotcha, Now i've updated the code as you suggested.

app/common/renderer/actions/SessionInspector.js
```diff
@@ -328,12 +330,51 @@ export function applyClientMethod(params) {
...
       showError(error, {methodName: params.methodName, secs: 10});
       dispatch({type: METHOD_CALL_DONE});
     }
   };
 }
 
+export function restartSession(error, params) {
+  return async (dispatch, getState) => {
+    const isApplyError = error.message === "Cannot read properties of undefined (reading 'apply')";
```
eglitise: Could you explain the situation in which you get this error? I was attempting to replicate various interactions after disconnecting the device under test, but only got the session crashed/socket hang up error.

app/common/renderer/actions/SessionInspector.js
```diff
@@ -328,12 +330,51 @@ export function applyClientMethod(params) {
...
+    const inspectorDriver = InspectorDriver.instance(getState().inspector.driver);
+    const sesState = await inspectorDriver.getSessionState();
+    const isWebDriverError = !!sesState && sesState === 'Session Expired';
```
eglitise: This feels like a bit of a hacky way of checking for the session state. A better approach would be to define a new method in this file that calls `callClientMethod` with `methodName` set to `getAppiumSessionCapabilities` (since `getSession` is deprecated), and catch any raised error. Then call that method here, and remove the `getSessionState` method.

app/common/renderer/actions/SessionInspector.js
```diff
@@ -328,12 +330,51 @@ export function applyClientMethod(params) {
...
+    if (isWebDriverError || isApplyError) {
```
eglitise: It's cleaner if you first check that the error _doesn't_ match the wanted ones, since then you can just return the default response, and avoid needing to define an `else` block.

app/common/renderer/lib/appium/inspector-driver.js
```diff
@@ -381,24 +398,39 @@ export default class InspectorDriver {
...
 
-  async getSourceUpdate() {
+  async getSourceUpdate(autoSessionRestart) {
     try {
       const source = parseHtmlSource(await this.driver.getPageSource());
       return {source};
     } catch (err) {
+      if (autoSessionRestart) {
+        throw new Error(err);
+      }
```
eglitise: Are these needed? Won't the method call raise an error anyway once the device is disconnected and the session expires?

app/common/renderer/actions/SessionInspector.js
mohangowdak: Thanks for checking! This error typically occurs when we attempt to interact with an element (for example, performing actions like click, sendKeys, or clear) after the device has been disconnected.

image

app/common/renderer/actions/SessionInspector.js
mohangowdak: Thanks for the suggestion. Since we’re already handling errors inside the callClientMethod function, using it directly here might not be the most suitable approach (because it may goes recursive mode). 
This function was created separately to provide a custom error message, rather than relying on the driver’s default error messages, which could make maintenance harder in the long run.
You are absolutely right about getSession being deprecated, we can update this to use getAppiumSessionCapabilities instead. I’ll hold off on finalizing the change until I get your confirmation.

app/common/renderer/actions/SessionInspector.js
mohangowdak: Sure, with that approach, the catch block would look like this. Please let me know if this aligns with what you had in mind:

```
catch (error) {
  log.error(error);
  const { inspector } = getState();
  const isApplyError =
    error.message === "Cannot read properties of undefined (reading 'apply')";
  const inspectorDriver = InspectorDriver.instance(inspector.driver);
  const sesState = await inspectorDriver.getSessionState();
  const isWebDriverError = !!sesState && sesState === 'Session Expired';

  if (inspector.autoSessionRestart && (isWebDriverError || isApplyError)) {
    const restartSes = restartSession(error, params);
    return await restartSes(dispatch, getState);
  }

  showError(error, { methodName: params.methodName, secs: 10 });
  dispatch({ type: METHOD_CALL_DONE });
}
```

app/common/renderer/lib/appium/inspector-driver.js
mohangowdak: Yes, these checks are required. The main idea is to centralize the verification of whether the session has truly expired when the user interacts with reload/back/home/option buttons. If the session is expired, we can provide a custom error message. Without this, the error is simply handled as return { sourceError: err }, and we wouldn’t have a clear indication of the session state.

app/common/renderer/actions/SessionInspector.js
eglitise: The code changes made here are in `applyClientMethod`, not `callClientMethod`, so as long as you use `callClientMethod` directly and avoid calling `applyClientMethod`, it won't be recursive.
And since the `getSessionState` function doesn't check the driver's error message, you won't have to define a custom message for my suggested function either.

app/common/renderer/actions/SessionInspector.js
eglitise: I don't mean changing the catch block of `applyClientMethod`, that looks fine. Just change the `restartSession` method to something like:
```js
// get isWebDriverError and isApplyError
if (!isWebDriverError && !isApplyError) {
  // call the same 2 lines as in `applyClientMethod`, but this is okay
  showError(error, {methodName: params.methodName, secs: 10});
  return dispatch({type: METHOD_CALL_DONE});
}
showError(error, {methodName: params.methodName, secs: 3});
dispatch({type: METHOD_CALL_DONE});
// the rest of the code in the current if block
```

app/common/renderer/lib/appium/inspector-driver.js
eglitise: What I mean is that if the session is truly expired (as opposed to just the source or screenshot retrieval failing), the server throws an error anyway, due to the inability to connect to the device. I'm not sure if these additional throws add anything.

app/common/renderer/actions/SessionInspector.js
eglitise: I still can't quite reproduce this myself, both with wired and wirelessly connected devices, though I was testing on the production version of the Inspector, without your changes. Maybe the error change is caused by the additional throws added in `inspector-driver.js`?

mohangowdak: Thanks for looking into this! I don’t think the issue is related to the changes in inspector-driver.js. If you could try the following steps, I believe the error will be reproducible:
- Connect the device and start a session → open any app → refresh the screen → disconnect and reconnect the device → now select any element and click Tap.
- Connect the device and start a session → open any app → refresh the screen → navigate to another page using the mobile device (not through the Inspector) → now select any element visible in the Inspector and click Tap.

Please let me know if it’s still not reproducible on your side, I really appreciate your help in checking this!

app/common/renderer/actions/SessionInspector.js
mohangowdak: Thanks for your detailed explanation, I got your point. 

Basically, here I’m trying to implement a lightweight health check for the session. One concern I ran into is that when I try using the getAppiumSessionCapabilities method, even if the device is disconnected, it still returns a success response. Because of this, I’m not able to perform a proper health check.

I was thinking of implementing it this way:

```
const callAction = callClientMethod({ methodName: 'sessionHealthCheck' }); // lightweight ping
const { isSessionAlive } = await callAction(dispatch, getState);
```


And inside the run function of session-inspector.js:

```
if (methodName === 'sessionHealthCheck') {
  let isSessionAlive = false;
  try {
    await this.driver.getOrientation();
    isSessionAlive=true;
  } catch {
    isSessionAlive = false;
  }
  return { isSessionAlive };
}
```


This approach seems to work properly for the health check. I wanted to get your suggestion before incorporating it, to make sure this aligns with best practices.

app/common/renderer/actions/SessionInspector.js
mohangowdak: Sure, I will update it like this in the next push. Thanks!

app/common/renderer/lib/appium/inspector-driver.js
mohangowdak: Here’s the full function for context:

```
async getScreenshotUpdate(autoSessionRestart) {
  try {
    const screenshot = await this.driver.takeScreenshot();
    return { screenshot };
  } catch (err) {
    if (autoSessionRestart) {
      throw new Error(err);
    }
    return { screenshotError: err };
  }
}

```

Since the error is already handled in the function’s catch block, it will always return an object and not throw meaning it won’t propagate to the applyClientMethod catch block. The additional throw I added ensures that the error reaches that outer catch block so we can handle it appropriately there.

I wanted to check if you see any alternative approach here, apart from this explicit throw, that could achieve the same propagation without breaking the current flow.

app/common/renderer/actions/SessionInspector.js
eglitise: Thanks, I can reproduce this now. Unfortunately this reveals that proper handling of device disconnection would require additional complexity: in your first example, the Appium server log returned HTTP 500, whereas in the second it was HTTP 404. Since the second case does not involve device disconnection, relying on the error you see in the GUI would result in this case incorrectly triggering a session reload.

My guess is that since the HTTP 500 error is coming from the `elementFind` command, it would likely need to be handled in [`inspector-driver.js:206`](https://github.com/appium/appium-inspector/blob/main/app/common/renderer/lib/appium/inspector-driver.js#L206), by checking the error code. If it's 500, you can propagate the error further, which should then be caught by `applyClientMethod` and `restartSession`. Implementing it this way would also trigger the session reload as soon as the element is selected, without needing to click Tap.

app/common/renderer/actions/SessionInspector.js
eglitise: Hmm, I do agree that it would be highly useful to have a dedicated method for a device health check. Sadly this isn't straightforward, since `getOrientation` is only used for mobile devices in native context.

Creating a custom method name is also not great, since it may become a real method in the future. I think a more universal solution would be what I outlined in another thread, where you can rely on an HTTP 500 response (possibly with additional error message text matching) to identify the device disconnection event.

app/common/renderer/actions/SessionInspector.js
eglitise: For full coverage I guess such error checks would also be needed in other `this.driver` methods that can be called from within `run()`. Its current code is far from clean, so if it helps, a refactor could also be useful.

app/common/renderer/lib/appium/inspector-driver.js
eglitise: I've described this in the other threads, but you would probably need to match against the HTTP 500 error code specifically. These methods may throw an error that is not related to the device being disconnected, such as the OS/app classifying the screen as secure and preventing screen capture. Having auto session restart on should not prevent these errors from being properly shown.

app/common/renderer/actions/SessionInspector.js
mohangowdak: @eglitise, Thanks a lot for reproducing.
I did try to catch the 500 error, but since the API calls happen inside the WebDriver library, we don’t actually get access to the status code of those calls—we only receive the error message. The challenge is that relying purely on the WebDriver error message doesn’t seem very reliable or future-proof.

eglitise: Looks like you are correct. In that case a more reliable solution could be to check the [W3C error code](https://w3c.github.io/webdriver/#errors), in the `name` field of the error. I believe the HTTP 500 errors stemming from device disconnection should have the code `unknown error`, so you could use that. For non-fatal errors (like HTTP 404 element not found), the `name` field seems to be just `Error`.

mohangowdak: Thanks so much for the suggestion. It worked exactly as you suggested, and I’ve implemented it in the recent push. Please let me know if you see any changes or improvements that are needed.

app/common/renderer/actions/SessionInspector.js
mohangowdak: Thanks for the suggestion. I’ve implemented it in the recent push. Please let me know if you see any changes or improvements that are needed.

app/common/renderer/lib/appium/inspector-driver.js
mohangowdak: Thanks for the suggestion. I’ve implemented it in the recent push. Please let me know if you see any changes or improvements that are needed.

app/common/renderer/actions/SessionInspector.js
```diff
@@ -328,12 +330,47 @@ export function applyClientMethod(params) {
       return commandRes;
     } catch (error) {
       log.error(error);
+      if (getState().inspector.autoSessionRestart) {
```
eglitise: What do you think about moving this try/catch code to `callClientMethod`? That way it also would be able to catch failed `findElement` attempts. If the error is different, it can just be re-raised and it would be caught by its calling function (e.g. `applyClientMethod`).

app/common/renderer/components/SessionInspector/Header/HeaderButtons.jsx
```diff
@@ -225,12 +228,24 @@ const HeaderButtons = (props) => {
     </Tooltip>
   );
 
+  const sessionReloadButton = (
+    <Tooltip title={t('ToggleRestartSession')}>
+      <Button
+        id={autoSessionRestart ? 'btnDisableRestartSession' : 'btnEnableRestartSession'}
+        icon={<RetweetOutlined />}
```
eglitise: I'm not sure if this is icon works that well. I do agree that it's not easy to find a good match, though. How about [`BiRecycle`](https://react-icons.github.io/react-icons/search/#q=birecycle)?

app/common/renderer/lib/appium/session-driver.js
```diff
@@ -46,11 +46,18 @@ export default class WDSessionDriver {
     if (!(commandName in this.client)) {
       throw new Error(`Command '${commandName}' is not available on this client`);
     }
-    const res = await this.client[commandName](...args);
-    if (res && res.error) {
-      throw new Error(res.message ? res.message : res.error);
+    try {
```
eglitise: I'm not sure if this code would even raise an error. It's also very central to the entire Inspector functionality, so out of caution I would probably suggest not making any changes here.

app/common/renderer/lib/appium/inspector-driver.js
```diff
@@ -252,6 +258,9 @@ export default class InspectorDriver {
         }
       }
     } catch (e) {
+      if (e.name === 'unknown error') {
+        throw {...e, customError: 'Session Expired'};
```
eglitise: It's also worth defining `Session Expired` as a constant since it's being used in multiple locations.

app/common/renderer/actions/SessionInspector.js
mohangowdak: Thanks a lot for the suggestion. I tried that approach, and the output remains the same. I agree we can move the try/catch block as you mentioned, but even after moving it, 
- we still won’t be able to catch the error for findElement attempts, since the exception is currently handled with an empty {} object in inspector-driver.js:217.
- Regarding interactions like tap or clear, we’re getting undefined for await this.driver[methodName] in inspector-driver.js:131 and 143, which leads to the "Cannot read properties of undefined (reading 'apply')" error. Because of that, we’re receiving undefined instead of a proper error object, so I’m unable to implement the W3C error code handling for 400 and 500 cases.

Could you please share your thoughts on whether it still makes sense to move the catch block in this situation?

app/common/renderer/components/SessionInspector/Header/HeaderButtons.jsx
mohangowdak: Sure, I will update it like this in the next push. Thanks!

app/common/renderer/lib/appium/session-driver.js
mohangowdak: Thanks for the clarification. Sure, if we remove the try/catch block, then I’ll need to move the handling logic to either the executeMethod() or run() function in inspector-driver.js to properly manage the executeScript call.
Please let me know if that approach sounds reasonable.

mohangowdak: Thanks for the clarification. Now I've updated the code

app/common/renderer/actions/SessionInspector.js
eglitise: > we still won’t be able to catch the error for findElement attempts, since the exception is currently handled with an empty {} object in inspector-driver.js:217.

I suggested this earlier, but you can probably just reuse the same error handler that you've already added for `getWindowUpdate`/`getSourceUpdate`/`getScreenshotUpdate`. It may also be relevant for `fetchElements`.

> we’re getting undefined for await this.driver[methodName]

I think this case won't be relevant after implementing the above error for `fetchElement` (which should be triggered instead), but if not, I tried something like this and it worked:
```js
// define this outside the InspectorDriver class
class WebdriverUnknownError extends Error {
  constructor(message) {
    super(message);
    this.name = UNKNOWN_ERROR;
  }
}

async executeMethod(...) {
      ...
      } else if (methodName !== 'getPageSource') {
        if (typeof this.driver[methodName] !== 'function') {
          throw new WebdriverUnknownError('Session Expired');
        }
        res = await this.driver[methodName].apply(this.driver, args);
      }
      ...
```

mohangowdak: @eglitise, Thanks so much for your suggestions. As you mentioned, the code is working fine, and I’ve updated it according to your instructions. Now the session restarts for all interactions if the device gets disconnected.

app/common/renderer/actions/SessionInspector.js
```diff
@@ -878,43 +905,55 @@ export function keepSessionAlive() {
...
   return async (dispatch, getState) => {
-    const {driver, appMode, isUsingMjpegMode, isSourceRefreshOn} = getState().inspector;
-    const {methodName, ignoreResult = true} = params;
-    params.appMode = appMode;
-
-    // don't retrieve screenshot if we're already using the mjpeg stream
-    if (isUsingMjpegMode) {
-      params.skipScreenshot = true;
-    }
+    try {
```
eglitise: For clarity you can probably move the `try` to right before the line `const action = keepSessionAlive();`

app/common/renderer/lib/appium/inspector-driver.js
```diff
@@ -162,7 +190,15 @@ export default class InspectorDriver {
...
     const start = Date.now();
-    const els = await this.driver.findElements(strategy, selector);
+    let els = null;
+    try {
+      els = await this.driver.findElements(strategy, selector);
+    } catch (err) {
+      if (err.name === UNKNOWN_ERROR) {
+        throw err;
+      }
+      return {};
```
eglitise: This line should probably be removed, because it will silently ignore all errors that aren't `unknown error`, which wasn't the case before

eglitise: Or rather, since there was no `try` block here before, all errors were already propagated to the calling method, so there's no need for any changes

mohangowdak: Thanks for pointing that out. Sure, I’ll remove the try/catch block as suggested.

app/common/renderer/actions/SessionInspector.js
mohangowdak: Thanks for the suggestion

app/common/public/locales/en/translation.json
```diff
@@ -314,5 +314,7 @@
   "Fireflink DeviceFarm Domain": "Fireflink DeviceFarm Domain",
   "Fireflink DeviceFarm Access Key": "Fireflink DeviceFarm Access Key",
   "Fireflink DeviceFarm License ID": "Fireflink DeviceFarm License ID",
-  "Fireflink DeviceFarm Project Name": "Fireflink DeviceFarm Project Name"
+  "Fireflink DeviceFarm Project Name": "Fireflink DeviceFarm Project Name",
+  "RestartSessionMessage": "Device Connection Lost - Restarting the Session...",
```
mykola-mokhnach: nit: Consider using Unicode ellipsis instead

app/common/renderer/constants/session-inspector.js
```diff
@@ -9,6 +9,9 @@ export const APP_MODE = {
 
 export const NATIVE_APP = 'NATIVE_APP';
 
+export const UNKNOWN_ERROR = 'unknown error';
+export const SESSION_EXPIRED = 'Session Expired';
```
mykola-mokhnach, line 13: should this sentence be localized?

app/common/renderer/actions/SessionInspector.js
```diff
@@ -272,65 +274,90 @@ export function applyClientMethod(params) {
...
       showError(error, {methodName: params.methodName, secs: 10});
-      dispatch({type: METHOD_CALL_DONE});
+      return dispatch({type: METHOD_CALL_DONE});
     }
+    showError(error, {methodName: params.methodName, secs: 3});
+    notification.info({
+      message: i18n.t('RestartSessionMessage'),
+      duration: 3,
+    });
+    const quitSes = quitSession('Window closed');
```
mykola-mokhnach, line 344: should this string be localized?

app/common/renderer/constants/session-inspector.js
mohangowdak: @mykola-mokhnach, Thanks for checking. I don’t think this needs to be localized, since it’s only used internally for logic to retrieve the W3C error code. These messages won’t be visible in the UI.

app/common/renderer/actions/SessionInspector.js
mohangowdak: @mykola-mokhnach, Thanks for checking. I don’t think this needs to be localized, since it’s only used internally to quit the session and isn’t visible in the UI. I’ve referred to the implementation in [SessionBuilder.js:940](https://github.com/appium/appium-inspector/blob/main/app/common/renderer/actions/SessionBuilder.js#L940) for consistency.

app/common/public/locales/en/translation.json
mohangowdak: @mykola-mokhnach, Thanks for the note. Sure, I’ve update with unicode.

PR #2393, opened 2025-10-31
refactor: Make locator generation utils easier to read and understand

mykola-mokhnach:
I did this refactoring with my friend Cursor, so please check carefully if the evil machine did not miss anything important. The tests still pass and I did not find anything potentially suspicious there, nevertheless an extra pair of eyes is always appreciated.

### DIFF
--- a/app/common/renderer/utils/locator-generation.js
+++ b/app/common/renderer/utils/locator-generation.js
@@ -4,36 +4,9 @@ import XPath, {select as xpathSelect} from 'xpath';
 import {log} from './logger.js';
 import {childNodesOf, domToXML, findDOMNodeByPath, xmlToDOM} from './source-parsing.js';
 
-// Attributes on nodes that are likely to be unique to the node so we should consider first when
-// suggesting xpath locators. These are considered IN ORDER.
-const UNIQUE_XPATH_ATTRIBUTES = ['name', 'content-desc', 'id', 'resource-id', 'accessibility-id'];
-
-// Attributes that we should recommend as a fallback but ideally only in conjunction with other
-// attributes
-const MAYBE_UNIQUE_XPATH_ATTRIBUTES = ['label', 'text', 'value'];
-
-const CHECKED_CLASS_CHAIN_ATTRIBUTES = ['name', 'label', 'value'];
-
-const CHECKED_PREDICATE_ATTRIBUTES = ['name', 'label', 'value', 'type'];
-
-// Map of element attributes to their UiAutomator syntax, ordered by (likely) decreasing uniqueness
-const CHECKED_UIAUTOMATOR_ATTRIBUTES = [
-  ['resource-id', 'resourceId'],
-  ['text', 'text'],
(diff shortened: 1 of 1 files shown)


app/common/renderer/utils/locator-generation.js
```diff
@@ -89,22 +52,25 @@ export function getComplexSuggestedLocators(path, sourceDoc, isNative, automatio
...
           domNode,
           path,
-        );
+        ).generate();
         break;
       }
     }
   }
-  complexLocators.xpath = getOptimalXPath(sourceDoc, domNode);
+  complexLocators.xpath = new XPathGenerator(sourceDoc, domNode).generate();
```
eglitise: These changes aren't needed since they just duplicate the still existing `getOptimalX` methods

app/common/renderer/utils/locator-generation.js
```diff
@@ -136,403 +102,795 @@ export function getSuggestedLocators(selectedElement, sourceXML, isNative, autom
...
+  }
+
+  /**
+   * Get suggested selectors for simple locator strategies (which match a specific attribute)
+   *
+   * @returns {Record<string, string>} mapping of strategies to selectors
+   */
+  generate() {
+    const res = {};
+    for (let [strategyAlias, strategy] of SimpleLocatorGenerator.STRATEGY_MAPPINGS) {
```
eglitise: The alias and strategy can probably be changed to `const`

app/common/renderer/utils/locator-generation.js
```diff
@@ -136,403 +102,795 @@ export function getSuggestedLocators(selectedElement, sourceXML, isNative, autom
...
       try {
-        othersWithAttr = xpathSelect(xpath, doc);
+        othersWithAttr = xpathSelect(xpath, this._doc);
       } catch {
         continue;
       }
 
-      // If the attribute isn't actually unique, get its index too
+      // Build the class chain from this attribute
+      let classChain = this._buildClassChainFromAttribute(attrName, attrValue);
```
eglitise, line 559: Interesting that Cursor moved this line after the XPath parsing check, while leaving the equivalent UiAutomator selector line (814) before the check. Not an issue, just an observation.

app/common/renderer/utils/locator-generation.js
mykola-mokhnach: changed to a reducer

eglitise (approved):
Cursor's approach of extracting simple string builders that get called only once into their own methods (e.g. `_buildUiSelectorWithInstance`) seems a bit odd to me, and there's some new minor duplication here and there (e.g. `const pathArray = this._path.split('.');` can get called twice in UiAutomator locator generation), but other than that it look largely correct to me 👍

app/common/renderer/utils/locator-generation.js
mykola-mokhnach: updated to use public APIs

PR #2414, opened 2025-11-11
feat: retrieve commands and execute methods from active extensions

eglitise:
This is a major enhancement to the Commands tab, adding the ability to retrieve the active driver's supported commands directly from the driver, and generate the visible command buttons from this response. Note that since the Inspector relies on WDIO for running any command, the visible commands are filtered only to those supported by WDIO.

Many other related enhancements are also included:
* Retrieve and run the active driver's supported execute methods
* Retrieve and run the commands and execute methods added by any active plugins (the WDIO support limitation still applies for commands)
* Filter the commands/methods using a search bar
* Show command/method specific information (deprecated status, info text, required status for any parameters)
* Keep the existing predefined commands list if no commands/methods can be retrieved (e.g. Appium <2.16)
* Add unit tests and update documentation

Functionality was tested on Appium 3.1.0, 2.19.0 and 2.15.0, with UiAutomator2 and XCUITest drivers.

Here's a brief demo:

https://github.com/user-attachments/assets/6d7a4ebe-f255-4379-96e0-f9e9de8654b0


app/common/renderer/utils/commands-tab.js
```diff
@@ -0,0 +1,229 @@
...
+ * Try to detect if the input value should be a boolean/number/array/object,
+ * and if so, convert it to that
+ *
+ * @param {string} value the value of a command parameter
+ * @returns the value converted to the type it matches best
+ */
+export function adjustParamValueType(value) {
+  if (value === '') {
+    return null;
+  } else if (Number(value).toString() === value) {
```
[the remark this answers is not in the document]
eglitise: Values like "01" are expected to be retained as strings, and the logic matches this.

app/common/renderer/utils/commands-tab.js
```diff
@@ -0,0 +1,229 @@
+import _ from 'lodash';
+
+import {APPIUM_TO_WD_COMMANDS, COMMANDS_WITH_MISMATCHED_PARAMS} from '../constants/commands.js';
+
+/**
+ * Try to detect if the input value should be a boolean/number/array/object,
+ * and if so, convert it to that
+ *
+ * @param {string} value the value of a command parameter
+ * @returns the value converted to the type it matches best
```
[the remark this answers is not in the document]
eglitise: VSCode shows the `any` return value regardless of whether these types are specified or not

app/common/public/locales/en/translation.json
```diff
@@ -320,5 +320,9 @@
   "toggleTableFormatting": "Toggle Table Formatting",
   "copyResultToClipboard": "Copy Result to Clipboard",
   "Property": "Property",
-  "noCapsFound": "No session capabilities found. Please use the Capability Builder to edit, add, and save your capability set. Refer to the tutorial at {{url}}"
+  "noCapsFound": "No session capabilities found. Please use the Capability Builder to edit, add, and save your capability set. Refer to the tutorial at {{url}}",
+  "executeMethods": "Execute Methods",
+  "dynamicCommandsDescription": "Run commands supported by the currently active driver and plugin(s). Note that certain commands may only work in certain contexts. Any commands not supported by WebdriverIO are not shown in this list.",
```
mykola-mokhnach: Run commands supported by the currently active driver and plugin(s). Note that some commands may only work in specific contexts. Any commands not supported by WebdriverIO are not included in this list.

app/common/renderer/actions/SessionInspector.js
```diff
@@ -824,21 +816,23 @@ export function selectInspectorTab(interaction) {
...
-}
+export function getSupportedSessionMethods() {
+  return async (_dispatch, getState) => {
+    async function safelyCallCommand(methodName) {
+      try {
+        const action = executeDriverCommand({methodName});
+        const {commandRes} = await action(getState);
+        return commandRes;
+      } catch {
+        return [];
```
mykola-mokhnach, line 827: do we want to show any error message here?

app/common/renderer/components/SessionInspector/CommandsTab/Commands.jsx
```diff
@@ -1,126 +1,120 @@
...
+      prepareAndRunCommand(commandDetails);
     }
+  };
 
-    // Special case for 'setGeoLocation'
-    if (commandName === 'setGeoLocation') {
-      copiedArgs = {latitude: args[0], longitude: args[1], altitude: args[2]};
+  const prepareCommand = (cmdName, cmdParams, isExecute) => {
+    const adjustedCmdName = isExecute ? 'executeScript' : cmdName;
+    let adjustedCmdParams = _.cloneDeep(curCommandParamVals.current).map((val) =>
```
mykola-mokhnach: could be simplified to `.map(adjustParamValueType)`

app/common/renderer/components/SessionInspector/CommandsTab/Commands.jsx
```diff
@@ -1,126 +1,120 @@
...
-        copiedArgs[1] = [];
-      } else {
-        copiedArgs[1] = parseJsonString(args[1]);
-        if (copiedArgs[1] === null) {
-          isJsonValid = false;
-        } else if (!_.isArray(copiedArgs[1])) {
-          copiedArgs[1] = [copiedArgs[1]];
-        }
+    // So we can have 5 cases for the argument: undefined, {}, [], {...}, [{...}]
+    if (adjustedCmdName === 'executeScript') {
```
mykola-mokhnach: should `executeScript` be a constant?

app/common/renderer/components/SessionInspector/CommandsTab/Commands.jsx
```diff
@@ -1,126 +1,120 @@
...
+      if (_.isEmpty(adjustedCmdParams[1])) {
+        adjustedCmdParams[1] = [];
+      } else if (typeof adjustedCmdParams[1] === 'object' && !_.isArray(adjustedCmdParams[1])) {
```
mykola-mokhnach: should we use _.isPlainObject?

app/common/renderer/components/SessionInspector/CommandsTab/Commands.jsx
```diff
@@ -1,126 +1,120 @@
...
-      }
-    }
+  const runCommand = async (methodName, args, skipRefresh) => {
+    const res = await applyClientMethod({
+      methodName,
+      args,
+      skipRefresh,
+    });
+    const formattedResult =
+      _.isObject(res) && _.isEmpty(res) ? null : JSON.stringify(res, null, '  ');
```
mykola-mokhnach: nit: JSON.stringify(res, null, '  ') -> JSON.stringify(res, null, 2)

app/common/renderer/utils/commands-tab.js
```diff
@@ -0,0 +1,229 @@
...
+/**
+ * Filter the array of method key-value pairs to only include methods matching the search query.
+ *
+ * @param {[string, Object][]} methodPairs array of [methodName, methodDetails] pairs
+ * @param {string} searchQuery user-provided search query
+ * @returns filtered array of [methodName, methodDetails] pairs
+ */
+export function filterMethodPairs(methodPairs, searchQuery) {
+  return _.filter(methodPairs, ([methodName]) => {
+    if (!searchQuery) {
```
mykola-mokhnach: I assume we can just skip filtering if searchQuery is falsy

app/common/renderer/utils/commands-tab.js
```diff
@@ -0,0 +1,229 @@
...
+      return true;
+    }
+    return methodName.toLowerCase().includes(searchQuery.toLowerCase());
```
mykola-mokhnach: I think it might be more useful to use [Levenshtein distance](https://github.com/gustf/js-levenshtein) here for comparison rather than strict inclusion

app/common/renderer/utils/commands-tab.js
```diff
@@ -0,0 +1,229 @@
...
+ * If this causes the parent object/array to become empty,
+ * this parent will be removed from its own parent as well.
+ *
+ * @param {*} value any value (object, array, primitive)
+ * @returns the value with empty entries removed
+ */
+export function deepFilterEmpty(value) {
+  if (_.isArray(value)) {
+    // Recurse into each array element, then remove empty entries
+    const mapped = value.map((v) => deepFilterEmpty(v));
```
mykola-mokhnach: may be simplified to

```
return value.map(deepFilterEmpty).filter(_.negate(isEmptyObject))
```

mykola-mokhnach: Just in case you are into functional programming, check the concept of lodash's flow: https://www.geeksforgeeks.org/javascript/lodash-_-flow-method/

app/common/renderer/utils/commands-tab.js
```diff
@@ -0,0 +1,229 @@
...
+    return mapped.filter((v) => !isEmptyObject(v));
+  }
+
+  if (_.isPlainObject(value)) {
+    // Recurse into object properties, then pick only non-empty values
+    const mapped = _.mapValues(value, (v) => deepFilterEmpty(v));
```
mykola-mokhnach: same as above

app/common/renderer/utils/commands-tab.js
```diff
@@ -0,0 +1,229 @@
...
+}
+
+/**
+ * Extract any parameter names (except sessionId) from a command path.
+ *
+ * @param {string} path command endpoint URL
+ * @returns array of parameter names
+ */
+export function extractParamsFromCommandPath(path) {
+  const paramNames = [];
```
mykola-mokhnach: this function may be replaced by single `reduce`

test/unit/utils-commands-tab.spec.js
```diff
@@ -0,0 +1,400 @@
...
+  deepFilterEmpty,
+  extractParamsFromCommandPath,
+  filterMethodPairs,
+  transformCommandsMap,
+  transformExecMethodsMap,
+} from '../../app/common/renderer/utils/commands-tab.js';
+
+describe('utils/commands-tab.js', function () {
+  describe('#adjustParamValueType', function () {
+    it('should detect the correct type for common examples', function () {
```
mykola-mokhnach: you may use the following approach to avoid unnecessary repetition in parametrized tests:

```
const cases = [
  ['test', 'test'],
  ['109', 109],
  ['true', true],
  ['false', false],
  ['null', null],
  ['[1, 2, 3]', [1, 2, 3]],
  ['{"a":1}', { a: 1 }],
];

cases.forEach(([input, expected]) => {
  it(`should detect correct type for input "${input}"`, function () {
    expect(adjustParamValueType(input)).to.eql(expected);
  });
});
```

app/common/renderer/actions/SessionInspector.js
eglitise: No, because those endpoints don't exist in Appium <2.16, so we would always have to show an error in that case. But since we use the static map fallback that still offers partial functionality, the error feels too disruptive, especially if the user is not in control of the Appium version.

app/common/renderer/utils/commands-tab.js
eglitise: I'm not really sure about its benefit. Unlike Appium's use of Levenshtein distance to print the closest matching script name, here the search results are shown in real-time and use partial matching, so it's simple enough to just adjust the query itself.

app/common/renderer/utils/commands-tab.js
eglitise: I checked the `reduce` documentation and concluded that `flatMap` is a better fit, since the array elements don't need to be combined

app/common/renderer/utils/commands-tab.js
eglitise: Thanks for this, I've refactored to this approach for better readability

app/common/renderer/components/SessionInspector/CommandsTab/Commands.jsx
```diff
@@ -45,10 +48,8 @@ const Commands = (props) => {
   };
 
   const prepareCommand = (cmdName, cmdParams, isExecute) => {
-    const adjustedCmdName = isExecute ? 'executeScript' : cmdName;
-    let adjustedCmdParams = _.cloneDeep(curCommandParamVals.current).map((val) =>
-      adjustParamValueType(val),
-    );
+    const adjustedCmdName = isExecute ? COMMAND_EXECUTE_SCRIPT : cmdName;
+    let adjustedCmdParams = curCommandParamVals.current.map((val) => adjustParamValueType(val));
```
mykola-mokhnach: could also be `.map(adjustParamValueType)`

app/common/renderer/utils/commands-tab.js
```diff
@@ -60,14 +60,18 @@ const isEmptyObject = (value) => _.isObjectLike(value) && _.isEmpty(value);
 export function deepFilterEmpty(value) {
   if (_.isArray(value)) {
     // Recurse into each array element, then remove empty entries
-    const mapped = value.map((v) => deepFilterEmpty(v));
-    return mapped.filter((v) => !isEmptyObject(v));
+    const recurse = (arr) => arr.map(deepFilterEmpty);
+    const clean = (arr) => arr.filter(_.negate(isEmptyObject));
+    const recurseAndClean = _.flow([recurse, clean]);
```
mykola-mokhnach, line 65: ❤️

app/common/renderer/utils/commands-tab.js
```diff
@@ -60,14 +60,18 @@ const isEmptyObject = (value) => _.isObjectLike(value) && _.isEmpty(value);
...
+    return recurseAndClean(value);
   }
 
   if (_.isPlainObject(value)) {
     // Recurse into object properties, then pick only non-empty values
-    const mapped = _.mapValues(value, (v) => deepFilterEmpty(v));
-    return _.pickBy(mapped, (v) => !isEmptyObject(v));
+    const recurse = (obj) => _.mapValues(obj, deepFilterEmpty);
+    const clean = (obj) => _.pickBy(obj, _.negate(isEmptyObject));
+    const recurseAndClean = _.flow([recurse, clean]);
```
mykola-mokhnach, line 73: maybe this part could be extracted to avoid duplication

eglitise: I think this is fine, since it's only 2 lines that are only duplicated once

PR #2457, opened 2025-12-01
fix: make caps editor behavior more deterministic

eglitise:
This is a slight improvement to handling changes to capabilities in a capability set, and synchronizing them between the capability builder and the JSON representation.
Currently a lot of the logic for finding the individual capability is based on its index in the capability array, but this is not reliable, since any capability can be added and removed, thereby changing the index. I have therefore added the `id` property, which is initialized with a random UUID (at first I used `Symbol` but React does not support symbols as element `key` values).

The PR also fixes a small issue where changing the capability type via editing the raw JSON did not immediately change the type in the capability builder.

### DIFF
--- a/app/common/renderer/actions/SessionBuilder.js
+++ b/app/common/renderer/actions/SessionBuilder.js
@@ -637,7 +637,7 @@ export function saveSessionAsFile() {
     const state = getState().builder;
     const sessionFileDetails = {
       version: APPIUM_SESSION_FILE_VERSION,
-      caps: state.caps.map((cap) => ({..._.omit(cap, 'id')})),
+      caps: state.caps.map((cap) => _.omit(cap, 'id')),
       server: state.server,
       serverType: state.serverType,
       visibleProviders: state.visibleProviders,
--- a/app/common/renderer/components/SessionBuilder/CapabilityBuilderTab/CapabilityEditor.jsx
+++ b/app/common/renderer/components/SessionBuilder/CapabilityBuilderTab/CapabilityEditor.jsx
@@ -116,7 +116,6 @@ const CapabilityEditor = (props) => {
                 <Form.Item>
                   <Select
                     disabled={isEditingDesiredCaps}
-                    defaultValue={cap.type}
                     value={cap.type}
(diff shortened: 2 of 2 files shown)


app/common/renderer/actions/SessionBuilder.js
```diff
@@ -809,36 +809,34 @@ export function abortDesiredCapsEditor() {
...
+      const currentCapsObj = _.fromPairs(
+        currentCapsArray.map((cap) => [cap.name, _.omit(cap, 'name')]),
+      );
+
+      // Translate the raw caps JSON to array format
+      const newCapsArray = Object.entries(rawCapsObj).map(([name, value]) => ({
+        id: crypto.randomUUID(),
+        enabled: true,
+        ...currentCapsObj[name], // overrides id and enabled, if present
+        type: typeof value === 'string' ? 'text' : typeof value,
```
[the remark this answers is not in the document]
eglitise: This is fine since the file type has not been supported for several versions

jlipps (approved):
👍  good idea!

app/common/renderer/actions/SessionBuilder.js
```diff
@@ -809,36 +809,34 @@ export function abortDesiredCapsEditor() {
...
+      // Since the user may change the order of existing caps in the raw array,
+      // we cannot use the element index - however, since the raw cap names are unique
+      // (due to being JSON keys), use that as the identifier.
+      // But if the user has changed the name of an existing cap, treat it as a new cap.
+
+      // First, convert the current caps array to an object, in order to use name indexing.
+      // This also removes any entries with duplicate names (the capability builder allows duplicates),
+      // which is fine, since JSON only allows the latest entry anyway.
+      const currentCapsObj = _.fromPairs(
+        currentCapsArray.map((cap) => [cap.name, _.omit(cap, 'name')]),
```
mykola-mokhnach, line 829: you may also use _.mapValues

eglitise: mapValues accepts an object and returns an object with the same keys. Not sure how that would work here.

PR #2481, opened 2025-12-08
chore: replace axios with ky

eglitise:
This replaces the `axios` dependency with the smaller `fetch`-based `ky`. The goal here is to slightly reduce the bundle size and use an API that is closer to the native `fetch()` API.
I've tested session retrieval in the Attach to Session tab and confirmed that it works as expected. Hopefully the same also applies to the Crowdin scripts.

### DIFF
--- a/scripts/crowdin-common.mjs
+++ b/scripts/crowdin-common.mjs
@@ -23,8 +23,10 @@ export async function performApiRequest(suffix = '', opts = {}) {
     ? `${API_ROOT}/projects/${PROJECT_ID}${suffix}`
     : `${API_ROOT}${suffix}`;
   log.debug(`Sending ${method} request to ${url}`);
-  if (_.isString(payload)) {
-    log.debug(`Request payload: ${payload}`);
+  let formattedPayload = payload;
+  if (_.isPlainObject(payload)) {
+    formattedPayload = JSON.stringify(payload);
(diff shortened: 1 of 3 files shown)


scripts/crowdin-common.mjs
```diff
@@ -23,20 +23,17 @@ export async function performApiRequest(suffix = '', opts = {}) {
     ? `${API_ROOT}/projects/${PROJECT_ID}${suffix}`
     : `${API_ROOT}${suffix}`;
   log.debug(`Sending ${method} request to ${url}`);
-  if (_.isPlainObject(payload)) {
-    log.debug(`Request payload: ${JSON.stringify(payload)}`);
+  if (_.isString(payload)) {
```
mykola-mokhnach: I'm not sure this is going to work

eglitise: I've adjusted the code to call JSON.stringify inside this method. This should allow for a safer way to distinguish object payloads from ReadStream payloads.

mykola-mokhnach (approved):

eglitise:
Yes, I checked the `package-lock.json` file, and all remaining instances of `axios` are now dev-only (specifically, from `appium`, `@appium/base-driver`, and `@appium/support`).

PR #2499, opened 2025-12-12
fix: remove extra quotes from string command results

unickq:
## Description

Minor improvement to how command results are formatted in the `CommandsTab` component. The change ensures that if the result is already a string, it is displayed as-is, rather than being stringified again.

#### from
image

#### to
image

just tired of removing quotes from base64 images 🥲

### DIFF
--- a/app/common/renderer/components/SessionInspector/CommandsTab/CommandResultModal.jsx
+++ b/app/common/renderer/components/SessionInspector/CommandsTab/CommandResultModal.jsx
@@ -208,7 +208,7 @@ const CommandResultModalFooter = ({
   </Row>
 );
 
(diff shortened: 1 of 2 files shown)


eglitise:
It's not that simple - if we're specifically referring to the Commands tab, then that list of supported commands is provided by the driver itself, and the Inspector merely formats and presents it. That is to say, it's not possible for the Inspector to know the type of the response in advance.
As for an example (not a command though), I believe that one or more of the UiAutomator2 element attributes of `checked`, `enabled`, `selected` or `displayed` returned the boolean value as a string - at least when using the Ruby client.

unickq:
> Would that work for your case?

Here it is 👍 

imageimage

PR #2515, opened 2025-12-18
feat: add 2 suggested locator strategies for web context

eglitise:
As another enhancement for usage in web context, this PR updates the list of suggested element locators:
* Do not show any strategies that do not exist in web context
* Add 2 new strategies: css (currently only based on the `id` property) and tag name
* Update tests and add many new ones

It is worth noting that the element tag name and link text are not yet shown in the Selected Element panel, as they are not part of the element's attributes. This will require further changes.

Quick demo (note: link text selector was removed as it required a lot more changes than anticipated, and will be explored again in the future):

https://github.com/user-attachments/assets/42276aff-6dfb-4225-9ae1-6948bf9ad52d

### DIFF
--- a/app/common/renderer/utils/locator-generation.js
+++ b/app/common/renderer/utils/locator-generation.js
@@ -1,6 +1,6 @@
 import cssEscape from 'css.escape';
 import _ from 'lodash';
-import {select as xpathSelect} from 'xpath';
+import XPath, {select as xpathSelect} from 'xpath';
 
 import {LOCATOR_STRATEGIES as STRATS} from '../constants/session-inspector.js';
 import {log} from './logger.js';


app/common/renderer/utils/locator-generation.js
```diff
@@ -9,31 +10,65 @@ import {childNodesOf, domToXML, findDOMNodeByPath, xmlToDOM} from './source-pars
...
+/**
+ * Check whether the provided element link text is unique in the source
+ *
+ * @param {string} textContent
+ * @param {Document} sourceDoc
+ * @returns {boolean}
+ */
+export function isLinkTextUnique(textContent, sourceDoc) {
+  return isXpathUnique(`//a[text()="${textContent.replace(/"/g, '')}"]`, sourceDoc);
+}
```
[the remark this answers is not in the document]
mykola-mokhnach: I support AI consideration here. Moreover the result of this method would probably be incorrect if quotes are part of the original content. They should be properly escaped instead of being replaced with empty strings

app/common/renderer/utils/locator-generation.js
```diff
@@ -836,49 +872,71 @@ class UiAutomatorGenerator extends LocatorGeneratorBase {
...
+   */
+  generateWebSelectors() {
+    const webStrategyMap = {};
+    // id (css)
+    const idValue = this._attributes?.id;
+    if (idValue && areAttrAndValueUnique('id', idValue, this._doc)) {
+      webStrategyMap[STRATS.CSS] = `#${idValue}`;
+    }
+    // link text
+    if (this._tag === 'a' && this._text && isLinkTextUnique(this._text, this._doc)) {
```
mykola-mokhnach: must link text always be an exact match?

eglitise: Yes - partial matches are only supported for the `partial link text` strategy, which can still be added in the future, but would require additional calculations

app/common/renderer/utils/locator-generation.js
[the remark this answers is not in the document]
eglitise: I've now added sanitization for attribute values and the assembled CSS id selector.

app/common/renderer/utils/locator-generation.js
```diff
@@ -76,19 +118,20 @@ export function getComplexSuggestedLocators(path, sourceDoc, isNative, automatio
 /**
  * Get suggested selectors for all locator strategies
  *
- * @param {string} selectedElement element node in JSON format
+ * @param {object} selectedElement element node in JSON format
```
mykola-mokhnach, line 109: does node has more specific type?

app/common/renderer/utils/locator-generation.js
```diff
@@ -149,6 +192,41 @@ export function getOptimalUiAutomatorSelector(doc, domNode, path) {
...
 
+/**
+ * Trim whitespace from a string value,
+ * otherwise return an empty string
+ *
+ * @param {*} value input value
+ * @returns {string} trimmed string
+ */
+function toTrimmedString(value) {
+  return typeof value === 'string' ? value.trim() : '';
```
mykola-mokhnach, line 190: should we also consider numbers?

app/common/renderer/utils/locator-generation.js
```diff
@@ -149,6 +192,41 @@ export function getOptimalUiAutomatorSelector(doc, domNode, path) {
...
+
+/**
+ * Convenience check for whether the document node exists
+ *
+ * @param {Document | undefined} node document node
+ * @returns {boolean}
+ */
+function doesDocumentExist(node) {
+  // If no node provided, assume the xpath is unique
+  return !(!node || _.isEmpty(node));
```
mykola-mokhnach: I assume `return Boolean(node) && !_.isEmpty(node)` would be more readable

app/common/renderer/utils/locator-generation.js
eglitise: Unfortunately it's not typed. The object itself is created in `xmlToJSON`.

app/common/renderer/utils/locator-generation.js
eglitise: I don't think so. Tags and attribute names are always treated as strings (as they must contain at least one non-numerical symbol), text is also a string, and attribute values must also be strings ([link to spec](https://www.w3.org/TR/2008/REC-xml-20081126/#NT-AttValue)).

app/common/renderer/utils/locator-generation.js
```diff
@@ -835,49 +900,66 @@ class UiAutomatorGenerator extends LocatorGeneratorBase {
...
+    const webStrategyMap = {};
+    // id (css)
+    const idValue = this._attributes?.id;
+    if (idValue && areAttrAndValueUnique('id', idValue, this._doc)) {
+      webStrategyMap[STRATS.CSS] = `#${cssEscape(idValue)}`;
+    }
+    // tag name
+    if (isTagUnique(this._tag, this._doc)) {
+      webStrategyMap[STRATS.TAG_NAME] = this._tag;
+    }
```
KazuCocoa, line 960: I'm not familiar with how normally used tag name over css, but then here also could map with `css` but as another example with tag name in css selector...?

So,

Screenshot 2025-12-23 at 12 00 32 AM

will have:

- css selector with id
- css selector with tag name
- xpath

Just wondered using only css (and xpath) might be helpful for web use

KazuCocoa (approved):
(otherwise lg)

app/common/renderer/utils/locator-generation.js
eglitise: Tag name is still a standard WebDriver locator strategy, so I don't see a problem in having it here.
I agree with having more CSS options though - class, test id-style properties, accessibility-related properties, or any other unique property can absolutely be added in the future. This was merely the simplest option.

PR #2535, opened 2026-01-07
feat: update .appiumsession file format

eglitise:
This is an update to the format of the exported `.appiumsession` file:
* Add the `name` property. In the future this will allow importing multiple sessions at once, similarly to gestures.
* Remove the `serverType` and `visibleProviders` properties, as these can be assembled from the `server` property
* Only keep the current server details in the `server` property, plus the `advanced` property
* Update the `version` property to `2.0`
* Add tests for parsing the updated file format

Also, the name of the exported file is now based on the saved session name, instead of the server type, which should allow to distinguish them more easily.

Important to note that importing of version `1.0` session files is still supported. However, older Inspector version will not be able to import version `2.0` session files.

### DIFF
--- a/app/common/renderer/utils/sessionfile-parsing.js
+++ b/app/common/renderer/utils/sessionfile-parsing.js
@@ -171,6 +171,9 @@ function isSessionServerValid(sessionJSON) {
       return logValidationError(`unsupported server type '${key}'`);
     }
     if (key !== SERVER_TYPES.ADVANCED) {
+      if (!_.isPlainObject(sessionJSON.server[key])) {
+        return logValidationError(`'${key}' server property must be an object`);
+      }
       continue;
     }
     for (const [advKey, advValue] of Object.entries(sessionJSON.server.advanced)) {
--- a/test/unit/utils-sessionfile-parsing.spec.js
+++ b/test/unit/utils-sessionfile-parsing.spec.js
@@ -100,8 +100,9 @@ describe('utils/sessionfile-parsing.js', function () {
       ['is missing', {}],
       ['is a number', {server: 123}],
       ['is an empty object', {server: {}}],
-      ['has invalid structure', {server: {invalid: {}, advanced: {}}}],
-      ['has remote and local only', {server: {remote: {}, local: {}}}],
+      ['has invalid property names', {server: {invalid: {}, advanced: {}}}],
+      ['has non-object properties', {server: {remote: 123, advanced: {}}}],
+      ['does not have the advanced property', {server: {remote: {}, local: {}}}],
     ])('should return null if server %s', (_desc, newServerProp) => {
       const session = {...versionProp, ...nameProp, ...capsProp, ...newServerProp};
       expect(validateSessionJSON(session)).toBeNull();


app/common/renderer/actions/SessionBuilder.js
```diff
@@ -624,6 +627,11 @@ export function setStateFromSessionFile(sessionFileString) {
       });
       return;
     }
+    if (sessionJSON.version === '2.0') {
+      sessionJSON.serverType = Object.keys(sessionJSON.server)[0];
+      sessionJSON.visibleProviders =
+        sessionJSON.serverType !== SERVER_TYPES.REMOTE ? [sessionJSON.serverType] : [];
```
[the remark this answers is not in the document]
eglitise: This is expected for now, since the file contents are loaded into the capability builder, instead of directly saved as a capability set. A future PR will change this.

app/common/renderer/actions/SessionBuilder.js
```diff
@@ -624,6 +627,11 @@ export function setStateFromSessionFile(sessionFileString) {
       });
       return;
     }
+    if (sessionJSON.version === '2.0') {
```
mykola-mokhnach: I am not convinced this approach is flexible enough for future scenarious if we decide to add further improvements to the config and bump its version.

Consider extracting the migration logic to a separate module and have proper migration paths implementations for each version transition. See https://github.com/appium/appium/blob/master/packages/appium/lib/extension/manifest-migrations.js for more details

app/common/renderer/utils/file-handling.js
```diff
@@ -35,31 +35,70 @@ export async function readTextFromUploadedFiles(fileList) {
...
-            return null;
-          }
-        }
-        if (!_.values(CAPABILITY_TYPES).includes(cap.type)) {
-          return null;
-        }
-      }
     }
-    return sessionJSON;
+    if (sessionJSON.version === '1.0') {
```
mykola-mokhnach: I assume we should always migrate earlier versions to the most recent one and only have a parser for the most recent config version

app/common/renderer/reducers/SessionBuilder.js
```diff
@@ -397,12 +397,17 @@ export default function builder(state = INITIAL_STATE, action) {
           enabled: true,
           ...cap,
         })),
+        // reset capsName since we now have an unnamed session
+        capsName: null,
```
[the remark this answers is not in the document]
eglitise: The name property will be applied in a future PR. Currently the behavior remains as-is, by opening an unnamed session in the session builder.

app/common/renderer/actions/SessionBuilder.js
```diff
@@ -635,17 +643,22 @@ export function setStateFromSessionFile(sessionFileString) {
...
       caps: state.caps.map((cap) => _.omit(cap, 'id')),
-      server: state.server,
-      serverType: state.serverType,
-      visibleProviders: state.visibleProviders,
     };
     const href = `data:text/json;charset=utf-8,${encodeURIComponent(
       JSON.stringify(sessionFileDetails, null, 2),
     )}`;
-    const fileName = `${state.serverType}.appiumsession`;
+    const escapedName = sessionName.replace(/[\s\\/:*?"<>|]/g, '_');
```
mykola-mokhnach: It might not be a very straightforward task to properly escape a string to make it a valid filename, especially cross platform. Consider using established utilities for such purpose instead of writing custom code, for example https://www.npmjs.com/package/sanitize-filename

app/common/renderer/utils/file-handling.js
```diff
@@ -35,31 +41,172 @@ export async function readTextFromUploadedFiles(fileList) {
...
-      if (!(sessionProp in sessionJSON)) {
-        return null;
+/**
+ * Upgrades a session file from 1.0 to v2.
+ * The file may or may not be a fully valid session file.
+ *
+ * @param {object} sessionJSON session file JSON with version 1.0
+ * @returns migrated JSON, or null if any validations failed
+ */
+function migrateSessionJsonToV2(sessionJSON) {
```
mykola-mokhnach: I think utility functions related to config migrations should be extracted to a separate module/class, this is not generic file handling

app/common/renderer/utils/file-handling.js
```diff
@@ -35,31 +41,172 @@ export async function readTextFromUploadedFiles(fileList) {
...
+  // Add name (placeholder)
+  sessionJSON.name = DEFAULT_SESSION_NAME;
+  // Bump version
+  sessionJSON.version = SESSION_FILE_VERSIONS.V2;
+  // Filter server to only the value that matches serverType (plus advanced),
+  // creating it if it does not exist
+  if (!('serverType' in sessionJSON)) {
```
mykola-mokhnach: I'm not quite sure about this condition. We should either completely rely on the version field or check/update all fields/being paranoidal

app/common/renderer/utils/file-handling.js
```diff
@@ -35,31 +41,172 @@ export async function readTextFromUploadedFiles(fileList) {
...
+ * @param {string} sessionFileString session file contents
+ * @returns session file in JSON format, or null if any validations failed
+ */
+export function parseSessionFileContents(sessionFileString) {
+  try {
+    const sessionJSON = JSON.parse(sessionFileString);
+    const updatedSessionJSON = migrateSessionJSON(sessionJSON);
+    return validateSessionJSON(updatedSessionJSON);
   } catch {
     return null;
```
mykola-mokhnach: not sure if we really want to mute possible exceptions here. It would make debugging much harder

app/common/renderer/utils/file-handling.js
eglitise: This is primarily to match the existing validation for v1 sessionfile. Without this field, we cannot identify the active server type, preventing migration.

app/common/renderer/actions/SessionBuilder.js
eglitise: Added

app/common/renderer/utils/file-handling.js
eglitise: Moved to a dedicated module

app/common/renderer/utils/file-handling.js
eglitise: Added console logs for all possible issues. I don't think it's worth showing them in the UI, since these files are not intended to be manually edited.

app/common/renderer/actions/SessionBuilder.js
```diff
@@ -635,17 +645,22 @@ export function setStateFromSessionFile(sessionFileString) {
...
       caps: state.caps.map((cap) => _.omit(cap, 'id')),
-      server: state.server,
-      serverType: state.serverType,
-      visibleProviders: state.visibleProviders,
     };
     const href = `data:text/json;charset=utf-8,${encodeURIComponent(
       JSON.stringify(sessionFileDetails, null, 2),
     )}`;
-    const fileName = `${state.serverType}.appiumsession`;
+    const escapedName = sanitize(sessionName, {replacement: '_'});
```
mykola-mokhnach, line 662: should we be explicit about the target platform (e.g. win vs linux)?

app/common/renderer/utils/sessionfile-parsing.js
```diff
@@ -0,0 +1,203 @@
...
+}
+
+/**
+ * Migrates a session file to the latest version, if required.
+ * The file may or may not be a fully valid session file.
+ *
+ * @param {object} sessionJSON session file contents in JSON
+ * @returns migrated JSON, or null if any validations failed
+ */
+export function migrateSessionJSON(sessionJSON) {
```
mykola-mokhnach: nit: I usually put exported/public entities first so they are easier to locate

app/common/renderer/utils/sessionfile-parsing.js
```diff
@@ -0,0 +1,203 @@
...
+ */
+function isSessionServerValid(sessionJSON) {
+  if (!('server' in sessionJSON && _.isObject(sessionJSON.server))) {
+    return logValidationError("'server' property is missing or not an object");
+  }
+  const serverKeys = Object.keys(sessionJSON.server);
+  if (
+    serverKeys.length !== 2 ||
+    !serverKeys.includes(SERVER_TYPES.ADVANCED) ||
+    !_.isObject(sessionJSON.server.advanced)
```
mykola-mokhnach: I assume you wanted to use isPlainObject

app/common/renderer/utils/sessionfile-parsing.js
```diff
@@ -0,0 +1,203 @@
...
+ * @param {string} sessionFileString session file contents
+ * @returns session file in JSON format, or null if any validations failed
+ */
+export function parseSessionFileContents(sessionFileString) {
+  try {
+    const sessionJSON = JSON.parse(sessionFileString);
+    const updatedSessionJSON = migrateSessionJSON(sessionJSON);
+    return validateSessionJSON(updatedSessionJSON);
+  } catch {
+    return logParsingError('file is not valid JSON');
```
mykola-mokhnach: I assume try/catch here should only cover `JSON.parse` entity

test/unit/utils-sessionfile-parsing.spec.js
```diff
@@ -0,0 +1,220 @@
...
+
+    it('should return null if name is missing or not a string', function () {
+      const sessionWithoutName = {...versionProp, ...serverProp, ...capsProp};
+      const sessionWithInvalidName = {...sessionWithoutName, name: 123};
+      expect(validateSessionJSON(sessionWithoutName)).toBeNull();
+      expect(validateSessionJSON(sessionWithInvalidName)).toBeNull();
+    });
+
+    it('should return null if server is missing or invalid', function () {
+      const sessionWithoutServer = {...versionProp, ...nameProp, ...capsProp};
```
mykola-mokhnach: It makes sense to make these tests parametrized, like

```js
it.each([
  ['missing server', (base) => base],
  ['server is a number', (base) => ({ ...base, server: 123 })],
  ['server is empty object', (base) => ({ ...base, server: {} })],
  ['server has invalid structure', (base) => ({ ...base, server: { invalid: {}, advanced: {} } })],
  ['server has remote and local only', (base) => ({ ...base, server: { remote: {}, local: {} } })],
])('should return null if server is missing or invalid (%s)', (_, buildSession) => {
  const baseSession = { ...versionProp, ...nameProp, ...capsProp };
  const session = buildSession(baseSession);

  expect(validateSessionJSON(session)).toBeNull();
});
```

app/common/renderer/actions/SessionBuilder.js
eglitise: The sanitized filenames will be the same regardless of the platform, which is probably for the best in terms of file portability

Issue #880, opened 2023-06-07
chore: migrate Redux `connect` API to hooks


eglitise:
Modern Redux does not appear to use container components anymore, and instead uses [hooks and callback functions](https://redux.js.org/usage/migrating-to-modern-redux#modernizing-react-components-with-react-redux).
The `connect` API should therefore be replaced with `useDispatch`, `useSelector`, `useCallback` and so on.

PR #2543, opened 2026-01-09
feat: import/export sessions from Saved Capability Sets

### LINKED ISSUE #1644: feature request: Importing and Exporting capability sets as JSON files
### Current Behavior

Whenever user installs the Appium Inspector, it is then required to manually setup capabilities for different applications and devices. This makes it difficult for new team members to setup capabilities manually in Appium inspector to start application with provided capabilities.

Having the option to export and import capabilities is also useful to move existing saved configureations between different Systems and have them retained.

### Suggested Solution

Please add an option to export and import the JSON Capabilities. This make it easier to export the saved session capabilities to be transferred to other team members who can then import the capabiliities in their Appium Inspector to get started with Appium automation. 

Next to Save As... option under the saved capabilities set, we can put an option to Export this capability set that then prompts user to supply the location and filename to export the existing capability set.

Similarly, an option can be put under File menu to Import Capability Set that then prompts user to choose the JSON file containing all the key value pairs to start a particular session. Users can then easily save this imported session in Appium Inspector.

### Additional Information

_No response_

### LINKED ISSUE #2026: feature request: Export Saved Devices for sharing/migration
### Current Behavior

I have tried the plugin version, great idea:) I've been wondering for some time where the devices from the Saved Capability Sets tab are stored and whether it's possible to export them, for example, to JSON. It would be useful now for migrating from the standalone app to the web interface.

### Suggested Solution

I think it would be nice to have an export option for saved devices.

### Additional Information

_No response_

eglitise:
This PR adds a long-requested feature that greatly simplifies the import and export of `.appiumsession` files. Both operations have now been moved from the File menu to the Saved Capability Sets tab, thereby making them available in the browser/plugin versions as well:

* Move the Save Session functionality from the desktop-only File menu to a new button in each capability set's row
* Move the Open File functionality from the desktop-only File menu to a new button at the bottom of the saved capability sets list, and change it to automatically save the session instead of only opening it in the Capability Builder
  * Opening a session file directly via file manager/command line will still only open it in the Capability Builder
* Add the ability to import multiple session files at once
  * If any files are rejected, any remaining valid files are still imported
* Fix an issue where advanced server properties were not imported properly
* Update the Saved Gestures table footer for consistency with Saved Capability Sets table
* Update documentation

Here is a brief demonstration using Firefox:

https://github.com/user-attachments/assets/f9192e29-b31f-451e-a3fc-fe22959d58bb

Closes #1644 
Closes #2026

### DIFF
--- a/app/common/renderer/actions/SessionBuilder.js
+++ b/app/common/renderer/actions/SessionBuilder.js
@@ -406,8 +406,8 @@ export function saveSession(sessionParams, checkDuplicateName = false) {
     const {server, serverType, caps, name, uuid: foundUUID} = sessionParams;
     const savedSessions = (await getSetting(SAVED_SESSIONS)) || [];
     if (checkDuplicateName) {
-      const sessionsWithSameName = savedSessions.filter((session) => session.name === name);
-      if (sessionsWithSameName.length > 0) {
+      const duplicateSessionNameExists = savedSessions.some((session) => session.name === name);
+      if (duplicateSessionNameExists) {
         return dispatch({type: SET_CAPABILITY_NAME_ERROR});
       }
     }
@@ -655,16 +655,14 @@ export function importSessionFiles(fileList) {
       parsedSessions.push(sessionJSON);
     }
 
-    if (parsedSessions.length) {
-      for (const parsedSession of parsedSessions) {
-        await saveSession(parsedSession)(dispatch);
-      }
+    for (const parsedSession of parsedSessions) {
+      await saveSession(parsedSession)(dispatch);
     }
     dispatch({type: SESSION_UPLOAD_DONE});
 
     if (!_.isEmpty(invalidSessionFiles)) {
       notification.error({
-        title: i18n.t('unableToImportSessionFiles', {fileNames: invalidSessionFiles.toString()}),
+        title: i18n.t('unableToImportSessionFiles', {fileNames: invalidSessionFiles.join(', ')}),
         duration: 0,
       });
     }


app/common/renderer/actions/SessionBuilder.js
```diff
@@ -602,64 +606,105 @@ export function setSavedServerParams() {
...
+    for (const session of sessions) {
+      const {fileName, content, error} = session;
+      // Some error occurred while reading the uploaded file
+      if (error) {
+        invalidSessionFiles.push(fileName);
+        continue;
+      }
+      const sessionJSON = parseAndValidateSessionFileString(content);
+      if (!sessionJSON) {
+        invalidSessionFiles.push(fileName);
```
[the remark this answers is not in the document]
eglitise: It does still work though, but using the same variable name is certainly clearer.

app/common/renderer/actions/SessionBuilder.js
```diff
@@ -398,22 +401,23 @@ export function newSession(originalCaps, attachSessId = null) {
...
-    let savedSessions = (await getSetting(SAVED_SESSIONS)) || [];
-    let duplicateSessions = savedSessions.filter((session) => session.name === name);
-    // Ignore the check if the user is updating an existing capability set with duplicate names
-    let isEditingExistingCapability = duplicateSessions.find((session) => session.uuid === uuid);
-    if (duplicateSessions.length > 0 && !isEditingExistingCapability) {
-      return dispatch({type: SET_CAPABILITY_NAME_ERROR});
+    const {server, serverType, caps, name, uuid: foundUUID} = sessionParams;
+    const savedSessions = (await getSetting(SAVED_SESSIONS)) || [];
+    if (checkDuplicateName) {
+      const sessionsWithSameName = savedSessions.filter((session) => session.name === name);
```
mykola-mokhnach: you could also use `.find` or `.some`

app/common/renderer/actions/SessionBuilder.js
```diff
@@ -602,64 +606,105 @@ export function setSavedServerParams() {
...
+      }
+      const sessionJSON = parseAndValidateSessionFileString(content);
+      if (!sessionJSON) {
+        invalidSessionFiles.push(fileName);
+        continue;
+      }
+      parsedSessions.push(sessionJSON);
+    }
+
+    if (parsedSessions.length) {
```
mykola-mokhnach: this condition is redundant as the below loop won't be executed if parsedSessions is empty

app/common/renderer/actions/SessionBuilder.js
```diff
@@ -602,64 +606,105 @@ export function setSavedServerParams() {
...
-    switchTabs(SESSION_BUILDER_TABS.CAPS_BUILDER)(dispatch, getState);
   };
 }
 
+function parseAndValidateSessionFileString(sessionFileString) {
+  const sessionJSON = parseSessionFileContents(sessionFileString);
+  if (sessionJSON === null) {
+    return null;
+  }
+  sessionJSON.serverType = Object.keys(sessionJSON.server).find(
```
mykola-mokhnach, line 677: can the result be undefined?

app/common/renderer/components/SessionBuilder/SavedCapabilitySetsTab/SavedCapabilitySets.jsx
```diff
@@ -63,6 +74,11 @@ const SavedCapabilitySets = (props) => {
     );
   };
 
+  const findAndExportSavedSession = (uuid) => {
+    const session = getSessionById(savedSessions, uuid, t);
+    exportSavedSession(session);
```
mykola-mokhnach, line 79: does this call do anything? exportSavedSession returns a function

app/electron/main/helpers.js
```diff
@@ -13,12 +14,7 @@ export function setupIPCListeners() {
   ipcMain.on('electron:openLink', (_evt, link) => shell.openExternal(link));
   ipcMain.on('electron:copyToClipboard', (_evt, text) => clipboard.writeText(text));
   ipcMain.on('electron:setTheme', (_evt, theme) => (nativeTheme.themeSource = theme));
-  ipcMain.handle('sessionfile:open', async (_evt, filePath) => openSessionFile(filePath));
+  ipcMain.handle('sessionfile:open', async (_evt, filePath) => readFileSync(filePath, 'utf8'));
```
mykola-mokhnach, line 17: cannot we read files asynchronously?

app/common/renderer/actions/SessionBuilder.js
eglitise: No, because line 673 calls `parseSessionFileContents` -> `isSessionServerValid`, which validates that a second `server` property exists. If it fails, `sessionJSON` is set to `null`, causing this method to return on line 677.

app/common/renderer/components/SessionBuilder/SavedCapabilitySetsTab/SavedCapabilitySets.jsx
eglitise: Yes, that's just how (old) Redux syntax works - it injects `exportSavedSession` as a prop to be used in React components. This approach is used throughout the code, though this particular method does not utilize the `dispatch` and `getState` properties used in other methods. It does still work as expected though.

app/electron/main/helpers.js
eglitise: I guess we can, but I don't really see a benefit for it here. This IPC is only invoked when the user opens an `.appiumsession` file via file manager/command line, so I find that it makes sense to always wait for the Inspector to load the file contents.

app/common/renderer/components/SessionBuilder/SavedCapabilitySetsTab/SavedCapabilitySets.jsx
mykola-mokhnach: this is one of the reasons I cannot stand frontend development, bloody framework dark magic

app/electron/main/helpers.js
mykola-mokhnach: wouldn't it wait anyway? my concern is that sync calls pause the event loop, so the whole app interface might become unresponsive if it takes too long to load the file.

app/common/renderer/components/SessionBuilder/SavedCapabilitySetsTab/SavedCapabilitySets.jsx
eglitise: Arguably this syntax is also outdated and should certainly be updated at some point as per #880/#881

app/electron/main/helpers.js
eglitise: True, but thankfully the files are quite small, and the current version of the app already uses sync loading - I just tested it and I did not observe any notable delays.

mykola-mokhnach: it's not about the file size, but rather about the source, e.g. it might be a slow network drive

PR #2565, opened 2026-01-16
feat: add buttons to switch between MJPEG stream and screenshot API

### LINKED ISSUE #1505: feature request: Allow switching between mjpeg stream and screenshot mode for loading images
### Current Behavior

When the session has the`mjpegServerPort` capability set, then the image will be loaded directly using the MJPEG stream. If the`mjpegServerPort` is not configured, the system will fall back to using the screenshot API.

### Suggested Solution

Using mjpegServerPort for image streaming is effective only when the inspector and device are on the same machine. In scenarios where users connect to a device on a remote Appium server, the image will not render due to an unreachable MJPEG stream URL. Additionally, any issues with the streaming server will prevent the image from loading.

To address this, we can offer a fallback mechanism, allowing users to choose between using the MJPEG stream or the screenshot API based on their requirements. This will enable users to utilize remote devices without interruptions.

### Additional Information

_No response_

### LINKED ISSUE #2033: bug: Appium inspector not loading image in android with "appium:mjpegServerPort" from remote host ip

### Current Behavior

when user uses capability - **appium:mjpegServerPort**, appium inspector is not loading image.
Issue only happens when device is connected to remote machine in same network and server is not localhost

Image

Image

Working when server is localhost or 127.0.0.1, but not working with ip-address from same network

*only happening in android

### Expected Behavior

appium should load image in appium inspector though user uses appium:mjpegServerPort. if mjpegServer is not working or video url not working it should fall back to default screenshot

### Operating System

Mac

### Appium Inspector Version

App Version: 2024.6.1 Electron: 13.6.9 Node.js: 14.16.0

### Appium Version

2.17.1

### Further Information

_No response_

eglitise:
This PR adds a requested feature for sessions with MJPEG streaming, that allows switching their screenshot origin between the MJPEG stream and the W3C screenshot API used in non-MJPEG sessions.

To minimise layout shifts, the download screenshot button is no longer hidden in MJPEG mode, but instead shown in a disabled state. It can be enabled by switching to the screenshot API.

Here is a simple demo:

https://github.com/user-attachments/assets/004fa463-9c21-41dd-b5d2-04ad537b3895

Closes #1505
Closes #2033


KazuCocoa (approved):

app/common/renderer/components/SessionInspector/SessionInspector.jsx
```diff
@@ -246,10 +263,25 @@ const Inspector = (props) => {
   const screenShotControls = (
     <div className={styles.screenshotControls}>
       <Space size="middle">
-        <Tooltip
-          title={t(showCentroids ? 'Hide Element Handles' : 'Show Element Handles')}
-          placement="topRight"
-        >
+        {serverDetails.mjpegScreenshotUrl !== null && (
```
mykola-mokhnach, line 266: so we only have this control if mjpegScreenshotUrl capability is provided?

mykola-mokhnach (approved):

app/common/renderer/components/SessionInspector/SessionInspector.jsx
eglitise: It also works when providing `mjpegServerPort`, as it is then used to assemble `mjpegScreenshotUrl`: https://github.com/appium/appium-inspector/blob/main/app/common/renderer/actions/SessionBuilder.js#L373

PR #2567, opened 2026-01-16
feat: switch syntax highlighting to react-refractor

eglitise:
This changes the syntax highlighting package from `highlight.js` to `react-refractor`. There are several reasons for this:
* `highlight.js` lacks built-in support for Robot Framework, and the `highlightjs-robot` package does not fully cover Robot's syntax. However, `react-refractor` is based on `prism`, which does have Robot support.
* `react-refractor` is built for React and is imported as a React component. This allows to remove the `dangerouslySetInnerHTML` directives used to make `highlight.js` work.
* Bundle size improvements: the size of the transformed `index.js` used in the browser bundle is reduced by almost 22%! Worth noting that optimized `highlight.js` imports do result in an even smaller package, but not by much.

The highlighter colors in both light and dark mode are also changed as a result:
dark-new
light-new
Old colors - note the inconsistency between numerical and boolean values:
dark-old
light-old


app/common/renderer/lib/client-frameworks/robot.js
```diff
@@ -95,7 +97,7 @@ Tap With Positions    $\{100}    $\{positions}`;
     // change the JSON object into a format accepted by Create Dictionary: a sequence of key=value
     const cleanedJson = _.omitBy(jsonArg[0], _.isUndefined);
     const argsValuesStrings = _.map(cleanedJson, (v, k) => `${k}=${this.getRobotVal(v)}`);
-    return `&{scriptArgument}    Create Dictionary    ${argsValuesStrings.join('    ')}
+    return `&{scriptArgument} =    Create Dictionary    ${argsValuesStrings.join('    ')}
```
mykola-mokhnach, line 100: not sure about this change, is it a correct RF syntax?

eglitise: Yes, this was a bug: https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Create%20Dictionary

PR #2641, opened 2026-02-09
feat: add button to detach from session without termination

eglitise:
This PR adds a new button in the Session Inspector header, allowing the user to detach the Inspector from the active session, without actually deleting the session. There can be various use cases (related to attaching to session) where this functionality can be quite useful.

Quick demo:

https://github.com/user-attachments/assets/f513c96b-ad09-4556-9dac-00cfed01a23d

### DIFF
--- a/docs/session-inspector/header.md
+++ b/docs/session-inspector/header.md
@@ -134,9 +134,10 @@ disconnected device, this functionality is disabled by default.
 ![Detach Button](./assets/images/header/detach-button.png)
 
 This button disconnects the Inspector from the active session and returns to the
-[Session Builder](../session-builder/index.md), but the session itself is not deleted, and remains
-running on the Appium server. This approach may be useful if you are attaching to an existing
-session mid-test and want to resume the test afterwards, or if you want to reuse the session later.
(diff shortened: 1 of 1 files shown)


docs/session-inspector/header.md
```diff
@@ -129,8 +129,17 @@ session (selecting an element, interacting with the screenshot, searching for el
 that returns a failed response. Since such a response is not guaranteed to be caused by a
 disconnected device, this functionality is disabled by default.
 
+## Detach from Session
+
+![Detach Button](./assets/images/header/detach-button.png)
+
+This button disconnects the Inspector from the active session and returns to the
+[Session Builder](../session-builder/index.md), but the session itself is not deleted, and remains
```
KazuCocoa: Can we emphasize this doesn't delete the session? Like:

```suggestion
[Session Builder](../session-builder/index.md), but the session itself is not **DELETED**, and **REMAINS**
```

Or add a guide to use `Quit Session` for a case that wants to delete the session.


In the past, I met users who didn't know about the difference between `driver.quit` and just stopped sending HTTP requests from the client (actually not with this inspector, though). For them, they though both cases terminated the session. So this detachment might let some users think they deleted the session as well.

PR #2649, opened 2026-02-11
chore: improve logging when installing dev extensions

eglitise:
I was having some issues with the React/Redux devtools extensions, and the current lack of logging information made debugging more difficult than it could have been. This PR therefore adds some logging in the extension-loading process in order to improve DX.
I've also added back `@tomjs/electron-devtools-installer` as it fixes some deprecation warnings.

### DIFF
--- a/app/electron/main/debug.js
+++ b/app/electron/main/debug.js
@@ -40,7 +40,7 @@ export async function installExtensions() {
       }
     }
   } catch (e) {
-    logWithExtPrefix(`Error installing extensions: ${e}`);
+    logWithExtPrefix(`Error installing extensions: ${e.message}`);
   }
   logEmptyLine();
 }


app/electron/main/debug.js
```diff
@@ -1,12 +1,46 @@
...
+      logWithExtPrefix('No extensions were installed');
+    } else {
+      logWithExtPrefix(`Installed extensions at ${join(installedExts[0].path, '..')}:`);
+      for (const ext of installedExts) {
+        logWithExtPrefix(`* ${ext.name} v${ext.version} (ID: ${ext.id})`);
+      }
+    }
   } catch (e) {
-    console.warn(`Error installing extension: ${e}`); // eslint-disable-line no-console
+    logWithExtPrefix(`Error installing extensions: ${e.message}`);
```
mykola-mokhnach, line 43: should we also log it under a different level (e.g. console.warn/error)?

eglitise: In theory yes, but I don't think it makes much of a difference here

PR #2819, opened 2026-05-03
chore: Migrate vendor scripts to typescript

mykola-mokhnach:


app/common/renderer/lib/vendor/base.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  protected _validateUrl(url: string): URL {
+    try {
+      return new URL(url);
+    } catch {
+      throw new Error(`${i18n.t('Invalid URL:')} ${url}`);
+    }
+  }
+
+  protected _saveProperties(vendor: MutableVendorSlice, props: VendorConnectionFields): void {
```
eglitise: The existing JS code uses VendorProperties instead of VendorConnectionFields

app/common/renderer/lib/vendor/rabbitqa.ts
```diff
@@ -0,0 +1,30 @@
+import {BaseVendor} from './base.js';
+
+export class RabbitQAVendor extends BaseVendor {
+  override async configureProperties(): Promise<void> {
+    const rabbitQA = this._server.rabbitqa ?? {};
```
eglitise: No need for the `?? {}`, it's also not used for other vendors

app/common/renderer/lib/vendor/saucelabs.ts
```diff
@@ -1,5 +1,5 @@
 import dayjs, {extend as dayjsExtend} from 'dayjs';
-import localizedFormat from 'dayjs/plugin/localizedFormat';
+import localizedFormat from 'dayjs/plugin/localizedFormat.js';
```
eglitise, line 2: This import does not have the `.js` extension: https://day.js.org/docs/en/plugin/localized-format

mykola-mokhnach: typescript complains it cannot find the module if I remove the extension

app/common/renderer/lib/vendor/rabbitqa.ts
mykola-mokhnach: removed

app/common/renderer/lib/vendor/base.ts
mykola-mokhnach: using VendorProperties

app/common/renderer/lib/vendor/saucelabs.ts
eglitise: I'd probably disable TS for that line. In the source code, this is a directory, not a file: https://github.com/iamkun/dayjs/tree/dev/src/plugin/localizedFormat

mykola-mokhnach: In the compiled module it is resolved properly: 

Screenshot 2026-05-04 at 17 19 36

eglitise (approved):

PR #2892, opened 2026-06-01
fix(lambdatest): fix session listing and attach-to-session for LambdaTest

SahilSawLT:
## Summary

Fixes for LambdaTest (TestMu AI) sessions in Appium Inspector:

- **Session listing**: Calls LambdaTest's sessions API directly instead of `/wd/hub/sessions`. The standard `/wd/hub/sessions` endpoint only returns app automation sessions; the dedicated API returns all running mobile sessions (RD App, RD Web, VD App, VD Web) across Android and iOS. LTMA host is derived automatically from the configured hub host — no hardcoded URLs.
- **Attach to session — capabilities unwrapping**: LambdaTest returns capabilities in a nested structure `{capabilities: {desired: {...}}}`. Unwrap before using as session caps to prevent errors during attach.
- **Attach to session — appMode detection**: Use `driver.capabilities.browserName` instead of `sessionCaps.browserName` when attaching to an existing session, so web sessions correctly use xpath/css locators instead of native accessibility id.

## Test plan

- [ ] Open Appium Inspector with TestMu AI server type configured
- [ ] Go to **Attach to Session** tab — running sessions list should populate
- [ ] Attach to a running web session — locators should work with xpath/css
- [ ] Attach to a running app session — locators should work with native accessibility id
- [ ] **Start Session** tab — start session with valid caps should work unchanged

🤖 Generated with [Claude Code](https://claude.com/claude-code)


japneetlambdatest (commented):
## PR Review: fix(lambdatest): fix session listing and attach-to-session for LambdaTest

Fix is correctly scoped and addresses three real bugs (sessions API, nested caps unwrap, web-session appMode detection). Main concerns are around the host-derivation logic for the sessions endpoint — it assumes a narrow set of host shapes and silently swallows errors, which will bite on-prem / staging configs.

| # | Severity | Category | Issue |
|---|----------|----------|-------|
| 1 | HIGH | Edge case / On-prem | Hardcoded `https://` and dropped port; sessions URL ignores `LAMBDATEST_PORT` and `https` flag |
| 2 | MEDIUM | Edge case | Host derivation regex misses non-canonical `LAMBDATEST_HOST` values |
| 3 | MEDIUM | UX | `catch` swallows all errors → empty list hides auth/network/5xx failures |
| 4 | LOW | Edge case | `unwrapTestMuAICaps` throws TypeError on `{capabilities: null}` |
| 5 | LOW | Architecture | TestMuAI-only branch for `driver.capabilities`; same bug likely exists for other vendors |

eglitise:
Thanks for the PR!
Code generally seems fine. I'm not a fan of adding vendor-specific `if` branches, but solving this problem would require a larger refactor.
Could you please sign the CLA and fix the lint errors?

app/common/renderer/actions/SessionBuilder.js
```diff
@@ -736,6 +749,22 @@ export function getRunningSessions() {
...
+        .replace(/^hub-/, 'api-');
+      const sessionListUrl = `https://${sessionListHost}/automation/api/v1/appium/inspector/sessions`;
+      try {
+        const res = await fetchSessionInformation({url: sessionListUrl, headers});
+        dispatch({type: GET_SESSIONS_DONE, sessions: res.value ?? []});
+      } catch (err) {
+        log.error('Failed to fetch running sessions', err);
+        dispatch({type: GET_SESSIONS_DONE, sessions: []});
+      }
+      return;
```
eglitise: Could you extract this into a separate function?

SahilSawLT:
Thanks for the review, @eglitise! I've pushed a few more changes and resolved your comment. Could you please take another look when you get a chance?

app/common/renderer/lib/appium/inspector-driver.js
```diff
@@ -335,8 +337,44 @@ export default class InspectorDriver {
...
     // Get all available contexts (or the error, if one appears)
     try {
-      contexts = await this.driver.executeScript('mobile:getContexts', []);
-      contexts = isAndroid ? this.parseAndroidContexts(contexts) : contexts;
+      if (this.serverType === SERVER_TYPES.TESTMUAI) {
+        // The standard /contexts endpoint is authoritative for which IDs the
+        // driver will accept in setContext. mobile:getContexts returns richer
+        // data (titles, URLs, pages), but in some configurations it can produce
+        // either too few entries (parser filters drop webviews) or too many
+        // (one entry per browser tab, all collapsing to the same webview ID).
```
eglitise: If you are having issues with this execute method, the correct way to fix this would be to create an issue report on the Appium repo, so that the problem is resolved for all consumers. I do not think that creating a custom workaround just for TestMu AI is a suitable approach.

eglitise:
Please also fix the linter errors by running `npm run lint`.

eglitise:
Sorry I forgot to add, please run `npm run prettier` to resolve the formatting errors as well.

SahilSawLT:
Hi @eglitise, completely understand your point about avoiding vendor-specific branches, and I agree the root `mobile: getContexts` inconsistency should be fixed upstream for all consumers — I'll file that issue on the Appium repo.

In the meantime, we have a LambdaTest customer currently blocked by this on the Inspector, so I'd really appreciate it if we could merge the context fix as an interim unblock. A couple of points that I hope make it lower-risk:

- The change is **fully gated to the TestMu AI server type** — every other server type takes the exact existing code path, so there's no behavior change for any other provider.
- Once the upstream fix lands, I'm happy to come back and remove this branch entirely so the workaround doesn't linger.

Would that be acceptable as a temporary measure? Thanks again for the reviews and your patience here.

SahilSawLT:
As mentioned, I've filed the upstream issue for the root `mobile: getContexts` inconsistency: appium/appium#22423.

eglitise:
Thanks, but I still believe that adding this workaround is not the way to solve this:

* These changes add vendor-dependent behavior into the currently vendor-agnostic core communication between the Inspector and the Appium server
* Any changes in the driver's `mobile: getContexts` behavior (due to the issue you reported above, or otherwise) may cause issues with this workaround code
* The UiAutomator2 driver receives more frequent releases than the Inspector, so applying the fixes from the driver side instead will cause them to be available sooner

app/common/renderer/actions/SessionBuilder.js
```diff
@@ -344,14 +358,23 @@ export function newSession(originalCaps, attachSessId = null) {
...
     // we want to keep the process equal to prevent complexity so we launch a default url here to make
     // sure we don't start with an empty page which will not show proper HTML in the inspector
-    const {browserName = ''} = sessionCaps;
+    // When attaching via TestMu AI, sessionCaps is empty (user only provided a session ID).
+    // Use driver.capabilities in that case, which the WebDriver client populates from the actual
+    // running session after a successful attach — correctly reflecting browserName for web sessions.
+    const {browserName = ''} =
+      attachSessId && session.serverType === SERVER_TYPES.TESTMUAI
+        ? driver.capabilities
+        : sessionCaps;
```
eglitise: I checked the surrounding code and it seems that if `attachSessId` is set, `sessionCaps` can never contain `browserName`, so this fix would benefit all server types. You can therefore remove the `serverType` gating and simplify this to a one-liner:

```js
const {browserName = ''} = attachSessId ? driver.capabilities : sessionCaps;
```

SahilSawLT:
Thanks @eglitise, that all makes sense — agreed the driver is the right place for the contexts fix.

I've updated the PR:
- **Removed the context-listing workaround entirely** — we'll let it be fixed on the driver side (tracked in appium/appium#22423).
- **Applied your simplification** for the attach case, dropping the `serverType` gating:
  ```js
  const {browserName = ''} = attachSessId ? driver.capabilities : sessionCaps;
  ```

Let me know if there's anything else you'd like changed.

eglitise (approved):
Thank you!

SahilSawLT:
Thanks for merging, @eglitise! Roughly when can we expect these changes to land in a release — and in the hosted web version of the Inspector? Just trying to give our customer a timeline. Appreciate all the reviews!

eglitise:
There's not that many unreleased features/fixes at the moment, so the next version (which will include these changes) will probably release sometime in July.

The hosted web version is no longer maintained - please refer to the README. Note that users can instead host a web version locally, using the Inspector's Appium plugin version.

SahilSawLT:
Hi @eglitise, just following up on this since we're now in July. Is there any update on when the next release (including these changes) is expected? We'd like to keep our customer informed on the timeline. Thanks again!

eglitise:
We don't have a defined release schedule. It's more likely be towards the end of the month.

eglitise:
@SahilSawLT version 2026.7.1 includes these changes.

PR #2921, opened 2026-06-11
fix: improve handling of UiA2 devices with different size displays

eglitise:
This is a fix related to devices with multiple displays. I was testing an Android emulator with two displays of different sizes, and noticed that, upon switching the active display, element highlighters were no longer aligned to the screenshot of the newly selected display.

The core issue stemmed from the `deviceScreenSize` capability, which was automatically set at the start of the session, and remained the same regardless of the active display. The fix no longer relies on this capability, but instead retrieves the display size using the `mobile: deviceInfo` method, whose output properly matches the active display. To avoid unnecessary computation, I have also added caching, which allows to skip invoking this method if the `getWindowRect` dimensions have not changed.

### DIFF
--- a/app/common/renderer/lib/appium/inspector-driver.js
+++ b/app/common/renderer/lib/appium/inspector-driver.js
@@ -42,6 +42,7 @@ export default class InspectorDriver {
     this.elementCache = {};
     this.elVarCount = 0;
     this.elArrayVarCount = 0;
+    this.windowSizeCache = {raw: {}, adjusted: {}};
   }
 
   static instance(driver) {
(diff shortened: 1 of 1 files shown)


app/common/renderer/lib/appium/inspector-driver.js
```diff
@@ -265,37 +266,45 @@ export default class InspectorDriver {
...
+      // Finally, update the cache
+      this.windowSizeCache.raw = driverWindowSize;
+      this.windowSizeCache.adjusted = adjustedWindowSize;
+      return {windowSize: adjustedWindowSize};
+    } catch (err) {
+      if (err.name === UNKNOWN_ERROR) {
+        throw err;
+      }
+      return {windowSizeError: err};
     }
```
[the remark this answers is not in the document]
eglitise: this is ok, we'll show the error message instead

app/common/renderer/lib/appium/inspector-driver.js
```diff
@@ -265,37 +266,45 @@ export default class InspectorDriver {
...
+        return {windowSize: driverWindowSize};
       }
-    } catch (e) {
-      if (e.name === UNKNOWN_ERROR) {
-        throw e;
+      // UiAutomator2 requires adjustments depending on the result of `mobile: deviceInfo`.
+      // Use a cache to avoid re-triggering this method every time.
+      if (_.isEqual(driverWindowSize, this.windowSizeCache.raw)) {
+        return {windowSize: this.windowSizeCache.adjusted};
+      }
```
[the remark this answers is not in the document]
eglitise: we probably want to update the cache even when only x/y change

app/common/renderer/lib/appium/inspector-driver.js
```diff
@@ -265,37 +266,45 @@ export default class InspectorDriver {
...
+      // Window size has changed since the cached value - trigger the method
+      const adjustedWindowSize = {...driverWindowSize};
+      const {realDisplaySize} = await this.driver.executeScript('mobile:deviceInfo', []);
+      // realDisplaySize value fits portrait mode for phones, but landscape mode for tablets
+      const [width, height] = realDisplaySize.split('x').map((param) => parseInt(param, 10));
```
mykola-mokhnach, line 288: this might not always be true

eglitise: The calculation here is the same as before -  the only difference is that `realDisplaySize` ([which is the same as the previous `deviceScreenSize`](https://github.com/appium/appium-uiautomator2-driver/blob/master/lib/driver.ts#L491)) is now recalculated upon `getWindowRect` changes, instead of always having the same value during the session.

mykola-mokhnach (approved):

app/common/renderer/lib/appium/inspector-driver.js
mykola-mokhnach: I mean that not all tablets default to landscape when they return their realDisplaySize.
In order to guarantee the orientation it must be retrieved separately.
Android API doc for Display.getRealSize() says

```
The size is adjusted based on the current rotation of the display.
```

eglitise: Interesting; this line of code had already been previously changed to improve compatibility with tablets in particular, since the value of `deviceScreenSize` had its dimensions in the opposite order to phones.
I did also test this PR in the Pixel Tablet simulator in both orientations, and it still looked fine.

PR #2949, opened 2026-06-24
feat: add Custom Export Modal for exporting UI elements with customiz…

cuneytcakir:
One-line summary
----------------
Make Custom Export profile import robust: accept single-profile JSON and validate imported profiles to avoid corrupt entries

--------------

## Summary

Improve the Custom Export profile import behavior so that:

- Import accepts both arrays of profiles and a single-profile JSON object.
- Imported profiles are validated and malformed/invalid profiles are rejected (or skipped) to avoid corrupt profiles in localStorage.

## Why

- Exporters can produce a single JSON object for a single profile. Previously the importer expected an array only, so single-profile export files were ignored.
- Malformed or incomplete profiles could be imported and persisted, causing incorrect behavior later.
- This change makes import behavior predictable and safe.

## What changed

- Modified: `app/common/renderer/components/SessionInspector/Header/CustomExportModal.jsx`
  - `handleImportProfiles` now accepts both a JSON array or a single JSON object (wrapped to an array).
  - Added an `isValidProfile(profile)` helper to validate required fields before import.
  - Invalid profiles are skipped; if none are valid an alert is shown and import is aborted.
  - If some entries are invalid, valid entries are imported and user is warned how many were skipped.
  - Duplicate profiles (same `name`) are not added again.
  - Malformed JSON still triggers an error alert.

## Validation rules (implemented in `isValidProfile`)

- `profile` must be an object
- `name` must be a non-empty string
- `filters` must be an array
- `attributes` must be an array
- `priority` must be an array
- `mode` must be `'array'` or `'object'`
- `outputFieldMap` must be an object with `customFields` array

### DIFF
--- a/app/common/renderer/actions/SessionInspector.js
+++ b/app/common/renderer/actions/SessionInspector.js
@@ -1097,3 +1097,191 @@ function parseAndValidateGestureFileString(gestureFileString) {
   }
   return _.omit(gestureJSON, ['id', 'date']);
 }
+
+function hasUsableLocatorAttrs(attributes) {
+  if (!attributes) return false;
+  const locatorAttrs = [
+    'resource-id',
+    'id',
+    'content-desc',
+    'contentDescription',
+    'text',
+    'value',
+    'name',
+    'label',
+  ];
+  return locatorAttrs.some((k) => typeof attributes[k] === 'string' && attributes[k].trim() !== '');
+}
+
+function isDisplayedAndEnabled(attributes) {
+  if (!attributes) return false;
+  const enabled = attributes.enabled ?? attributes['enabled'];
+  const displayed = attributes.displayed ?? attributes.visible ?? attributes['visible'];
+  // values are strings like 'true' / 'false' in xmlToJSON
+  const isEnabled = typeof enabled === 'undefined' ? true : String(enabled) === 'true';
+  const isDisplayed = typeof displayed === 'undefined' ? true : String(displayed) === 'true';
+  return isEnabled && isDisplayed;
+}
+
+function safeNameFrom(attrValue, fallback) {
+  if (attrValue && typeof attrValue === 'string') {
+    return attrValue.replace(/\s+/g, '_').replace(/[^a-zA-Z0-9_]/g, '').replace(/^(\d)/, '_$1') || fallback;
+  }
+  return fallback;
+}
+
+/**
+ * Export Visible Elements (JSON)
+ */
+export function exportVisibleElements() {
+  return async (dispatch, getState) => {
+    try {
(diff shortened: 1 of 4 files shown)


eglitise:
Your PR description does not clearly state the purpose of this PR. Could you please explain these changes? Preferably in your own words, without using an LLM.

cuneytcakir:
With this feature, instead of searching for elements one by one on a page, testers can download them in their desired array or JSON format by applying specific filters. They can also save this as a profile and share it with all their teammates. I've been using it in my team for a few weeks now and have seen a significant improvement in our element finding and editing speed.Finding elements within crowded pages has become very difficult, so using filters to download them in our desired format has been really helpful.

eglitise:
Thanks, I think I'm starting to see the picture, though it was still very unclear, so I took a closer look at the code.

It seems like your new functionality generates suggested locators for all elements that match your filters, and then exports the elements and locators together. I personally find this to have very limited value:
* The exported file has the context of neither the original XML structure nor the screenshot. Without them, I don't see how it's possible to verify which element in the custom export is the one you actually need.
* In my experience it's quite common that I need to find an element that has no unique properties. Such situations usually require relying on parent/sibling/child elements, in which case the logic for generating suggested locators will be quite suboptimal, commonly returning brittle selectors. A better way to craft selectors for elements like this is by assembling them yourself, which is only possible with the context of the full XML tree structure.
* I don't understand the benefit of having multiple export formats (array or JSON)

Perhaps I'm missing something, and your workflows make use of these exported files for different purposes?

The code itself has various other issues as well:
* It seems to be heavily or entirely AI generated, with many issues that would be caught by a human reviewer (e.g. `xpath` is not a valid element attribute, `text` is not a valid locator strategy, plenty of unusual code structures that aren't used anywhere else in the project (`alert()`, `Modal.confirm`, mixed uses of `Input` and `input`, clicking on a nearby element using `document.getElementById`, etc.), plus a lot of lint/formatting warnings)
* The changes appear to be primarily targeting the UiAutomator2 driver, meaning the functionality might not fully work for any other driver
* Available options for properties like filters, attributes and locator strategies seem to be arbitrary - both the UiAutomator2 and XCUITest drivers define additional options for all of these

On the whole, I feel like these changes are too specialized towards the particular uses cases in your own team, and might not be as applicable for general use.

cuneytcakir:
You might be right; in our automation project, we store elements in JSON format. Generally, there's a neat element structure, and we can use the elements by generating JSON from the relevant page, but not everyone's implementation is that secure and neat. We might be able to convert it to a more generally usable format. Thanks for your comments :)

PR #3021, opened 2026-07-21
chore: Drop lodash

mykola-mokhnach:
Summary

- Remove the lodash runtime dependency and all _.xxx() usage across the renderer app, actions/reducers, client-framework code generators, and test helpers
- Replace the handful of non-trivial functions (isEmpty, isPlainObject, isEqual, omit, debounce) with a small shared helper module (app/common/renderer/utils/common.js, with unit tests); everything else converts to native JS (Object.entries/values/keys, Array.isArray, structuredClone, Object.hasOwn, etc.)
- Drop lodash / @types/lodash from package.json and update the lockfile (only a transitive dep of electron-builder/electron-settings remains)


app/common/renderer/utils/common.js
```diff
@@ -0,0 +1,110 @@
+/**
```
eglitise, line 1: The filename is a bit confusing; something like `general`, `generic` or `common` would be better.

mykola-mokhnach: I don't have any special preferences on this one. renamed to `common`

scripts/crowdin-common.mjs
```diff
@@ -24,7 +23,7 @@ export async function performApiRequest(suffix = '', opts = {}) {
     : `${API_ROOT}${suffix}`;
   log.debug(`Sending ${method} request to ${url}`);
   let formattedPayload = payload;
-  if (_.isPlainObject(payload)) {
+  if (payload !== null && typeof payload === 'object' && !Array.isArray(payload)) {
     formattedPayload = JSON.stringify(payload);
     log.debug(`Request payload: ${formattedPayload}`);
   }
```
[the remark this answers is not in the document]
mykola-mokhnach: fixed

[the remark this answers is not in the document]
mykola-mokhnach: fixed

app/common/renderer/actions/SessionBuilder.js
```diff
@@ -172,8 +172,8 @@ export function showError(e, params = {methodName: null, secs: 5, url: null}) {
   }
   if (
     errMessage === 'ECONNREFUSED' ||
-    _.includes(errMessage, 'Failed to fetch') ||
-    _.includes(errMessage, 'The requested resource could not be found')
+    errMessage.includes('Failed to fetch') ||
+    errMessage.includes('The requested resource could not be found')
   ) {
```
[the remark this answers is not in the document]
mykola-mokhnach: fixed

app/common/renderer/components/SessionBuilder/ServerDetails/CloudProviderSelector.jsx
```diff
@@ -36,10 +44,10 @@ const CloudProviderSelector = (props) => {
       title={t('Select Cloud Providers')}
     >
       {[
-        ..._.map(providersGrid, (row, key) => (
-          <Row gutter={16} key={key}>
+        ...providersGrid.map((row) => (
+          <Row gutter={16} key={row.join(',')}>
```
eglitise, line 48: Not sure why the `(row, key)` part was changed to just `(row)`

app/common/renderer/components/SessionInspector/CommandsTab/CommandParametersModal.jsx
```diff
@@ -32,13 +32,13 @@ const CommandParametersModal = ({
...
-      open={!_.isEmpty(curCommandDetails.details.params)}
+      open={!isEmpty(curCommandDetails.details.params)}
       onOk={() => prepareAndRunCommand(curCommandDetails)}
       onCancel={() => clearCurrentCommand()}
       footer={(_, {OkBtn}) => <OkBtn />}
     >
-      {_.map(curCommandDetails.details.params, (param, index) => (
-        <Space.Compact block key={index} className={styles.commandArgInputRow}>
+      {curCommandDetails.details.params.map((param, index) => (
+        <Space.Compact block key={param.name} className={styles.commandArgInputRow}>
```
eglitise, line 41: This key can also remain as-is

app/common/renderer/lib/client-frameworks/java-common.js
```diff
@@ -12,10 +11,11 @@ export default class JavaFramework extends CommonClientFramework {
...
       return `{${convertedItems.join(', ')}}`;
     } else if (typeof jsonVal === 'object') {
-      const cleanedJson = _.omitBy(jsonVal, _.isUndefined);
-      const convertedItems = _.map(
-        cleanedJson,
-        (v, k) => `Map.entry(${JSON.stringify(k)}, ${this.getJavaVal(v)})`,
+      const cleanedJson = Object.fromEntries(
+        Object.entries(jsonVal).filter(([, v]) => v !== undefined),
+      );
+      const convertedItems = Object.entries(cleanedJson).map(
```
eglitise: Since `cleanedJson` is only used in` convertedItems`, one `fromEntries` and one `entries` call can be removed

app/common/renderer/lib/client-frameworks/python.js
```diff
@@ -15,21 +14,21 @@ export default class PythonFramework extends CommonClientFramework {
...
       return `[${convertedItems.join(', ')}]`;
     } else if (typeof jsonVal === 'object') {
-      const cleanedJson = _.omitBy(jsonVal, _.isUndefined);
-      const convertedItems = _.map(
-        cleanedJson,
-        (v, k) => `${JSON.stringify(k)}: ${this.getPythonVal(v)}`,
+      const cleanedJson = Object.fromEntries(
+        Object.entries(jsonVal).filter(([, v]) => v !== undefined),
+      );
+      const convertedItems = Object.entries(cleanedJson).map(
```
eglitise: Same here

app/common/renderer/lib/client-frameworks/robot.js
```diff
@@ -95,8 +94,12 @@ Tap With Positions    $\{100}    $\{positions}`;
 
   codeFor_executeScriptWithArgs(scriptCmd, jsonArg, varAssignment = '') {
     // change the JSON object into a format accepted by Create Dictionary: a sequence of key=value
-    const cleanedJson = _.omitBy(jsonArg[0], _.isUndefined);
-    const argsValuesStrings = _.map(cleanedJson, (v, k) => `${k}=${this.getRobotVal(v)}`);
+    const cleanedJson = Object.fromEntries(
+      Object.entries(jsonArg[0]).filter(([, v]) => v !== undefined),
+    );
+    const argsValuesStrings = Object.entries(cleanedJson).map(
```
eglitise: Same here

app/common/renderer/lib/client-frameworks/ruby.js
```diff
@@ -14,18 +13,21 @@ export default class RubyFramework extends CommonClientFramework {
       const convertedItems = jsonVal.map((item) => this.getRubyVal(item));
       return `[${convertedItems.join(', ')}]`;
     } else if (typeof jsonVal === 'object') {
-      const cleanedJson = _.omitBy(jsonVal, _.isUndefined);
-      const convertedItems = _.map(cleanedJson, (v, k) => `${k}: ${this.getRubyVal(v)}`);
+      const cleanedJson = Object.fromEntries(
+        Object.entries(jsonVal).filter(([, v]) => v !== undefined),
+      );
+      const convertedItems = Object.entries(cleanedJson).map(
```
eglitise: Same here

app/common/renderer/lib/client-frameworks/ruby.js
```diff
@@ -296,3 +298,19 @@ driver.quit`;
     return `# Not supported: createWindow`;
   }
 }
+
+// Mimics lodash's `_.lowerCase`: splits a string into words on camelCase boundaries,
+// letter/digit boundaries, and non-alphanumeric separators, then lower-cases and
+// joins the words with a single space (e.g. 'fooBar' -> 'foo bar', 'FOO_BAR' -> 'foo bar').
+function lowerCaseWords(str) {
```
eglitise: Since `_.lowerCase` is only used in two functions, and lowercasing isn't used for any other client library, I think it's simpler to just remove the functionality

app/common/renderer/utils/locator-generation/common.js
```diff
@@ -67,5 +65,7 @@ export function getComplexSuggestedLocators(path, sourceDoc, isNative, automatio
   complexLocators.xpath = getOptimalXPath(sourceDoc, domNode);
 
   // Remove entries for locators where the optimal selector could not be found
-  return _.omitBy(complexLocators, _.isNil);
+  return Object.fromEntries(
+    Object.entries(complexLocators).filter(([, v]) => v !== null && v !== undefined),
```
eglitise: Probably can be simplified to `v != null`

app/common/renderer/utils/source-parsing.js
```diff
@@ -66,7 +67,8 @@ export function xmlToJSON(sourceXML) {
     }
 
     // Dot Separated path of indices
-    const path = _.isNil(index) ? '' : `${!parentPath ? '' : parentPath + '.'}${index}`;
+    const path =
+      index === null || index === undefined ? '' : `${!parentPath ? '' : parentPath + '.'}${index}`;
```
eglitise: Can probably simplify to `index == null`

app/common/renderer/utils/common.js
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Returns true if the value is a plain object (created via `{}` or `Object.create(null)`),
+ * as opposed to an array, class instance, or other object type.
+ *
+ * @param {*} value
+ * @returns {boolean}
+ */
+export function isPlainObject(value) {
+  if (value === null || typeof value !== 'object') {
```
eglitise: Maybe `value == null` or even `!value`. Same for the copy in `scripts/crowdin-common.mjs`

app/common/renderer/utils/common.js
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Returns true when the value has no elements/properties (mirrors lodash's `isEmpty`
+ * for the value types actually used in this codebase: null/undefined, strings, arrays,
+ * and plain objects).
+ *
+ * @param {*} value
+ * @returns {boolean}
+ */
+export function isEmpty(value) {
+  if (value === null || value === undefined) {
```
eglitise: `value == null`

app/common/renderer/components/SessionInspector/CommandsTab/CommandParametersModal.jsx
eglitise: Nevermind, this causes a lint warning

app/common/renderer/components/SessionInspector/CommandsTab/Commands.jsx
eglitise, line 124: This check should be changed to `curCommandDetails?.details?.params` to prevent an error in `CommandParametersModal` (as per one of Copilot's low confidence comments)

app/common/renderer/components/SessionInspector/SessionInfoTab/SessionInfoTable.jsx
```diff
@@ -32,7 +31,7 @@ const tableColumns = [
 ];
 
 const innerDataSource = (tableData) =>
-  _.toPairs(tableData).map(([propName, propValue]) => ({
+  Object.entries(tableData).map(([propName, propValue]) => ({
```
eglitise: As per one of Copilot's low-confidence comments, `tableData` can be undefined, which causes an error. I think this could be fixed by importing `isEmpty` and using something like this:

```js
const innerDataSource = (tableData) =>
  !isEmpty(tableData)
    ? Object.entries(tableData).map(([propName, propValue]) => ({
        key: propName,
        property: formatMono(propName),
        value: formatMono(String(propValue)),
      }))
    : [];
```

app/common/renderer/lib/client-frameworks/java-common.js
mykola-mokhnach: Fixed by removing the redundant `fromEntries`/`entries` roundtrip.

app/common/renderer/lib/client-frameworks/python.js
mykola-mokhnach: Fixed, same as the `java-common.js` case.

app/common/renderer/lib/client-frameworks/robot.js
mykola-mokhnach: Fixed, same as above.

app/common/renderer/lib/client-frameworks/ruby.js
mykola-mokhnach: Fixed, same as above.

app/common/renderer/lib/client-frameworks/ruby.js
mykola-mokhnach: Removed `lowerCaseWords` and switched both call sites to plain `.toLowerCase()`.

app/common/renderer/utils/locator-generation/common.js
mykola-mokhnach: Done.

app/common/renderer/utils/source-parsing.js
mykola-mokhnach: Done.

app/common/renderer/utils/common.js
mykola-mokhnach: Done for both, and for the copy in `scripts/crowdin-common.mjs`.

app/common/renderer/utils/common.js
mykola-mokhnach: Done.

app/common/renderer/components/SessionInspector/SessionInfoTab/SessionInfoTable.jsx
mykola-mokhnach: Fixed as suggested.

app/common/renderer/components/SessionInspector/CommandsTab/Commands.jsx
mykola-mokhnach: This is no longer applicable: `CommandParametersModal` only renders when `curCommandDetails` is truthy (guarded by `{!!curCommandDetails && (...)}`), and every caller of `startCommand` always passes a `details` object, so `curCommandDetails.details.params` can't throw here.

eglitise: I tested this myself and it threw an error for commands without parameters. This is because `curCommandDetails.details` will indeed always be defined, but `curCommandDetails.details.params` will only be defined for commands that have parameters.

mykola-mokhnach: Good catch — this was the lost `_.map(undefined) === []` safety net from lodash. Fixed with `(curCommandDetails.details.params ?? []).map(...)`.

app/common/renderer/utils/other.js
```diff
@@ -16,9 +14,9 @@ export function addVendorPrefixes(caps) {
   return caps.map((cap) => {
     // if we don't have a valid unprefixed cap or a cap with an existing prefix, update it
     if (
-      !_.isUndefined(cap.name) &&
+      cap.name !== undefined &&
       !STANDARD_W3C_CAPS.includes(cap.name) &&
-      !_.includes(cap.name, ':')
+      !cap.name.includes(':')
     ) {
```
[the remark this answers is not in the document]
mykola-mokhnach: Good catch. Fixed by requiring `typeof cap.name === 'string'` before calling `.includes`, so malformed non-string cap names are skipped instead of crashing.

app/common/renderer/reducers/SessionBuilder.js
```diff
@@ -421,3 +418,13 @@ export default function builder(state = INITIAL_STATE, action) {
...
+
+// `server` is always exactly two levels deep (serverType -> flat scalar config), so merging
+// each serverType's config one level deep is equivalent to a full recursive deep merge here.
+function mergeServerState(currentServer, incomingServer) {
+  const nextServer = {...currentServer};
+  for (const [serverType, config] of Object.entries(incomingServer)) {
+    nextServer[serverType] = {...(nextServer[serverType] || {}), ...config};
+  }
+  return nextServer;
+}
```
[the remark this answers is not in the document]
mykola-mokhnach: Confirmed — verified against real `_.merge` behavior: a non-plain-object source value fully replaces the destination rather than being merged into it. `mergeServerState` now uses `isPlainObject(config)` to only spread-merge when `config` is a plain object, and overwrites otherwise, matching lodash's semantics without throwing.

app/common/renderer/utils/common.js
```diff
@@ -0,0 +1,110 @@
+/**
+ * Small collection of generic helpers that used to be provided by lodash.
+ * Kept minimal on purpose: only the semantics actually needed by this codebase are implemented.
+ */
```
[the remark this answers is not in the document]
mykola-mokhnach: Good catch — updated the PR description to reference `common.js` (the actual final filename after the earlier rename).

eglitise (approved):

Issue #1342, opened 2024-02-12
XPath as a locator coming in recorded script


bawarim:
### Is this an issue specifically with Appium Inspector?

- [X] I have verified that my issue does _not_ occur with Appium, and should be investigated as an Appium Inspector issue

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When recording operation using appium inspector recorder then only XPath is coming as a locator strategy. 
![image](https://github.com/appium/appium-inspector/assets/46487726/cbd29b0d-3449-4410-86de-95acdd598617)

### Expected Behavior

There should be an option of having different locator to be configured while recording. And same should be generated in recorded script. 

### Operating System

Windows

### Appium Inspector Version

2023.12.2

bawarim:
The elements has unique IDs available, but sill it uses the XPath. 

![image](https://github.com/appium/appium-inspector/assets/46487726/b1895175-fd74-4c8a-b490-0fe5028f4279)

![image](https://github.com/appium/appium-inspector/assets/46487726/3a9a81a6-a631-4235-911b-a4711bd657be)

eglitise:
Assuming this is an Android hybrid app, this makes sense, as the `id` locator strategy [is not part of the W3C spec](https://www.w3.org/TR/webdriver2/#locator-strategies). Chrome runs in W3C mode by default, so it will only work with the available W3C strategies, of which the only one here is `xpath`.
However, you can disable Chrome's W3C mode, and the recorder will then be able to use other strategies. You can do this by starting the Inspector session with the `"appium:chromeOptions": {"w3c": false}` capability.
Check the [UiAutomator2 web context capabilities](https://github.com/appium/appium-uiautomator2-driver#web-context) for more info.

PR #3039, opened 2026-07-24
feat: add Tap By Element screenshot interaction mode

### LINKED ISSUE #3037: Feature Proposal: "Tap By Element" screenshot interaction mode for the Recorder
# Feature Proposal: "Tap By Element" screenshot interaction mode for the Recorder

## Motivation

Today, the Recorder can already produce locator-based code (e.g. `driver.$('~someId').click()`)
for element interactions — but only via a two-step flow: switch to **Element Mode**, click a
highlighter to select the element, then click the **Tap** button in the Selected Element panel.

If a user instead directly taps the app screenshot while in **Coordinates Mode** (the other
existing interaction mode), the Recorder always records the interaction as raw `x, y`
coordinates (a W3C pointer `tap` action), because no element resolution happens in that code
path at all — see "Current behavior" below.

Users who want to record a sequence of taps quickly and expect an element-locator-based script
(so that the generated code doesn't break as soon as the screen layout shifts by a few pixels)
currently have to do the two-step Element Mode dance for every single tap, which is slow and easy
to get wrong (e.g. forgetting to click Tap after selecting).

## Current behavior (for reference)

Two screenshot interaction modes exist today (`app/common/renderer/constants/screenshot.js`,
`SCREENSHOT_INTERACTION_MODE`):

- **Element Mode** (`select`): clicking a highlighted element dispatches `selectElement(path)`
  (`app/common/renderer/actions/SessionInspector.js`), which resolves the element via Appium's
  `findElement` and populates the Selected Element panel. The user must then separately click
  **Tap** (`SelectedElementActions.jsx`), which calls
  `applyClientMethod({methodName: 'elementClick', elementId})`. Only this second step is
  recorded, and it *is* recorded with a locator (via `findAndAssign`), because the resolved
  element's `strategy`/`selector` are cached and returned by `InspectorDriver.executeMethod`
  (`app/common/renderer/lib/appium/inspector-driver.js`).
- **Coordinates Mode** (`tap_swipe`): clicking directly on the screenshot calls
  `applyClientMethod({methodName: 'tap', args: [...raw x,y pointer actions...]})`
  (`ScreenshotImgWithOverlays.jsx`, `handleTapOnScreenshot`). No element is ever resolved, so the
  Recorder always emits literal coordinates (see `codeFor_tap` in each
  `app/common/renderer/lib/client-frameworks/*.js`).

This split is intentional and documented (`docs/session-inspector/screenshot.md`,
`docs/session-inspector/header.md`), but there is no single-tap way to get locator-based
recording.

## Proposed UX

Add a third screenshot interaction mode, **Tap By Element**, alongside the existing **Select
Elements** and **Tap/Swipe By Coordinates** buttons in the Screenshot panel header
(`ScreenshotControls.jsx`).

- Visually and functionally identical to Element Mode for *hovering* (highlighter rectangles /
  centroids appear the same way).
- The difference is on **click**: instead of just selecting the element, clicking a highlighter
  immediately selects **and** taps it (`elementClick`) in one action — equivalent to the existing
  two-step Element Mode + Tap button flow, but collapsed into a single click.
- While [recording](https://appium.github.io/appium-inspector/latest/session-inspector/header/#toggle-recorder)
  is enabled, each tap is recorded exactly like the existing "Tap Element" button today: with a
  `findAndAssign` locator line plus the `elementClick` call, using whichever locator strategy the
  Inspector already suggests for that element (`getSuggestedLocators`).

## Known limitation / open question

For elements that overlap at the same screen coordinates, Element Mode currently shows a "+/-"
group handle that must be expanded before selecting one of the overlapping elements
individually (`ElementCentroid.jsx`, `RENDER_CENTROID_AS.EXPAND`). In Tap By Element mode, that
disambiguation step is unchanged (group handles still just expand/collapse); only the leaf
element handles (rect or individual centroid) trigger an immediate tap. This seems like the
right tradeoff (a group handle click can't unambiguously mean "tap", since the group represents
multiple elements), but it's worth confirming with maintainers before merging.

## Prior art / related discussion

- appium/appium-inspector#1342 ("XPath as a locator coming in recorded script") — confirms
  (maintainer comment) that Element Mode selection already picks whichever locator strategy the
  Inspector suggests, which this proposal reuses as-is.
- appium/appium-inspector#615 ("Recorder generated scripts do not fetch coordinates") — a past
  bug in the *coordinates* code-gen path, opposite direction from this request, but confirms
  Coordinates Mode has never done any element resolution.
- No existing open issue asks for this exact "one-tap, locator-based recording" flow as far as
  we could find via GitHub search.

## Rough implementation sketch

- `app/common/renderer/constants/screenshot.js`: add `TAP_ELEMENT: 'tap_element'` to
  `SCREENSHOT_INTERACTION_MODE`.
- `app/common/renderer/actions/SessionInspector.js`: extract the shared "resolve element via
  Appium findElement" logic out of the debounced `selectElement` internals into a reusable
  function, and add a new `tapElement(path)` action that resolves the element **immediately**
  (no debounce, since a tap is a deliberate one-shot action) and then calls
  `applyClientMethod({methodName: 'elementClick', elementId})`.
- `ScreenshotControls.jsx`: add a third mode button (e.g. `IconHandClick` from
  `@tabler/icons-react`) next to "Select Elements" and "Tap/Swipe By Coordinates".
- `ScreenshotImgWithOverlays.jsx`, `Overlays/ElementOverlays.jsx`, `Overlays/ElementRect.jsx`,
  `Overlays/ElementCentroid.jsx`: render the same highlighter overlays as Element Mode, but wire
  the click handler to `tapElement` instead of `selectElement`/`unselectElement` when the new
  mode is active.
- `app/common/public/locales/en/translation.json`: add the new tooltip string.
- Docs: `docs/session-inspector/screenshot.md` (Interaction Mode table) and
  `docs/session-inspector/header.md` (recordable interactions list).

A prototype implementing exactly this is available and passes `npm run test:lint`,
`npm run test:format`, and `npm run test:unit` (192/192); happy to open it as a draft PR for
discussion if the general direction sounds right.

tatanota:
Closes #3037

## Summary

Adds a new **Tap By Element** screenshot interaction mode, so that tapping an element
highlighter on the app screenshot immediately selects *and* taps the element
(`elementClick`) in one click, instead of requiring the existing two-step flow (switch to
Element Mode → click highlighter to select → click **Tap** in the Selected Element panel).

This gives the Recorder tab a fast, single-click way to produce locator-based code
(`driver.$('~someId').click()`-style) for a sequence of taps, as an alternative to the
existing **Tap/Swipe By Coordinates** mode, which always records raw `x, y` coordinates
because it never resolves an element.

Related discussion: see #3037 for the full "why" and prior art (#1342, #615).

## What changed

- `constants/screenshot.js`: added `TAP_ELEMENT: 'tap_element'` to
  `SCREENSHOT_INTERACTION_MODE`.
- `actions/SessionInspector.js`: extracted the element-selection prep (`prepareElementSelection`)
  and the Appium `findElement`-based resolution (`resolveElementId`) out of `selectElement` into
  shared helpers. `selectElement` keeps its existing debounced behavior unchanged. Added a new
  `tapElement(path)` action that resolves the element immediately (no debounce — a tap is a
  single deliberate action, not a browsing/hover selection) and then performs
  `applyClientMethod({methodName: 'elementClick', elementId})`, reusing the existing recording
  pipeline (`findAndAssign` + `RECORD_ACTION`) with zero changes to it.
- `components/.../Screenshot/ScreenshotControls.jsx`: added a third interaction-mode button
  ("Tap By Element", `IconHandClick`) alongside the existing "Select Elements" and
  "Tap/Swipe By Coordinates" buttons.
- `components/.../Screenshot/ScreenshotImgWithOverlays.jsx`: show the element highlighter
  overlays (and hide the raw-coordinates overlay) in the new mode too, same as Element Mode.
- `components/.../Screenshot/Overlays/ElementOverlays.jsx`,
  `Overlays/ElementRect.jsx`, `Overlays/ElementCentroid.jsx`: compute a mode-aware click handler
  — in Element Mode it keeps the existing select/unselect toggle behavior; in Tap By Element mode
  it calls `tapElement(path)` directly. Group ("+/-") handles for overlapping elements are
  unchanged (still just expand/collapse) in both modes.
- `public/locales/en/translation.json`: added the `"Tap By Element"` tooltip string (English
  only — other languages are managed via Crowdin).
- `docs/session-inspector/screenshot.md`, `docs/session-inspector/header.md`: documented the new
  mode alongside the existing two.

## Design notes

- Reused the existing element-resolution machinery (`findElement`/`getSuggestedLocators`/
  `InspectorDriver`) rather than trying to resolve an element from raw tap coordinates. This
  keeps the change additive and low-risk: the Recorder's `applyClientMethod` recording branch
  (`strategy && selector` → `findAndAssign`) and every client-framework code generator
  (`lib/client-frameworks/*.js`) needed **no changes**.
- `selectElement`'s existing 1-second debounce exists to avoid redundant Appium calls while a
  user browses several elements in a row in Element Mode. That debounce implementation
  (`utils/common.js` `debounce`) returns synchronously and doesn't resolve a promise tied to the
  actual (delayed) invocation, so simply `await`-ing the existing debounced `selectElement` would
  not reliably yield a usable `elementId` in time for a true "one click = one tap" experience.
  `tapElement` therefore calls the shared resolution logic directly (undebounced), and cancels
  any pending debounced call left over from Element Mode to avoid a race.
- Overlapping-element group handles ("+/-") are left as pure UI disambiguation in both modes;
  only leaf element handles trigger a tap. See the "Known limitation" note in the issue proposal.

## Test plan

- [x] `npm run test:lint` — 0 errors (pre-existing `jsdoc/require-jsdoc` warnings only, none new)
- [x] `npm run test:format` — passes (prettier-formatted the docs table after editing it)
- [x] `npm run test:unit` — 192/192 passing, no regressions
- [x] `npm run test:integration` / manual real-device verification — not run in this environment;
      recommended before merge: connect a real session, switch to **Tap By Element** mode, tap a
      few elements while recording is enabled, and confirm the Recorder tab emits
      `findAndAssign` + `elementClick` (locator-based) lines rather than raw coordinates, for at
      least one Android (uiautomator2) and one iOS (XCUITest) session.
- [x] Manual UX check for overlapping elements (group handle expand/collapse still works, and
      leaf handles tap correctly) once expanded.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/common/renderer/components/SessionInspector/Screenshot/ScreenshotControls.jsx
+++ b/app/common/renderer/components/SessionInspector/Screenshot/ScreenshotControls.jsx
@@ -2,6 +2,7 @@ import {
   IconCrosshair,
   IconDownload,
   IconEyePlus,
+  IconHandClick,
   IconMovie,
   IconObjectScan,
   IconPhoto,
@@ -14,7 +15,7 @@ import {SCREENSHOT_INTERACTION_MODE} from '../../../constants/screenshot.js';
 import {downloadFile} from '../../../utils/file-handling.js';
 import styles from './Screenshot.module.css';
 
-const {SELECT, TAP_SWIPE} = SCREENSHOT_INTERACTION_MODE;
+const {SELECT, TAP_SWIPE, TAP_ELEMENT} = SCREENSHOT_INTERACTION_MODE;
 
 const downloadScreenshot = (screenshot) => {
   const href = `data:image/png;base64,${screenshot}`;
@@ -85,8 +86,8 @@ const ToggleElementHandlesButton = ({
 };
 
 /**
- * Button allowing to switch between the Element Mode and Coordinates Mode
- * when interacting with the screenshot.
+ * Button allowing to switch between the Element Mode, Tap By Element Mode, and
+ * Coordinates Mode when interacting with the screenshot.
  */
 const ScreenshotInteractionModeControls = ({
   screenshotInteractionMode,
@@ -108,6 +109,14 @@ const ScreenshotInteractionModeControls = ({
           icon={<IconObjectScan size={18} />}
           onClick={() => screenshotInteractionChange(SELECT)}
           type={screenshotInteractionMode === SELECT ? BUTTON.PRIMARY : BUTTON.DEFAULT}
+          disabled={isGestureEditorVisible}
+        />
+      </Tooltip>
+      <Tooltip title={t('Tap By Element')}>
+        <Button
+          icon={<IconHandClick size={18} />}
+          onClick={() => screenshotInteractionChange(TAP_ELEMENT)}
+          type={screenshotInteractionMode === TAP_ELEMENT ? BUTTON.PRIMARY : BUTTON.DEFAULT}
           disabled={isGestureEditorVisible}
         />
       </Tooltip>
--- a/app/common/renderer/components/SessionInspector/Screenshot/ScreenshotControls.jsx
+++ b/app/common/renderer/components/SessionInspector/Screenshot/ScreenshotControls.jsx
@@ -2,10 +2,10 @@ import {
   IconCrosshair,
   IconDownload,
   IconEyePlus,
-  IconHandClick,
   IconMovie,
   IconObjectScan,
   IconPhoto,
+  IconSquarePlus,
 } from '@tabler/icons-react';
 import {Button, Space, Tooltip} from 'antd';
 import {useTranslation} from 'react-i18next';
@@ -114,7 +114,7 @@ const ScreenshotInteractionModeControls = ({
       </Tooltip>
       <Tooltip title={t('Tap By Element')}>
         <Button
-          icon={<IconHandClick size={18} />}
+          icon={<IconSquarePlus size={18} />}
           onClick={() => screenshotInteractionChange(TAP_ELEMENT)}
           type={screenshotInteractionMode === TAP_ELEMENT ? BUTTON.PRIMARY : BUTTON.DEFAULT}
           disabled={isGestureEditorVisible}
@@ -171,7 +171,7 @@ const ScreenshotControls = (props) => {
 
   return (
     <div className={styles.screenshotControls}>
-      <Space size="middle">
+      <Space size="small">
         {serverDetails.mjpegScreenshotUrl !== null && (
           <ScreenshotCaptureModeControls
             setMjpegState={setMjpegState}
--- a/docs/session-inspector/header.md
+++ b/docs/session-inspector/header.md
@@ -131,6 +131,7 @@ and translate them into code that can be used with various [Appium clients](http
 Interactions that can be recorded include:
 
 - Actions for a specific element (tap/send keys/clear)
+- Tapping an element directly on the screenshot, in [Tap By Element Mode](./screenshot.md#interaction-mode)
 - Generic tap/swipe actions on the application screenshot
 - [Mobile device system actions](#device-system-buttons)
 - [Driver commands](./commands.md)
--- a/docs/session-inspector/header.md
+++ b/docs/session-inspector/header.md
@@ -131,7 +131,6 @@ and translate them into code that can be used with various [Appium clients](http
 Interactions that can be recorded include:
 
 - Actions for a specific element (tap/send keys/clear)
-- Tapping an element directly on the screenshot, in [Tap By Element Mode](./screenshot.md#interaction-mode)
 - Generic tap/swipe actions on the application screenshot
 - [Mobile device system actions](#device-system-buttons)
 - [Driver commands](./commands.md)
--- a/app/common/public/locales/en/translation.json
+++ b/app/common/public/locales/en/translation.json
@@ -79,6 +79,7 @@
   "Show Element Handles": "Show Element Handles",
   "Hide Element Handles": "Hide Element Handles",
   "Select Elements": "Select Elements",
+  "Tap By Element": "Tap By Element",
   "Tap/Swipe By Coordinates": "Tap/Swipe By Coordinates",
   "Download Screenshot": "Download Screenshot",
   "Back": "Back",
--- a/app/common/renderer/actions/SessionInspector.js
+++ b/app/common/renderer/actions/SessionInspector.js
@@ -120,8 +120,42 @@ export const SET_AUTO_SESSION_RESTART = 'SET_AUTO_SESSION_RESTART';
 const KEEP_ALIVE_PING_INTERVAL = 20 * 1000;
 const NO_NEW_COMMAND_LIMIT = 24 * 60 * 60 * 1000; // Set timeout to 24 hours
 
-// A debounced function that calls findElement and gets info about the element
-const findElement = debounce(async function (strategyMap, dispatch, getState, path) {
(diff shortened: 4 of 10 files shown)


app/common/renderer/components/SessionInspector/Screenshot/ScreenshotControls.jsx
```diff
@@ -111,6 +112,14 @@ const ScreenshotInteractionModeControls = ({
           disabled={isGestureEditorVisible}
         />
       </Tooltip>
+      <Tooltip title={t('Tap By Element')}>
+        <Button
+          icon={<IconHandClick size={18} />}
```
eglitise: I'm not sure about this icon, since it can mean any form of tap. It's a bit tricky to find a suitable icon that would also match up with the other two - but how about `IconSquarePlus`?

app/common/renderer/components/SessionInspector/Screenshot/ScreenshotControls.jsx
eglitise: The new button makes the screenshot control block quite wide, so I would suggest reducing the padding:
```suggestion
      <Space size="small">
```

docs/session-inspector/header.md
```diff
@@ -131,6 +131,7 @@ and translate them into code that can be used with various [Appium clients](http
 Interactions that can be recorded include:
 
 - Actions for a specific element (tap/send keys/clear)
+- Tapping an element directly on the screenshot, in [Tap By Element Mode](./screenshot.md#interaction-mode)
```
eglitise: This line can be omitted since the previous one already infers it

app/common/renderer/components/SessionInspector/Screenshot/ScreenshotControls.jsx
tatanota: Thanks for the suggestion — `IconSquarePlus` fits much better here. Swapped it in 39d4528. Verified visually by connecting the Inspector to a real Android emulator via a local Appium server.

app/common/renderer/components/SessionInspector/Screenshot/ScreenshotControls.jsx
tatanota: Good catch, thanks! Applied your suggestion (`size="small"`) in 39d4528 — the control block sits much tighter now. Confirmed visually against a real Android emulator session.

docs/session-inspector/header.md
tatanota: You're right, that's redundant — removed in 39d4528. Thanks for the thorough review!

eglitise (approved):
Thanks!

KazuCocoa:
Hi @tatanota, congrats: the Appium project wants to compensate you for this contribution! Please reply to this comment mentioning @jlipps and @KazuCocoa and sharing your [OpenCollective](https://opencollective.com/) account name, so that we can initiate payment! Or let us know if you decline to receive compensation via OpenCollective. Thank you!

tatanota:
@jlipps and @KazuCocoa,

Thank you for the wonderful news!

I've created an OpenCollective account. My account name is @tatanota.

I'm grateful for the opportunity to receive compensation for my contribution.

PR #3040, opened 2026-07-24
refactor: merge TS support into JS - WebdriverIO recorder

### LINKED ISSUE #3038: Feature Proposal: Add TypeScript (WebdriverIO) as a Recorder output language
# Feature Proposal: Add TypeScript (WebdriverIO) as a Recorder output language

## Is your feature request related to a problem? Please describe.

The Recorder tab currently offers 8 client-framework/language options for generated code
(`.NET NUnit`, `JS - WebdriverIO`, `JS - Oxygen`, `Java - JUnit4`, `Java - JUnit5`, `Python`,
`Robot Framework`, `Ruby`), defined in
[`CLIENT_FRAMEWORKS`](https://github.com/appium/appium-inspector/blob/main/app/common/renderer/constants/session-inspector.js)
and registered in
[`CLIENT_FRAMEWORK_MAP`](https://github.com/appium/appium-inspector/blob/main/app/common/renderer/lib/client-frameworks/map.js).

There is no TypeScript option, even though WebdriverIO (the JS framework already supported) is
commonly used with TypeScript in real projects, and the WebdriverIO team's own documentation and
starter templates default to TypeScript. Users who work in TS projects currently have to
copy the JS output and manually adapt it (import syntax is fine as-is, but the boilerplate/run
instructions target plain Node with `.js` files).

I searched existing issues/PRs for "TypeScript" and found no prior request or rejected proposal
for this — this looks like an open gap rather than a deliberately-omitted feature.

## Describe the solution you'd like

Add a new `TS - WebdriverIO` option to the Recorder language dropdown, implemented as a thin
subclass of the existing `JsWdIoFramework` (same pattern used for `Java - JUnit5`, which
subclasses `Java - JUnit4`'s shared base class — see
[#1181](https://github.com/appium/appium-inspector/pull/1181)).

Since the WebdriverIO JS API calls generated by `JsWdIoFramework`'s `codeFor_*` methods
(`driver.$()`, `el.click()`, etc.) are already valid TypeScript syntax, the new class only needs
to override:

- `static readableName` → `'TS - WebdriverIO'`
- `static refractorLang` / `static refractorLib` → use `refractor/typescript` for syntax
  highlighting (already bundled with the `react-refractor` dependency this project uses)
- `wrapWithBoilerplate()` → TS-appropriate run instructions (e.g. `npx tsx <file>.ts`) instead of
  plain `node <file>.js`

No changes are required to the Recorder UI itself — the language dropdown is built dynamically
from `CLIENT_FRAMEWORK_MAP`, so registering the new class is sufficient.

## Describe alternatives you've considered

- Asking users to manually adapt the JS output — works, but adds friction for a common client
  setup (WebdriverIO + TS) and doesn't surface TS-specific run instructions.
- A completely separate implementation duplicating all `codeFor_*` methods — unnecessary, since
  the JS-generated command bodies are valid TS as-is; only the boilerplate wrapper differs.

## Additional context

I'd like to submit a PR implementing this (new file
`app/common/renderer/lib/client-frameworks/js-wdio-ts.js`, plus a few lines in `map.js` and
`constants/session-inspector.js`). Filing this issue first per the contribution guidelines /
CONTRIBUTING.md before opening the PR — happy to adjust the readable name or scope based on
maintainer feedback (e.g. whether type annotations should go further, such as importing
`Capabilities.WebdriverIOConfig` from `@wdio/types` for the connection options, which I found is
not re-exported by the top-level `webdriverio` package itself).

tatanota:
# refactor: merge TS support into JS - WebdriverIO recorder

Closes #3038

## What

This PR updates `JsWdIoFramework` (`app/common/renderer/lib/client-frameworks/js-wdio.js`) to
explicitly document support for both JavaScript and TypeScript, instead of introducing a separate
`JsWdIoTsFramework` class (the approach in the previous version of this PR).

Changes:

- `static readableName`: `'JS - WebdriverIO'` → `'JS/TS - WebdriverIO'`
- `wrapWithBoilerplate()` run instructions now mention both `.js` and `.ts` files (modern Node can
  run TypeScript directly when types are strippable)
- `static refractorLang` / `static refractorLib` are unchanged (`'js'` / `refractor/javascript`)
- Removed `js-wdio-ts.js`, and the now-unused `JS_WDIO_TS` references in
  `constants/session-inspector.js` and `lib/client-frameworks/map.js`

## Why

@eglitise reviewed the original approach and pointed out that the generated code was identical
between the JS and TS versions, and suggested editing the existing JS recorder's strings rather
than maintaining a separate class:

> Overall, I don't see the benefits of a separate TS recorder, so I would suggest simply modifying
> the few JS-specific strings in the JS recorder, so that it's clear that the code applies to both
> JS and TS.

This PR implements that suggestion.

## Test plan

- [x] `npm run test:lint` — 0 errors (pre-existing jsdoc warnings only, unrelated to this change)
- [x] `npm run test:format` (prettier -c) — passes
- [x] `npm run test:unit` — 192/192 passing
- [x] Manually verified in the running Inspector app: started an Android emulator + local Appium
      server, recorded a session in the browser-based Inspector, and confirmed the Recorder tab's
      language dropdown shows `JS/TS - WebdriverIO` with the updated run instructions
- [x] Took the actual recorded output (e.g. `driver.$("accessibility id:Chrome")`) embedded in the
      updated `wrapWithBoilerplate()` template and type-checked it with `tsc --noEmit --strict`
      against `webdriverio`'s bundled types, for both `.ts` and `.mts` extensions — no errors

### DIFF
--- a/app/common/renderer/lib/client-frameworks/js-wdio.js
+++ b/app/common/renderer/lib/client-frameworks/js-wdio.js
@@ -3,15 +3,17 @@ import refractorJs from 'refractor/javascript';
 import CommonClientFramework from './common.js';
 
 export default class JsWdIoFramework extends CommonClientFramework {
-  static readableName = 'JS - WebdriverIO';
+  static readableName = 'JS/TS - WebdriverIO';
   static refractorLang = 'js';
   static refractorLib = refractorJs;
 
   wrapWithBoilerplate(code) {
     return `// This sample code supports WebdriverIO client >=9.7.0
 // (npm i --save webdriverio)
-// Then paste this into a .js file and run with Node:
+// Then paste this into a .js or .ts file and run with Node
+// (modern Node can run TypeScript files directly if types are strippable):
 // node <file>.js
+// or: node <file>.ts
 
 import {remote} from 'webdriverio';
 async function main () {


eglitise (approved):
Thanks!

PR #3063, opened 2026-07-27
fix: make the New Session screen responsive on narrow windows

### LINKED ISSUE #3061: bug: New Session screen fields overlap / become unreadable on narrow windows

## Current Behavior

In the browser build (`npm run dev:browser`), resizing the window narrower than ~870px breaks the **New Session** screen layout:

- The Remote Host / Remote Port / Remote Path row (`ServerTabCustom.jsx`) uses a percentage-based `Row`/`Col` grid with non-wrapping `Space.Addon` labels. Below a certain width, the "Remote Port" and "Remote Path" labels visually overlap and become unreadable, since neither the columns nor the labels shrink or wrap.
- The Capability Builder / JSON Representation panels (`CapabilityEditor.jsx`) sit side-by-side via `Splitter` with no responsive fallback, so the JSON preview just becomes unusably narrow instead of adapting.
- This is mostly hidden in the Electron desktop build, since the native window has a hardcoded `minWidth`/`minHeight` (`app/electron/main/windows.js`) that prevents the window from ever getting this narrow. The *browser* build has no equivalent constraint - the only thing enforcing a floor is `body { min-width: 870px; min-height: 610px; }` in `main.css`, and it does so by simply refusing to shrink the page rather than by adapting the layout. That's a problem on its own (e.g. a narrow browser tab, embedding the inspector in an iframe, or browser zoom can all still produce an effectively-narrower viewport than the body's forced minimum, causing the very overlap this rule was presumably meant to avoid).

## Expected Behavior

The New Session screen should adapt to narrow widths instead of overlapping:
- Remote Port should have a fixed, small width (just enough for a 4-5 digit port number); Remote Port / Remote Path / SSL should wrap onto a new line below Remote Host once space runs out, instead of overlapping.
- The JSON Representation panel should stack below the Capability Builder panel once the window is too narrow for them to sit side-by-side.

## Operating System

Mac (this is a CSS/layout issue, not OS-specific - reproducible on any OS running the browser build)

## Appium Inspector Version

2026.7.1 (browser build)

## Further Information

I have a fix ready and will open a PR shortly.

tatanota:
## What

Fixes the New Session screen's layout so it adapts to narrow windows instead of overlapping, in the browser build.

- Removed the `body` `min-width`/`min-height` floor in `main.css`. It's redundant for the Electron build (which enforces its own native window `minWidth`/`minHeight`), and actively harmful for the browser build (nothing prevents the actual viewport from going narrower than that forced floor, e.g. via a narrow browser tab, embedding, or zoom - and the CSS-mandated body size did nothing but prevent the page from shrinking to match).
- `ServerTabCustom.jsx` / `ServerDetails.module.css`: replaced the percentage-based `Row`/`Col` grid for Remote Host/Port/Path/SSL with a `flex-wrap` layout. Remote Port is now a small fixed-width field (fits a 4-5 digit port); Remote Port/Remote Path/SSL are grouped so they wrap together onto a new line below Remote Host once space runs out.
- `CapabilityEditor.jsx`: the Capability Builder / JSON Representation `Splitter` now switches to a vertical layout below a width breakpoint (700px), so the JSON preview stacks below the builder instead of becoming unreadably narrow. The `Splitter` is remounted (`key`) on each orientation change - without this it retains stale panel sizes from the previous orientation, so resizing the window back and forth could leave it stuck showing a shrunken panel even at a wide width.

Closes #3061

## Test plan

- [x] `npm run test:lint` / `test:format` / `test:unit` all pass (192/192)
- [x] Manually verified in the browser build (`npm run dev:browser`) at a range of widths (1200 → 320px, both English and Japanese locale) that:
  - Remote Host/Port/Path/SSL no longer overlap at any width, and wrap as expected below ~700px
  - The JSON Representation panel stacks below the Capability Builder panel below the same breakpoint
  - Repeatedly narrowing/widening the window past the breakpoint no longer leaves either panel stuck at a stale size

### DIFF
--- a/app/common/renderer/components/SessionBuilder/CapabilityBuilderTab/CapabilityEditor.jsx
+++ b/app/common/renderer/components/SessionBuilder/CapabilityBuilderTab/CapabilityEditor.jsx
@@ -12,14 +12,19 @@ import {
   Splitter,
   Tooltip,
 } from 'antd';
-import {useEffect, useRef} from 'react';
+import {useEffect, useRef, useState} from 'react';
 import {useTranslation} from 'react-i18next';
 
 import {CAPABILITY_TYPES} from '../../../constants/session-builder.js';
+import {debounce} from '../../../utils/common.js';
 import CapabilityJSON from '../CapabilityJSON/CapabilityJSON.jsx';
 import builderStyles from '../SessionBuilder.module.css';
 import styles from './CapabilityBuilderTab.module.css';
 import CapabilityControl from './CapabilityControl.jsx';
+
+// Below this window width, the JSON preview panel no longer has enough room
+// to sit beside the capability builder panel, so it wraps below it instead.
+const NARROW_LAYOUT_BREAKPOINT = 700;
 
 const whitespaces = /^\s|\s$/;
 
@@ -84,6 +89,19 @@ const CapabilityEditor = (props) => {
   const latestCapFieldRef = useRef(null);
   const {t} = useTranslation();
 
+  const [isNarrow, setIsNarrow] = useState(window.innerWidth < NARROW_LAYOUT_BREAKPOINT);
+  useEffect(() => {
+    const updateIsNarrow = debounce(
+      () => setIsNarrow(window.innerWidth < NARROW_LAYOUT_BREAKPOINT),
+      100,
+    );
+    window.addEventListener('resize', updateIsNarrow);
+    return () => {
+      window.removeEventListener('resize', updateIsNarrow);
+      updateIsNarrow.cancel();
+    };
+  }, []);
+
   // if we have more than one cap and the most recent cap name is empty,
   // it means we've just added a new cap field, so focus that input element
   useEffect(() => {
@@ -93,7 +111,12 @@ const CapabilityEditor = (props) => {
   }, [caps.length, latestCapFieldRef]);
 
   return (
-    <Splitter>
+    // Remounting on orientation change avoids Splitter retaining stale panel
+    // sizes/state from the previous layout when switching back and forth.
+    <Splitter
+      key={isNarrow ? 'vertical' : 'horizontal'}
+      layout={isNarrow ? 'vertical' : 'horizontal'}
+    >
       <Splitter.Panel collapsible resizable={false}>
         <Form className={styles.newSessionForm}>
           {caps.map((cap, index) => (
--- a/app/common/renderer/components/SessionBuilder/CapabilityBuilderTab/CapabilityEditor.jsx
+++ b/app/common/renderer/components/SessionBuilder/CapabilityBuilderTab/CapabilityEditor.jsx
@@ -16,7 +16,6 @@ import {useEffect, useRef, useState} from 'react';
 import {useTranslation} from 'react-i18next';
 
 import {CAPABILITY_TYPES} from '../../../constants/session-builder.js';
-import {debounce} from '../../../utils/common.js';
 import CapabilityJSON from '../CapabilityJSON/CapabilityJSON.jsx';
 import builderStyles from '../SessionBuilder.module.css';
 import styles from './CapabilityBuilderTab.module.css';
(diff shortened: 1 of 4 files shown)


eglitise:
Thanks. While this does improve responsiveness, it's worth highlighting that this is only a partial fix (components like cloud provider tabs and the Saved Capability Sets tab still have issues). Still, useful to have.

app/common/renderer/components/SessionBuilder/CapabilityBuilderTab/CapabilityEditor.jsx
```diff
@@ -93,7 +111,12 @@ const CapabilityEditor = (props) => {
   }, [caps.length, latestCapFieldRef]);
 
   return (
-    <Splitter>
+    // Remounting on orientation change avoids Splitter retaining stale panel
+    // sizes/state from the previous layout when switching back and forth.
+    <Splitter
+      key={isNarrow ? 'vertical' : 'horizontal'}
+      layout={isNarrow ? 'vertical' : 'horizontal'}
```
eglitise: `key` isn't needed and `layout` is deprecated and replaced with `orientation`

app/common/renderer/components/SessionBuilder/CapabilityBuilderTab/CapabilityEditor.jsx
```diff
@@ -93,7 +111,12 @@ const CapabilityEditor = (props) => {
   }, [caps.length, latestCapFieldRef]);
 
   return (
-    <Splitter>
+    // Remounting on orientation change avoids Splitter retaining stale panel
```
eglitise: I was testing this and noticed that when resizing from narrow to wide window size, the splitter orientation changes as expected, but the inner panel widths remain narrow (specifically, their `flex-basis` stays the same), and require a separate second resize action to update. Can this be fixed?

app/common/renderer/components/SessionBuilder/ServerDetails/ServerDetails.module.css
```diff
@@ -2,6 +2,33 @@
...
+}
+
+.serverHostGroup {
+  flex: 1 1 240px;
+}
+
+.serverPortPathGroup {
+  display: flex;
+  flex: 2 1 320px;
+  align-items: center;
```
eglitise: `align-items` doesn't seem to affect anything here so it can be removed

app/common/renderer/components/SessionBuilder/ServerDetails/ServerDetails.module.css
```diff
@@ -2,6 +2,33 @@
...
+
+.serverPortInput {
+  width: 4.5em;
+  flex: 0 0 auto;
+}
+
+.serverPathCompact {
+  flex: 1 1 auto;
+  min-width: 120px;
+}
```
eglitise: Instead of setting a width for the whole label+input, where the label width can scale depending on the language, it's better to set the min-width for the input itself (consistent with `serverPortInput`). A value of `50px` seemed fine in my testing.

app/common/renderer/components/SessionBuilder/ServerDetails/ServerDetails.module.css
```diff
@@ -2,6 +2,33 @@
...
+.serverPortPathGroup {
+  display: flex;
+  flex: 2 1 320px;
+  align-items: center;
+  gap: 8px;
+}
+
+.serverPortInput {
+  width: 4.5em;
+  flex: 0 0 auto;
```
eglitise: A width of `5em` looks better at 5 digits, and the `flex` doesn't seem to be necessary

app/common/renderer/components/SessionBuilder/CapabilityBuilderTab/CapabilityEditor.jsx
tatanota: Done — swapped to `orientation` and dropped the `key`.

app/common/renderer/components/SessionBuilder/ServerDetails/ServerDetails.module.css
tatanota: Removed.

app/common/renderer/components/SessionBuilder/ServerDetails/ServerDetails.module.css
tatanota: Good point — moved to `.serverPathInput` on the input itself, matching `serverPortInput`.

app/common/renderer/components/SessionBuilder/ServerDetails/ServerDetails.module.css
tatanota: Done, and dropped the flex override too.

app/common/renderer/components/SessionBuilder/CapabilityBuilderTab/CapabilityEditor.jsx
tatanota: Good catch, thanks for testing that. Root cause: the `isNarrow` resize listener was debounced by 100ms, so `Splitter`'s own `ResizeObserver` could fire and measure the wrong axis (still using the stale orientation) before the debounced update landed, caching a bad container size. Removed the debounce (a plain width comparison doesn't need it) so the orientation stays in sync with the actual DOM resize — confirmed with a script that jumps the viewport straight from narrow to wide in one step, no intermediate resize needed to correct it anymore.

app/common/renderer/components/SessionBuilder/ServerDetails/ServerDetails.module.css
eglitise: `serverPathCompact` doesn't seem to have any effect now, so it can be removed as a whole

tatanota: Removed.

eglitise (approved):
Thank you!

PR #3064, opened 2026-07-27
fix: make the Session Inspector screen responsive on narrow windows

### LINKED ISSUE #3062: bug: Session Inspector screen (header, screenshot toolbar, source/selected element panels) overlaps on narrow windows

## Current Behavior

Same root cause as #3061 - in the browser build, the only defense against narrow windows is a hard `body { min-width: 870px; min-height: 610px; }` floor in `main.css`, which doesn't reflow anything, it just refuses to shrink the page. Below that width (narrow browser tab, embedding, zoom, etc.), several areas of the post-session **Session Inspector** screen break:

- The header button row (`HeaderButtons.jsx` / `Header.module.css`) is centered (`align-items: center`) and neither wraps nor scrolls. Once its content is wider than the window, it overflows symmetrically on both sides - the left-side icons become permanently unreachable, since browsers don't allow negative `scrollLeft`.
- The Screenshot panel's toolbar (`ScreenshotControls.jsx`) sits in a container (`.screenshotContainer`) that can shrink to near-zero width (no `min-width` set), but the toolbar's own button row doesn't shrink with it and has no overflow handling, so the buttons visually render on top of the adjacent Source/Selected Element panel.
- In the **Source tab**, the App Source tree and Selected Element panel are laid out side-by-side via `Splitter` (`SourceTab.jsx`) with no responsive fallback. There's also no good way to temporarily see more of one panel without permanently losing the other - `Splitter`'s own collapse control only becomes visible on hover over a near-invisible divider line, so it's very easy to miss.

## Expected Behavior

- The header buttons and the screenshot toolbar should shrink or become scrollable instead of overlapping / becoming unreachable.
- On narrow windows, App Source and Selected Element should stack vertically instead of squeezing side-by-side.
- There should be an easy, discoverable way to collapse one of the stacked panels (keeping its header visible, so it's clear it's still there) so the other one is easier to read.

## Operating System

Mac (this is a CSS/layout issue, not OS-specific - reproducible on any OS running the browser build)

## Appium Inspector Version

2026.7.1 (browser build)

## Further Information

I have a fix ready and will open a PR shortly.

tatanota:
## What

Fixes the post-session Session Inspector screen's layout so it adapts to narrow windows instead of overlapping, in the browser build.

- Removed the `body` `min-width`/`min-height` floor in `main.css` (same reasoning as #3063 - redundant for Electron, harmful for the browser build).
- **Header buttons row**: now scrolls (`overflow-x: auto`) and uses `align-items: safe center` instead of `center`. Previously, once the button row was wider than the window, it overflowed symmetrically on both sides and the left-side icons became permanently unreachable (browsers don't allow negative `scrollLeft`).
- **Screenshot toolbar**: the screenshot container now has a sane `min-width` instead of collapsing to near-zero width, and the main row (`screenshot | source/selected-element`) scrolls horizontally as a fallback. Previously the container could shrink so far that its toolbar buttons rendered on top of the adjacent panel.
- **Source tab (App Source / Selected Element)**: the `Splitter` now switches to a vertical layout below a 700px width breakpoint, stacking the panels instead of squeezing them side-by-side. It's remounted (`key`) on each orientation change, since `Splitter` otherwise retains stale panel sizes from the previous orientation.
- **Stacked-layout collapse**: each panel's card gets its own collapse toggle when stacked, which hides only the body while keeping the header visible (`Splitter`'s own collapse hides the header too, which isn't useful for an accordion-style layout). The collapsed state is lifted to `SourceTab`, which also drives the `Splitter.Panel`'s `size`, so the freed space goes to the other panel. Collapsing one panel always re-expands the other first, since `Splitter` can't shrink both panels to their collapsed size simultaneously without a panel left over to absorb the freed space (confirmed by testing - without this, collapsing both panels at once silently resets both back to their default size).
- Also made the horizontal-layout `Splitter`'s own collapse arrows always visible (`showCollapsibleIcon: true`) instead of hover-only - the divider was otherwise a near-invisible line, easy to miss.

Closes #3062

## Test plan

- [x] `npm run test:lint` / `test:format` / `test:unit` all pass (192/192)
- [x] Manually verified in the browser build (`npm run dev:browser`), connected to a real Android emulator via a local Appium server, at a range of widths (1200 → 320px, both English and Japanese locale):
  - Header buttons and screenshot toolbar no longer overlap adjacent content at any width, and remain fully reachable via scroll
  - App Source / Selected Element stack vertically below 700px, each keeping its header visible when its own collapse toggle is used
  - Collapsing one stacked panel correctly grows the other to fill the freed space; collapsing both in sequence re-expands the first instead of resetting both to their default size
  - Repeatedly narrowing/widening past the breakpoint doesn't leave anything stuck at a stale size

### DIFF
--- a/app/common/renderer/components/SessionInspector/Header/Header.module.css
+++ b/app/common/renderer/components/SessionInspector/Header/Header.module.css
@@ -5,11 +5,12 @@
 .headerButtons {
   display: flex;
   flex-direction: column;
-  align-items: center;
+  align-items: safe center;
   padding: 10px 0;
   gap: 10px;
   width: 100%;
   height: 50px;
+  overflow-x: auto;
   -webkit-app-region: drag;
 }
 
--- a/app/common/renderer/components/SessionInspector/Header/Header.module.css
+++ b/app/common/renderer/components/SessionInspector/Header/Header.module.css
@@ -9,8 +9,7 @@
   padding: 10px 0;
   gap: 10px;
   width: 100%;
-  height: 50px;
-  overflow-x: auto;
+  min-height: 50px;
   -webkit-app-region: drag;
 }
 
--- a/app/common/renderer/components/SessionInspector/Header/Header.module.css
+++ b/app/common/renderer/components/SessionInspector/Header/Header.module.css
@@ -13,6 +13,10 @@
   -webkit-app-region: drag;
 }
 
+.headerButtonsSpace {
+  justify-content: center;
+}
+
 .headerButtons :global(.ant-divider-horizontal) {
   margin: 0px;
 }
--- a/app/common/renderer/components/SessionInspector/SourceTab/AppSource/AppSourceCard.jsx
+++ b/app/common/renderer/components/SessionInspector/SourceTab/AppSource/AppSourceCard.jsx
@@ -32,9 +32,12 @@ const AppSourcePanelTitle = () => {
 };
 
 /**
- * Header action buttons for source XML copy and download.
+ * Header action buttons for source XML copy and download, plus (when panels
+ * are stacked) the toggle that collapses/expands the card body while keeping
+ * its header visible - the side-by-side layout already offers a full
+ * collapse via the Splitter divider.
  */
-const AppSourceHeaderButtons = ({sourceXML}) => {
+const AppSourceHeaderButtons = ({sourceXML, collapsible, collapsed, onToggleCollapse}) => {
   const {t} = useTranslation();
 
   return (
@@ -55,27 +58,16 @@ const AppSourceHeaderButtons = ({sourceXML}) => {
           onClick={() => downloadXML(sourceXML)}
         />
       </Tooltip>
+      {collapsible && (
+        <Tooltip title={t(collapsed ? 'Expand Panel' : 'Collapse Panel')}>
+          <Button
+            type="text"
+            icon={collapsed ? <IconChevronDown size={18} /> : <IconChevronUp size={18} />}
+            onClick={onToggleCollapse}
+          />
+        </Tooltip>
+      )}
     </span>
-  );
-};
-
-/**
- * Toggle button that collapses/expands the card body while keeping its
- * header visible. Only shown when panels are stacked (narrow layout),
- * since the side-by-side layout already offers a full collapse via the
- * Splitter divider.
- */
-const CollapseToggleButton = ({collapsed, onClick}) => {
-  const {t} = useTranslation();
-
-  return (
-    <Tooltip title={t(collapsed ? 'Expand Panel' : 'Collapse Panel')}>
-      <Button
-        type="text"
-        icon={collapsed ? <IconChevronDown size={18} /> : <IconChevronUp size={18} />}
-        onClick={onClick}
-      />
-    </Tooltip>
   );
 };
 
@@ -86,10 +78,12 @@ const AppSourceCard = ({sourceXML, collapsible, collapsed, onToggleCollapse, chi
   <Card
     title={<AppSourcePanelTitle />}
     extra={
-      <span>
-        <AppSourceHeaderButtons sourceXML={sourceXML} />
-        {collapsible && <CollapseToggleButton collapsed={collapsed} onClick={onToggleCollapse} />}
-      </span>
+      <AppSourceHeaderButtons
+        sourceXML={sourceXML}
+        collapsible={collapsible}
+        collapsed={collapsed}
+        onToggleCollapse={onToggleCollapse}
+      />
     }
   >
     {!collapsed && children}
--- a/app/common/renderer/components/SessionInspector/SourceTab/AppSource/AppSourceCard.jsx
+++ b/app/common/renderer/components/SessionInspector/SourceTab/AppSource/AppSourceCard.jsx
@@ -1,4 +1,10 @@
-import {IconDownload, IconFiles, IconFileText} from '@tabler/icons-react';
+import {
+  IconChevronDown,
+  IconChevronUp,
+  IconDownload,
+  IconFiles,
+  IconFileText,
+} from '@tabler/icons-react';
 import {Button, Card, Flex, Tooltip} from 'antd';
 import {useTranslation} from 'react-i18next';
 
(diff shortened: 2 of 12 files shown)


app/common/renderer/components/SessionInspector/Header/Header.module.css
```diff
@@ -5,11 +5,12 @@
 .headerButtons {
   display: flex;
   flex-direction: column;
-  align-items: center;
+  align-items: safe center;
   padding: 10px 0;
   gap: 10px;
   width: 100%;
   height: 50px;
+  overflow-x: auto;
```
eglitise: While this does make the buttons reachable on narrow screens, it also results in the `<Divider>` element below the buttons to no longer be visible, thereby adding a small vertical scrollbar.
I think a better approach would be to just move the overflowing buttons to a new second row, thereby avoiding horizontal scrolling as well.

app/common/renderer/components/SessionInspector/SourceTab/SourceTab.jsx
```diff
@@ -1,25 +1,103 @@
...
 
+  const [isNarrow, setIsNarrow] = useState(window.innerWidth < NARROW_LAYOUT_BREAKPOINT);
+  useEffect(() => {
+    // Deliberately not debounced: Splitter measures its container via its
+    // own ResizeObserver, which can fire before a debounced update here
+    // would land. If that happens while this is still reporting the old
+    // orientation, Splitter reads the wrong axis (width vs height) and
+    // caches a stale container size, leaving panels stuck at the wrong
+    // size until another resize happens to nudge it again.
+    const updateIsNarrow = () => setIsNarrow(window.innerWidth < NARROW_LAYOUT_BREAKPOINT);
```
eglitise: Unlike the Session Builder screen, the collapsible elements share horizontal space with the screenshot, which can have different sizes (such as portrait/landscape orientation). It's better to have the breakpoint depend on the width of the Source tab alone, rather than the whole window.

app/common/renderer/components/SessionInspector/SourceTab/AppSource/AppSourceCard.jsx
```diff
@@ -53,12 +59,40 @@ const AppSourceHeaderButtons = ({sourceXML}) => {
   );
 };
 
+/**
+ * Toggle button that collapses/expands the card body while keeping its
+ * header visible. Only shown when panels are stacked (narrow layout),
+ * since the side-by-side layout already offers a full collapse via the
+ * Splitter divider.
+ */
+const CollapseToggleButton = ({collapsed, onClick}) => {
```
eglitise: This tooltip/button can be merged into `<AppSourceHeaderButtons>`. Same for `SelectedElementCard`

app/common/renderer/components/SessionInspector/SourceTab/SourceTab.jsx
```diff
@@ -1,25 +1,103 @@
...
+            : false
+        }
+        size={
+          canAccordionCollapse && appSourceCollapsed
+            ? PANEL_HEADER_HEIGHT
+            : hasSelectedElement
+              ? undefined
+              : 100
+        }
+        min={canAccordionCollapse && appSourceCollapsed ? PANEL_HEADER_HEIGHT : 210}
```
eglitise: The amount of conditionals here make it difficult to read; can they be extracted as separate consts? This should also allow reusing them for the Selected Element Splitter panel, which has the same issue.

app/common/renderer/components/SessionInspector/SourceTab/SourceTab.jsx
```diff
@@ -1,25 +1,103 @@
...
+    // orientation, Splitter reads the wrong axis (width vs height) and
+    // caches a stale container size, leaving panels stuck at the wrong
+    // size until another resize happens to nudge it again.
+    const updateIsNarrow = () => setIsNarrow(window.innerWidth < NARROW_LAYOUT_BREAKPOINT);
+    window.addEventListener('resize', updateIsNarrow);
+    return () => window.removeEventListener('resize', updateIsNarrow);
+  }, []);
+
+  const [appSourceCollapsed, setAppSourceCollapsed] = useState(false);
+  const [selectedElementCollapsed, setSelectedElementCollapsed] = useState(false);
```
eglitise, line 28: For consistency with other files, prefer if `useState`s are moved to the top of the component (after the `props` destructuring), while `useEffect`s are moved to the bottom, before the `return` call.
I'm not really sure if this is commonly used sorting though, but best to remain consistent.

app/common/renderer/components/SessionInspector/Header/Header.module.css
tatanota: Done — replaced the horizontal scroll with `<Space wrap>`, so overflowing buttons wrap to a new row instead. `.headerButtons` is now `min-height` instead of a fixed `height`/`overflow-x`, so it can grow across rows, and the sibling main content area now sizes via flex (`min-height: 0`) instead of a hardcoded `calc(100% - 50px)` that assumed a single-row header.

app/common/renderer/components/SessionInspector/SourceTab/SourceTab.jsx
tatanota: Done — the breakpoint is now driven by a ResizeObserver on the Source tab's own container instead of `window.innerWidth`.

app/common/renderer/components/SessionInspector/SourceTab/AppSource/AppSourceCard.jsx
tatanota: Done — merged the collapse toggle into `AppSourceHeaderButtons`/`SelectedElementHeaderButtons` (removed the separate `CollapseToggleButton` and the extra wrapping `<span>` in each card's `extra`).

app/common/renderer/components/SessionInspector/SourceTab/SourceTab.jsx
tatanota: Done — extracted `splitterCollapsible` and a `getPanelSizing(collapsed, defaultSize, defaultMin)` helper, reused by both the App Source and Selected Element panels.

app/common/renderer/components/SessionInspector/SourceTab/SourceTab.jsx
tatanota: Done — moved the `useState`s to the top (after the props destructuring) and the `useEffect` to the bottom before the return.

app/common/renderer/components/SessionInspector/SourceTab/SourceTab.jsx
eglitise: Looks much better. I would even suggest reducing the breakpoint to 600px.

app/common/renderer/components/SessionInspector/Header/Header.module.css
eglitise: Better, but below the breakpoint where the second row is added, the rows are no longer centered. Not a huge problem, but would be nice to fix this if possible

tatanota: Fixed — added `justify-content: center` to the Space (via a new `.headerButtonsSpace` class), since `flex-wrap` left-aligns a row that doesn't fill the container's full width by default.

app/common/renderer/components/SessionInspector/SourceTab/SourceTab.jsx
tatanota: Done — lowered to 600px.

eglitise (approved):
Thanks!

PR #3145, opened 2026-09-06
feat: allow the Screenshot panel to be resized

### LINKED ISSUE #766: Ideas for UI/UX improvements

eglitise:
This PR adds support for resizing the Screenshot panel, using the same antd Splitter-based approach as #2196. It also adds better responsiveness for the screenshot controls buttons, allowing them to overflow onto a new row at small screenshot widths, using a similar approach to the one in #3064.

As part of this change, the maximum screenshot width has also been increased from 40 to 50% of the window width.
Another interesting feature here is that if the source screenshot is small enough to not be bound by the window height/width, then if the panel width is increased, the screenshot continues to scale above 100%, effectively acting like zoom.

Here's a few demos on different devices. All videos begin at the same panel size that is set at the start of a session:

<details>
<summary>Mobile (iOS) - panel size can be both decreased and increased</summary>

https://github.com/user-attachments/assets/1ac01ab9-5fe8-48d6-94aa-2b915913f132

</details>

<details>
<summary>TV (tvOS) - already starts at half the window width, so the panel can only be decreased</summary>

https://github.com/user-attachments/assets/7620c7f2-c592-4e07-9764-b629a5977ff9

</details>

<details>
<summary>Watch (watchOS) - starts at native resolution, so increasing the panel size results in zoom</summary>

https://github.com/user-attachments/assets/7d4546b6-3b43-4888-b1e5-0de65c78470d

</details>

Resolves #766


app/common/renderer/components/SessionInspector/Screenshot/Screenshot.jsx
```diff
@@ -28,42 +28,42 @@ const ScreenshotOuterSpinner = () => (
   </Spin>
 );
 
-// If the screenshot has too much space to the right or bottom, adjust the max width
-// of its container, so the source tree always fills the remaining space.
-// This keeps everything looking tight.
-const updateScreenshotScale = (screenshotContainerElRef, setScaleRatio, windowSize) => {
-  const screenshotContainer = screenshotContainerElRef.current;
-  if (!screenshotContainer) {
+// Calculate the target screenshot panel width (unless changed manually),
```
mykola-mokhnach, line 31: lets ask ai to make comments more concise

app/common/renderer/components/SessionInspector/Screenshot/Screenshot.jsx
```diff
@@ -28,42 +28,42 @@ const ScreenshotOuterSpinner = () => (
...
+// as well as the scale ratio of the original screenshot width to its target rendered width,
+// which is used to position screenshot overlay items like highlighters and gesture trails.
+// The image itself is rescaled purely using CSS. Scale calculation depends on the original image dimensions,
+// the width of the screenshot panel, and the width/height of the entire Inspector window.
+const updateScreenshotScale = (
+  screenshotContentElRef,
```
mykola-mokhnach: I think it makes more sense to make single arg object. Functions having more than 3 args are hard to read and maintain

app/common/renderer/components/SessionInspector/Screenshot/Screenshot.jsx
eglitise: It originally generated a similarly wordy comment, I rewrote it afterwards. I think all of this info is relevant enough to keep

eglitise: slimmed it down a bit

app/common/renderer/components/SessionInspector/Screenshot/Screenshot.jsx
eglitise: combined the size-related variables into one, reducing the argument count to 3

app/common/renderer/components/SessionInspector/Screenshot/Screenshot.jsx
```diff
@@ -28,42 +28,34 @@ const ScreenshotOuterSpinner = () => (
...
-    screenshotContainer.style.maxWidth = `${curMaxImgWidth}px`;
-  } else if (imgRect.width < containerRect.width) {
-    screenshotContainer.style.maxWidth = `${imgRect.width}px`;
-  }
-
-  // Calculate the ratio for scaling items overlaid on the screenshot
-  // (highlighter rectangles/circles, gestures, etc.)
-  const newImgWidth = screenshotImg.getBoundingClientRect().width;
-  setScaleRatio(windowSize.width / newImgWidth);
+  const imageWidthToFitHeight = (availableHeight * sizes.rawImage.width) / sizes.rawImage.height;
```
KazuCocoa: [P2] Use the screenshot's actual aspect ratio when calculating its displayed width

`sizes.rawImage` is populated from `windowSize` (`getWindowRect`), whose aspect ratio can differ from the screenshot's—for example, a web viewport paired with a full-device screenshot. CSS `object-fit: contain` uses the image's intrinsic aspect ratio, so this calculation can disagree with the actual displayed width and misalign both overlays and gesture coordinates. In a browser reproduction with `windowSize: {width: 390, height: 664}` and a 1170×2532 screenshot, the same position at 80% of the displayed image width maps to x=312 on the base revision but x=246 here. Could we derive the height-fitting width from the image's `naturalWidth`/`naturalHeight`, while retaining `windowSize.width` as the coordinate-space numerator for `scaleRatio`?

app/common/renderer/components/SessionInspector/Screenshot/Screenshot.module.css
```diff
@@ -1,26 +1,39 @@
...
 }
 
 .screenshotContainer img {
-  max-width: 100%;
-  max-height: 100%;
+  width: 100%;
+  height: 100%;
+  /* object-fit (rather than max-width/max-height) ensures the image is never stretched/distorted
+    in either direction, while still allowing it to be scaled up past its natural resolution */
+  object-fit: contain;
```
KazuCocoa, line 14: [P2] Exclude the blank area beside the image from screenshot interactions

When a portrait panel is widened past the point where the image is height-bound, `object-fit: contain` leaves a blank strip on the right, but the image element and `screenshotBox` still cover the full panel. The existing mouse handlers accept that strip and convert its offsets into device coordinates without checking the visible image bounds. In a browser reproduction with a 390×844 device coordinate space, clicking the strip dispatched a tap at `{x: 494, y: 130}`, outside the device screen. Could we restrict the interactive area to the displayed image, or reject pointer events outside its bounds before creating tap/swipe or gesture coordinates?

KazuCocoa (commented):
Two coordinate-handling issues reproduced in browser checks using the screenshot/splitter components with mocked session data.

app/common/renderer/components/SessionInspector/SessionInspector.jsx
```diff
@@ -43,6 +46,29 @@ const Inspector = (props) => {
...
+      // ignore sub-pixel differences to avoid a resizing loop
+      return Math.abs(suggestedWidth - curWidth) < 1 ? curWidth : suggestedWidth;
+    });
+  };
+
+  // Triggered when manually adjusting the splitter. Only needed to trip the manual resize flag.
+  const setPanelWidthManually = (widths) => {
+    screenshotPanelResizedManually.current = true;
+    setScreenshotPanelWidth(widths[0]);
+  };
```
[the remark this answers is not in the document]
eglitise: Not true, `onResize` has the signature `(sizes: number[]) => void`: https://ant.design/components/splitter#splitter

app/common/renderer/components/SessionInspector/Screenshot/Screenshot.module.css
eglitise: This isn't a new issue, but I've fixed it in https://github.com/appium/appium-inspector/pull/3145/commits/cf6c0f0507c98e8978716d84f5a8794be8ffad27

app/common/renderer/components/SessionInspector/SessionInspector.module.css
```diff
@@ -8,46 +8,18 @@
...
-  height: calc(100% - 54px);
-  padding: 12px;
-  scrollbar-width: none; /* -webkit-scrollbar for Firefox */
-}
-
-.inspectorMain :global(.ant-card-body::-webkit-scrollbar) {
-  background: transparent;
-  width: 14px;
+  padding: 1em;
 }
```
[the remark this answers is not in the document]
eglitise: fixed

app/common/renderer/components/SessionInspector/Screenshot/Screenshot.jsx
```diff
@@ -108,15 +102,22 @@ const Screenshot = (props) => {
...
 
-  /**
-   * Ensures component dimensions are adjusted only once windowSize exists.
-   */
+  // set up triggers for recalculating the screenshot image scaling factor, upon changes in size
+  // for the raw image, MJPEG stream, screenshot panel, and the Inspector window
   useEffect(() => {
     if (!windowSize || !JSON.stringify(windowSize)) {
       return;
     }
```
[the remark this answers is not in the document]
eglitise: fixed

app/common/renderer/components/SessionInspector/Screenshot/Screenshot.jsx
eglitise: Yeah, these changes seem to have broken webview element coordinate calculations as a whole. Looking into it.

KazuCocoa (approved):

app/common/renderer/components/SessionInspector/Screenshot/Screenshot.jsx
eglitise: Fixed in https://github.com/appium/appium-inspector/pull/3145/changes/efd1ffbcc597ec5ee054e41876ed1e81ca702909 . Tested on iOS and Android emulators on the WDIO homepage in webview - moving the splitter keeps the highlighters in their correct positions, and coordinate-based tap actions land in the correct locations.

PR #3150, opened 2026-09-07
feat: customize important source attributes

### LINKED ISSUE #2976: feature request: allow customizing 'important' attributes in source tree
### Current Behavior

Appium Inspector v2025.7.1 do not show **`hint`** attribute with value in source tree on Android if this attribute present for this node

### Suggested Solution

Appium Inspector v2025.7.1 show **`hint`** attribute with value in source tree on Android if this attribute present for this node (e.g. text, resource-id and content-desc attributes)

### Additional Information

_No response_

lucasmariano003-wq:
Fixes #2976.

Adds a **Customize Visible Attributes** dialog beside the source-tree attribute toggle. Users can choose default or previously selected attributes, enter custom names, clear the selection, and restore the shared defaults. Source search follows the same visible attributes as the tree. Suggestions do not require scanning the entire source tree.

Defaults are defined in sorted order in `setting-defs.js`. Selections are sorted on update and saved through the shared web/Electron settings layer. A small initialization cache is refreshed after edits so new Inspector sessions in the same renderer retain the latest selection, including an empty selection.

Validation:
- `npm test`: lint, 200 unit tests and 12 integration tests passed. The preference tests are in `test/integration/` and cover saved/default initialization, session remounts, empty selections and default restoration.
- `npm run format:check` passed.
- `npm run build`: browser, Electron and plugin builds passed; Electron was not launched.
- Production browser build with FakeDriver: custom `width` appeared in the source and produced four search matches; it survived a new session. Clearing the selection survived another session, and restoring defaults worked. The dialog layout was checked visually.
- The final published tree was fetched back and matches the tested revision.

This contribution was implemented and validated with OpenAI Codex on behalf of this GitHub account.


lucasmariano003-wq:
The CLA is now signed and all nine checks have passed, including the macOS, Ubuntu and Windows build/test jobs. This is ready for review; the local production-browser interaction checks are summarized above.

app/common/public/locales/en/translation.json
```diff
@@ -340,5 +340,8 @@
   "toggleMultiDisplayMode": "Toggle Multi-Display Mode",
   "switchToEspressoSubdriver": "Switch to Espresso Subdriver",
   "switchToComposeSubdriver": "Switch to Compose Subdriver",
-  "w3cActionsUnsupported": "W3C Actions are not supported on this platform"
+  "w3cActionsUnsupported": "W3C Actions are not supported on this platform",
+  "Important Attributes": "Important Attributes",
```
eglitise: Since this is also used as the tooltip label, which usually starts with a verb, I'd prefer something like 'Customize Visible Attributes'

app/common/renderer/components/SessionInspector/SourceTab/AppSource/AppSourceTreeWrapper.jsx
```diff
@@ -30,19 +34,46 @@ const AppSourceTreeWrapper = ({
 
   const [searchValue, setSearchValue] = useState('');
   const [autoExpandParent, setAutoExpandParent] = useState(true);
+  const [importantAttrs, setImportantAttrs] = useState(IMPORTANT_SOURCE_ATTRS);
+  const [settingsLoaded, setSettingsLoaded] = useState(false);
```
eglitise: I'm not sure this is needed; other settings like theme don't have this safeguard and still work fine.
I think it's better to move the `IMPORTANT_SOURCE_ATTRS` definition from `constants/source.js` to the `setting-defs.js` file (while still allowing it to be exported), and set it as the default value of `IMPORTANT_SOURCE_ATTRIBUTES`. That way it will be initialized to `IMPORTANT_SOURCE_ATTRS`, while still being editable and resettable.

app/common/renderer/components/SessionInspector/SourceTab/AppSource/AppSourceTreeWrapper.jsx
```diff
@@ -30,19 +34,46 @@ const AppSourceTreeWrapper = ({
 
   const [searchValue, setSearchValue] = useState('');
   const [autoExpandParent, setAutoExpandParent] = useState(true);
+  const [importantAttrs, setImportantAttrs] = useState(IMPORTANT_SOURCE_ATTRS);
```
eglitise: You can probably do a top-level `const retrievedAttrs = await getSetting(IMPORTANT_SOURCE_ATTRIBUTES)` and then do `useState(retrievedAttrs)` here. That way you'll be able to remove the useEffect hook entirely

app/common/renderer/components/SessionInspector/SourceTab/AppSource/AppSourceTreeWrapper.jsx
```diff
@@ -30,19 +34,46 @@ const AppSourceTreeWrapper = ({
...
+    await setSetting(IMPORTANT_SOURCE_ATTRIBUTES, attrs);
+  };
 
   const flatten = (elemObj) => [elemObj, ...(elemObj.children?.flatMap(flatten) || [])];
   const flatTreeData = sourceJSON && flatten(sourceJSON);
+  const availableAttrs = [
+    ...new Set([
+      ...IMPORTANT_SOURCE_ATTRS,
+      ...importantAttrs,
+      ...(flatTreeData || []).flatMap((element) => Object.keys(element.attributes)),
```
eglitise: I'm not sure this is worth the performance of another flatMap just to get all the attributes, if the user can just enter them manually

app/common/renderer/components/SessionInspector/SourceTab/AppSource/SourceAttributeSettings.jsx
```diff
@@ -0,0 +1,52 @@
...
+        onCancel={() => setModalOpen(false)}
+        centered
+      >
+        <Space orientation="vertical" style={{width: '100%'}}>
+          <span>{t('importantAttributesDescription')}</span>
+          <Select
+            aria-label={label}
+            mode="tags"
+            value={importantAttrs}
+            options={availableAttrs.map((value) => ({value}))}
```
eglitise, line 38: `availableAttrs` is sorted but `importantAttrs` is not

app/common/renderer/components/SessionInspector/SourceTab/AppSource/SourceAttributeSettings.jsx
```diff
@@ -0,0 +1,52 @@
...
+        styles={{title: {fontSize: '18px'}}}
+        open={modalOpen}
+        footer={null}
+        onCancel={() => setModalOpen(false)}
+        centered
+      >
+        <Space orientation="vertical" style={{width: '100%'}}>
+          <span>{t('importantAttributesDescription')}</span>
+          <Select
+            aria-label={label}
```
eglitise: The app currently doesn't have `aria-label`s for Select elements, plus this label matches the one used for the Button, which is not great. I would suggest simply removing this one.

app/common/renderer/components/SessionInspector/SourceTab/AppSource/AppSourceTreeWrapper.jsx
```diff
@@ -151,6 +180,10 @@ const AppSourceTreeWrapper = ({
               onSearchChange={onSearchChange}
               searchValue={searchValue}
               matchingElementsCount={matchingElements.length}
+              importantAttrs={importantAttrs}
+              availableAttrs={availableAttrs}
+              onImportantAttrsChange={updateImportantAttrs}
```
eglitise: Recommend keeping the `updateImportantAttrs` name throughout all affected components

docs/session-inspector/source.md
```diff
@@ -42,6 +42,13 @@ the tree only shows the class name for each element entry, along with several 'i
 such as `value` and `content-desc`. All non-'important' attributes are hidden by default for ease of
 navigation, but can be revealed with the [Toggle Attributes button](#source-action-buttons).
 
+Use the **Important Attributes** settings button beside **Toggle Attributes** to customize the compact
```
eglitise: You can omit this, I'll need to update the docs anyway to adjust the table and add a screenshot

eglitise (commented):
Thanks! The GUI looks nice and works as expected, but I have a few comments.

lucasmariano003-wq:
Thanks for the review. Addressed the requested changes through ee36896: the action is now Customize Visible Attributes; the default list is exported from setting-defs.js and used as the setting default; the loading safeguard and loading effect are removed. Preferences are read before rendering, and the cached selection is updated after edits so a newly mounted Inspector session in the same renderer retains the latest saved choice.

The picker offers only defaults and previously selected values, without scanning every element for attribute names. Selected values are sorted without mutating the shared defaults. The callback is consistently named updateImportantAttrs, the duplicate Select aria-label is removed, and the added documentation paragraph is removed so you can update the table and screenshot together.

Validation: 205 unit tests, 7 integration tests, lint, formatting, and browser/Electron/plugin builds passed. Five new settings regressions cover defaults, saved ordering, empty selections across sessions/reloads, and restoring defaults without mutation. The published tree was fetched back and matches the tested files.

app/common/renderer/components/SessionInspector/SourceTab/AppSource/AppSourceTreeWrapper.jsx
```diff
@@ -30,19 +34,21 @@ const AppSourceTreeWrapper = ({
 
   const [searchValue, setSearchValue] = useState('');
   const [autoExpandParent, setAutoExpandParent] = useState(true);
+  const [importantAttrs, setImportantAttrs] = useState(retrievedAttrs);
+
+  const updateImportantAttrs = async (attrs) => {
+    // Keep the next inspector session in sync with the saved preferences.
+    retrievedAttrs = [...attrs].sort();
```
eglitise: There's no reason for this to reuse the top-level `retrievedAttrs` variable, so I'd suggest just making this into its own const with a different name to avoid confusion

app/common/renderer/components/SessionInspector/SourceTab/AppSource/AppSourceTreeWrapper.jsx
```diff
@@ -2,13 +2,17 @@ import {Spin, Tree} from 'antd';
...
+import {IMPORTANT_SOURCE_ATTRIBUTES} from '../../../../../shared/setting-defs.js';
+import {getSetting, setSetting} from '../../../../polyfills.js';
+import {getVisibleSourceAttributes, sourceElementMatchesSearch} from '../../../../utils/source-attributes.js';
 import AppSourceTree from './AppSourceTree.jsx';
 import AppSourceTreeActions from './AppSourceTreeActions.jsx';
 
 import inspectorStyles from '../../SessionInspector.module.css';
 import styles from './AppSource.module.css';
 
+let retrievedAttrs = [...(await getSetting(IMPORTANT_SOURCE_ATTRIBUTES))].sort();
```
eglitise: You can simplify this by sorting the order inside `IMPORTANT_SOURCE_ATTRIBUTES` itself, so this can turn into `const retrievedAttrs = await getSetting(IMPORTANT_SOURCE_ATTRIBUTES);`

app/common/renderer/components/SessionInspector/SourceTab/AppSource/SourceAttributeSettings.jsx
```diff
@@ -0,0 +1,53 @@
...
+      </Tooltip>
+      <Modal
+        title={label}
+        styles={{title: {fontSize: '18px'}}}
+        open={modalOpen}
+        footer={null}
+        onCancel={() => setModalOpen(false)}
+        centered
+      >
+        <Space orientation="vertical" style={{width: '100%'}}>
```
eglitise: The width: 100% can be removed, I don't see a difference with or without it. Same for the Select component

app/common/renderer/components/SessionInspector/SourceTab/AppSource/SourceAttributeSettings.jsx
```diff
@@ -0,0 +1,53 @@
...
+          <Select
+            mode="tags"
+            value={importantAttrs}
+            options={availableAttrs.map((value) => ({value}))}
+            onChange={updateImportantAttrs}
+            tokenSeparators={[',']}
+            style={{width: '100%'}}
+            allowClear
+          />
+          <Button onClick={() => updateImportantAttrs([...IMPORTANT_SOURCE_ATTRS])}>
```
eglitise: `updateImportantAttrs` already spreads the parameter, no need to also do it here, so just `updateImportantAttrs(IMPORTANT_SOURCE_ATTRS)` is fine

app/common/renderer/components/SessionInspector/SourceTab/AppSource/AppSourceTreeWrapper.jsx
eglitise: This would also allow you to omit some `sort()` calls in `source-attribute-settings.spec.jsx`

test/unit/utils-source-attributes.spec.js
```diff
@@ -0,0 +1,49 @@
...
+  sourceElementMatchesSearch,
+} from '../../app/common/renderer/utils/source-attributes.js';
+import {IMPORTANT_SOURCE_ATTRS} from '../../app/common/shared/setting-defs.js';
+
+const element = {
+  tagName: 'android.widget.EditText',
+  attributes: {text: 'Email', hint: 'Enter your email', focused: 'false', 'content-desc': ''},
+};
+
+describe('Source attribute preferences', () => {
```
eglitise: For consistency with other unit tests I would suggest:
* Rename this describe block to `utils/source-attributes.js`
* Add two inner describe blocks, one for each method (`#getVisibleSourceAttributes` and `#sourceElementMatchesSearch`)

test/unit/source-attribute-settings.spec.jsx
```diff
@@ -0,0 +1,116 @@
+import {createElement} from 'react';
```
eglitise: This file seems to contain integration tests, so it's best to move it to the `tests/integration` directory

test/unit/source-attribute-settings.spec.jsx
```diff
@@ -0,0 +1,116 @@
...
+  });
+
+  it('initializes the first session from the shared default without changing its order', async () => {
+    const {DEFAULT_SETTINGS, IMPORTANT_SOURCE_ATTRS} = await import('../../app/common/shared/setting-defs.js');
+    const originalDefaults = [...IMPORTANT_SOURCE_ATTRS];
+    renderSession(await loadWrapper());
+
+    expect(rendered.actions.importantAttrs).toEqual([...originalDefaults].sort());
+    expect(DEFAULT_SETTINGS[settingKey]).toBe(IMPORTANT_SOURCE_ATTRS);
+    expect(IMPORTANT_SOURCE_ATTRS).toEqual(originalDefaults);
```
eglitise: `IMPORTANT_SOURCE_ATTRS` and `originalDefaults` are never changed, so this assertion will always be true and can be removed. Same for the last test

test/unit/source-attribute-settings.spec.jsx
```diff
@@ -0,0 +1,116 @@
...
+    expect(renderedNode()).not.toContain('Enter email');
+  });
+
+  it('loads and sorts saved attributes before rendering the first session', async () => {
+    savedSettings.set(settingKey, ['text', 'hint']);
+    renderSession(await loadWrapper());
+
+    expect(rendered.actions.importantAttrs).toEqual(['hint', 'text']);
+    expect(renderedNode()).toContain('Enter email');
+    expect(savedSettings.get(settingKey)).toEqual(['text', 'hint']);
```
eglitise: This assertion can probably also be removed, since it only checks that `get`/`set` work as expected

eglitise (commented):
Looking better 👍

lucasmariano003-wq:
Thanks for the follow-up. Updated through `f8b8153`. The defaults are now ordered in `setting-defs.js`; the callback uses its own `const sortedAttrs`; the duplicate widths and reset-button copy are removed. The utility tests now follow the existing nested describe convention, and the preference tests live in `test/integration/` with the redundant assertions and default-list sorts removed.

I kept one exception: `retrievedAttrs` remains a mutable initialization cache. `MemoryRouter` navigates from the Inspector to the session builder and back without reloading this module. Making it a constant and removing the cache update makes the next session use the old selection. I reproduced this with three failing regressions (edited attributes, empty selection, and restored defaults), then retained the cache update with an explanatory comment. The local constant handles sorting, state and persistence independently.

Validation: 200 unit tests and 12 integration tests passed, as did lint, formatting and the browser/Electron/plugin builds. In the production browser build with FakeDriver, adding `width` displayed it and produced four search matches; a second session retained it, a third preserved an empty selection, and restoring defaults worked. Electron was built but not launched. The final published tree was fetched back and matches the tested files.

eglitise (approved):

KazuCocoa (approved):

PR #3152, opened 2026-09-08
feat: reorder actions in the gesture builder

### LINKED ISSUE #1968: feature request: Allow Rearranging Actions in Gesture Builder
### Current Behavior

The Gesture Builder in Appium Inspector currently allows users to create multi-pointer gestures but cannot reorder actions. This makes it difficult to modify gestures dynamically without deleting and recreating steps.

### Suggested Solution

- Users should be able to drag and drop or use arrow buttons to rearrange actions within a pointer sequence.
- The updated order should be reflected immediately in the gesture timeline.

### Benefits
Increases usability by allowing users to tweak gestures without deleting steps.
Saves time when designing complex multi-pointer interactions.
Aligns with other UI-based gesture editors that allow reordering.

Would love to see this enhancement in a future release! 🚀

lucasmariano003-wq:
Gesture actions can be reordered by dragging an SVG grip onto another card within the same pointer. The drop target is highlighted. Keyboard pickup/drop with Space or Enter, arrow navigation, and Escape cancellation are supported. The grip is hidden for a single action, while the action fields and delete/coordinate-picker controls remain usable.

Moves retain coordinates, duration and other parameters, refresh the timeline, renumber position-based IDs and clear the coordinate picker. Each pointer has its own drag context. Optimistic sorting and drop animation are disabled to keep DOM ordering consistent with positional IDs; focus is restored to the destination after keyboard moves. The saved gesture format is unchanged.

Closes #1968.

Validation:
- Clean `npm ci` passed; `@dnd-kit/dom` 0.5.0 resolves through `@dnd-kit/react`. The lockfile change only removes the direct declaration.
- 207 unit tests, 7 integration tests, lint and formatting passed.
- Browser, Electron and plugin builds passed.
- The production Chromium browser build with FakeDriver verified mouse/keyboard sorting, destination focus, Escape, pointer switching, one/two-action transitions, deletion during drag, editable single-action fields, and saving/reopening with parameters preserved.
- All nine published files were fetched back and match the tested revision.

Native touch, screen-reader speech and the native Electron UI were not exercised; Electron compiled successfully. OpenAI Codex performed the implementation and local validation for this GitHub account.

### DIFF
--- a/app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditor.module.css
+++ b/app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditor.module.css
@@ -75,6 +75,19 @@
   flex-shrink: 0;
 }
 
+.tickCard[data-drop-target] {
+  outline: 2px solid var(--ant-color-primary);
+}
+
+.tickDragHandle:not(:disabled) {
+  cursor: grab;
+  touch-action: none;
+}
+
+.tickDragHandle:not(:disabled):active {
+  cursor: grabbing;
+}
+
 .tickPlusBtnWrapper {
   height: 100%;
   display: flex;
--- a/app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditor.module.css
+++ b/app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditor.module.css
@@ -54,6 +54,11 @@
   text-align: center;
 }
 
+.tickCard {
+  display: flex;
+  flex-direction: column;
+}
+
 .tickCard :global(.ant-card-head) {
   min-height: unset;
   padding: 0px;
@@ -62,7 +67,12 @@
 
 .tickCard :global(.ant-card-body) {
   padding-top: 6px;
-  height: calc(100% - 25px);
+  height: auto;
+  flex: 1;
+}
+
+.tickCard :global(.ant-card-actions) {
+  flex-shrink: 0;
 }
 
 .tickPlusBtnWrapper {
--- a/app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorPointerTabContents.jsx
+++ b/app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorPointerTabContents.jsx
@@ -1,7 +1,12 @@
+import {Feedback} from '@dnd-kit/dom';
+import {SortableKeyboardPlugin} from '@dnd-kit/dom/sortable';
+import {DragDropProvider} from '@dnd-kit/react';
+import {isSortable, useSortable} from '@dnd-kit/react/sortable';
 import {IconPlus} from '@tabler/icons-react';
 import {Button, Col, Row, Tooltip} from 'antd';
 import {useTranslation} from 'react-i18next';
 
+import {moveGestureTick} from '../../../../utils/gesture-editing.js';
 import GestureEditorTickCard from './GestureEditorTickCard.jsx';
 import GestureEditorTickCardContents from './GestureEditorTickCardContents.jsx';
 
@@ -40,30 +45,49 @@ const AddNewTickButton = ({id, pointers, setPointers}) => {
  */
 const GestureEditorTick = ({
   tick,
+  index,
+  pointer,
   pointers,
   setPointers,
   selectedTick,
   selectTick,
   unselectTick,
   getDefaultMoveDuration,
-}) => (
-  <GestureEditorTickCard
-    tick={tick}
-    pointers={pointers}
-    setPointers={setPointers}
-    selectedTick={selectedTick}
-    selectTick={selectTick}
-    unselectTick={unselectTick}
-  >
-    <GestureEditorTickCardContents
-      tick={tick}
-      selectTick={selectTick}
-      getDefaultMoveDuration={getDefaultMoveDuration}
-      pointers={pointers}
-      setPointers={setPointers}
-    />
-  </GestureEditorTickCard>
-);
+}) => {
+  const {ref, handleRef, isDropTarget} = useSortable({
+    id: tick.id,
+    index,
+    group: pointer.id,
+    disabled: pointer.ticks.length < 2,
+    // Tick IDs describe positions, not stable identities. Commit the new order on drop
+    // instead of optimistically reordering DOM nodes behind React's positional keys.
+    plugins: [SortableKeyboardPlugin, Feedback.configure({dropAnimation: null})],
+  });
+
+  return (
+    <Col xs={12} sm={12} md={12} lg={8} xl={6} xxl={4} xxxl={3} ref={ref}>
+      <GestureEditorTickCard
+        tick={tick}
+        dragHandleRef={handleRef}
+        dragDisabled={pointer.ticks.length < 2}
+        isDropTarget={isDropTarget}
+        pointers={pointers}
+        setPointers={setPointers}
+        selectedTick={selectedTick}
+        selectTick={selectTick}
+        unselectTick={unselectTick}
+      >
+        <GestureEditorTickCardContents
+          tick={tick}
+          selectTick={selectTick}
+          getDefaultMoveDuration={getDefaultMoveDuration}
+          pointers={pointers}
+          setPointers={setPointers}
+        />
+      </GestureEditorTickCard>
+    </Col>
+  );
+};
 
 /**
  * Contents of a pointer tab in the gesture editor.
@@ -76,25 +100,51 @@ const GestureEditorPointerTabContents = ({
   selectTick,
   unselectTick,
   getDefaultMoveDuration,
-}) => (
-  <Row gutter={[24, 24]}>
-    {pointer.ticks.map((tick) => (
-      <Col xs={12} sm={12} md={12} lg={8} xl={6} xxl={4} xxxl={3} key={tick.id}>
-        <GestureEditorTick
-          tick={tick}
-          pointers={pointers}
-          setPointers={setPointers}
-          selectedTick={selectedTick}
-          selectTick={selectTick}
-          unselectTick={unselectTick}
-          getDefaultMoveDuration={getDefaultMoveDuration}
-        />
-      </Col>
-    ))}
-    <Col xs={12} sm={12} md={12} lg={8} xl={6} xxl={4} xxxl={3}>
-      <AddNewTickButton id={pointer.id} pointers={pointers} setPointers={setPointers} />
-    </Col>
-  </Row>
-);
+}) => {
+  const handleDragEnd = (event) => {
+    const {source, target, activatorEvent} = event.operation;
+    if (!isSortable(source)) {
+      return;
+    }
+    const updatedPointers =
+      !event.canceled && isSortable(target)
+        ? moveGestureTick(pointers, pointer.id, source.id, target.index - source.index)
+        : pointers;
+    if (updatedPointers !== pointers) {
+      unselectTick();
+      setPointers(updatedPointers);
+    }
+    if (activatorEvent instanceof KeyboardEvent) {
+      // The default drop animation restores the old positional ID. Keep keyboard
+      // focus on the moved action instead, or on the source if the drag was canceled.
+      const handle = updatedPointers !== pointers ? target.sortable.draggable.handle : source.handle;
+      requestAnimationFrame(() => handle?.focus());
+    }
+  };
+
+  return (
+    <DragDropProvider onDragEnd={handleDragEnd}>
+      <Row gutter={[24, 24]}>
+        {pointer.ticks.map((tick, index) => (
+          <GestureEditorTick
+            key={tick.id}
+            tick={tick}
+            index={index}
+            pointer={pointer}
+            pointers={pointers}
+            setPointers={setPointers}
+            selectedTick={selectedTick}
+            selectTick={selectTick}
+            unselectTick={unselectTick}
+            getDefaultMoveDuration={getDefaultMoveDuration}
+          />
+        ))}
+        <Col xs={12} sm={12} md={12} lg={8} xl={6} xxl={4} xxxl={3}>
+          <AddNewTickButton id={pointer.id} pointers={pointers} setPointers={setPointers} />
+        </Col>
+      </Row>
+    </DragDropProvider>
+  );
+};
 
 export default GestureEditorPointerTabContents;
--- a/app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorTickCard.jsx
+++ b/app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorTickCard.jsx
@@ -1,4 +1,4 @@
-import {IconArrowLeft, IconArrowRight, IconFocus2, IconX} from '@tabler/icons-react';
+import {IconArrowLeft, IconArrowRight, IconFocus2, IconGripVertical, IconX} from '@tabler/icons-react';
 import {Button, Card, Tooltip} from 'antd';
 import {useTranslation} from 'react-i18next';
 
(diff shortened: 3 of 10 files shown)


eglitise:
Thanks. It's a good approach (haven't tested this yet), but I wonder if it's possible to implement this using drag-and-drop for a better UX? Probably with something like https://dndkit.com/react/quickstart/

lucasmariano003-wq:
Thanks for the suggestion, @eglitise. Implemented drag-and-drop with `@dnd-kit/react`, through 6e256bf. Each card has a dedicated grip and a highlighted drop target. Keyboard pickup/drop, arrow navigation and Escape cancellation work too, with focus following the moved action; the earlier/later buttons remain available.

The drag context is scoped to each pointer and is destroyed when switching tabs. Reordering preserves the action parameters and saved gesture format, renumbers positional IDs, updates the timeline and clears the coordinate picker. Optimistic DOM sorting is disabled to keep those positional IDs consistent.

Validation: 207 unit tests, 7 integration tests, lint, formatting and browser/Electron/plugin builds passed. The production browser UI with FakeDriver passed repeated non-adjacent moves, keyboard focus/cancellation, dragging between grid rows, switching pointers during drag, parameter preservation and saving/reopening at 900, 1280 and 1680 px. Published files were fetched back and match the tested files byte for byte.

eglitise:
Thanks. With the drag-and-drop added, I think it's fine to remove the earlier/later buttons to simplify the implementation and make the cards less busy.

lucasmariano003-wq:
Thanks, @eglitise. Removed the earlier/later buttons, their component and unused translations through `f4a7bab`. The gesture documentation is updated as well. The drag grip and its existing keyboard support remain available, alongside the coordinate-picker and delete controls.

Validation: 207 unit tests, 7 integration tests, lint, formatting and browser/Electron/plugin builds passed. In the production browser with FakeDriver, keyboard sorting and non-adjacent pointer dragging preserved all durations and coordinates; the saved gesture reopened with the same order. The published tree was fetched back and matches the tested files.

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorTickCard.jsx
```diff
@@ -59,23 +59,53 @@ const GestureEditorTickCardHeaderButtons = ({tick, pointers, setPointers, select
...
+            ref={dragHandleRef}
+            aria-label={dragLabel}
+            className={styles.tickDragHandle}
+            size="small"
+            type="text"
+            disabled={dragDisabled}
+            icon={<IconGripVertical size={18} />}
+          />
+        </Tooltip>
+      }
```
eglitise: I don't think this needs a tooltip, and it doesn't even need to be a button (so the translation key can also be removed). Try this instead, which will simply hide the icon if there's only one tick. It also allows you to remove the `not(:disabled)` directive for the `tickDragHandle` class CSS. For a nicer look, I would further recommend adding `margin-top: 3px` for that class.
```suggestion
      title={!dragDisabled && <IconGripVertical size={18} className={styles.tickDragHandle} ref={dragHandleRef} />}
```

docs/session-inspector/gestures.md
```diff
@@ -95,6 +95,14 @@ The title of each pointer can also be clicked and edited.
 Pointer actions are used to define the behavior of each pointer. Each individual action is presented
 as a rectangular card, and new cards can be freely added and removed.
 
+Drag a card by the grip in its upper-left corner onto another card to move the action to that
```
eglitise: This whole paragraph can be removed and merged into the preceding sentence:

`[..] and new cards can be freely added, removed, or reordered using drag-and-drop.`

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorTickCard.jsx
```diff
@@ -59,23 +59,53 @@ const GestureEditorTickCardHeaderButtons = ({tick, pointers, setPointers, select
...
+  unselectTick,
+}) => {
+  const {t} = useTranslation();
+  const dragLabel = t('dragGestureAction', {id: tick.id});
+
+  return (
+    <Card
+      hoverable={true}
+      className={styles.tickCard}
+      data-drop-target={isDropTarget || undefined}
```
eglitise: This approach of adding a custom property purely as a CSS class link is confusing and isn't used elsewhere in the codebase. For consistency I would instead suggest combining multiple classes:
```suggestion
      className={isDropTarget ? `${styles.tickCard} ${styles.tickCardDropTarget}` : styles.tickCard}
```
(rename the `tickCard[data-drop-target]` to `tickCardDropTarget` in the CSS file)

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorPointerTabContents.jsx
```diff
@@ -40,30 +45,49 @@ const AddNewTickButton = ({id, pointers, setPointers}) => {
...
-      setPointers={setPointers}
-    />
-  </GestureEditorTickCard>
-);
+}) => {
+  const {ref, handleRef, isDropTarget} = useSortable({
+    id: tick.id,
+    index,
+    group: pointer.id,
+    disabled: pointer.ticks.length < 2,
```
eglitise: Recommend defining `const dragDisabled = pointer.ticks.length < 2;` so it's not duplicated

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorPointerTabContents.jsx
```diff
@@ -40,30 +45,49 @@ const AddNewTickButton = ({id, pointers, setPointers}) => {
...
-      pointers={pointers}
-      setPointers={setPointers}
-    />
-  </GestureEditorTickCard>
-);
+}) => {
+  const {ref, handleRef, isDropTarget} = useSortable({
+    id: tick.id,
+    index,
+    group: pointer.id,
```
eglitise: Is the group needed if only one pointer is editable at any time?

package.json
```diff
@@ -61,6 +61,8 @@
   },
   "dependencies": {
     "@ant-design/icons": "6.3.4",
+    "@dnd-kit/dom": "0.5.0",
```
eglitise: `@dnd-kit/dom` doesn't need to be explicitly added since it's a dependency of `@dnd-kit/react`

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorPointerTabContents.jsx
```diff
@@ -40,30 +45,49 @@ const AddNewTickButton = ({id, pointers, setPointers}) => {
...
+    disabled: pointer.ticks.length < 2,
+    // Tick IDs describe positions, not stable identities. Commit the new order on drop
+    // instead of optimistically reordering DOM nodes behind React's positional keys.
```
eglitise: I was originally unsure why the `plugins` parameter is even needed, so I suggest extending this comment to clarify that it's done to explicitly remove the default `OptimisticSortingPlugin`, and disable the drop animation, since it looks confusing with the position-based tick IDs

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorTickCard.jsx
```diff
@@ -59,23 +59,53 @@ const GestureEditorTickCardHeaderButtons = ({tick, pointers, setPointers, select
...
+  selectedTick,
+  selectTick,
+  unselectTick,
+}) => {
+  const {t} = useTranslation();
+  const dragLabel = t('dragGestureAction', {id: tick.id});
+
+  return (
+    <Card
+      hoverable={true}
```
eglitise: This already existed previously, but it's probably best to remove this hoverable line, since it doesn't do much highlighting and its absence makes the drag cursor more noticeable

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorPointerTabContents.jsx
```diff
@@ -76,25 +100,51 @@ const GestureEditorPointerTabContents = ({
...
+    if (updatedPointers !== pointers) {
+      unselectTick();
+      setPointers(updatedPointers);
+    }
+    if (activatorEvent instanceof KeyboardEvent) {
+      // The default drop animation restores the old positional ID. Keep keyboard
+      // focus on the moved action instead, or on the source if the drag was canceled.
+      const handle = updatedPointers !== pointers ? target.sortable.draggable.handle : source.handle;
+      requestAnimationFrame(() => handle?.focus());
+    }
```
eglitise: Is this needed if the default drop animation is disabled? I don't see any difference with or without this block

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorPointerTabContents.jsx
```diff
@@ -76,25 +100,51 @@ const GestureEditorPointerTabContents = ({
...
-    ))}
-    <Col xs={12} sm={12} md={12} lg={8} xl={6} xxl={4} xxxl={3}>
-      <AddNewTickButton id={pointer.id} pointers={pointers} setPointers={setPointers} />
-    </Col>
-  </Row>
-);
+}) => {
+  const handleDragEnd = (event) => {
+    const {source, target, activatorEvent} = event.operation;
+    if (!isSortable(source)) {
```
eglitise: Are either of the `isSortable` guards needed?

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorPointerTabContents.jsx
```diff
@@ -76,25 +100,51 @@ const GestureEditorPointerTabContents = ({
...
+      return;
+    }
+    const updatedPointers =
+      !event.canceled && isSortable(target)
+        ? moveGestureTick(pointers, pointer.id, source.id, target.index - source.index)
+        : pointers;
+    if (updatedPointers !== pointers) {
+      unselectTick();
+      setPointers(updatedPointers);
+    }
```
eglitise: If the `isSortable` and `activatorEvent` blocks end up being removed as per my other comments, you can simplify this. Note that `target` can be null.
```suggestion
    if (!event.canceled && Number.isInteger(target?.index) && target.index - source.index !== 0) {
      const updatedPointers = moveGestureTick(pointers, pointer.id, source.id, target.index - source.index);
      unselectTick();
      setPointers(updatedPointers);
    }
```

app/common/renderer/utils/gesture-editing.js
```diff
@@ -0,0 +1,31 @@
+/** Move one action within its pointer and keep the position-based tick IDs consistent. */
+export function moveGestureTick(pointers, pointerId, tickId, offset) {
```
eglitise: Many variables in this method have fairly generic names and could be clarified, e.g. `offset` -> `tickIndexOffset`, `index` -> `tickIndex`, `targetIndex` -> `targetTickIndex`, `ticks` -> `newTicks`

test/unit/utils-gesture-editing.spec.js
```diff
@@ -0,0 +1,77 @@
...
+      {id: '1.2', type: 'pointerDown', button: 0},
+      {id: '1.3', type: 'pause', duration: 750},
+    ],
+  },
+  {id: '2', name: 'Second pointer', ticks: [{id: '2.1', type: 'pointerUp', button: 0}]},
+];
+
+describe('moveGestureTick', () => {
+  it('moves an action earlier, retaining its parameters and renumbering every position', () => {
+    const original = structuredClone(pointers);
```
eglitise: Since `moveGestureTick` doesn't modify in-place, I'm not sure of the purpose for this `structuredClone` and later `expect(pointers).toEqual(original)` call, both here and in the third test

eglitise (commented):
Thanks! It works as expected, but can still be improved.

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditor.module.css
```diff
@@ -62,7 +67,25 @@
...
+.tickCard[data-drop-target] {
+  outline: 2px solid var(--ant-color-primary);
+}
+
+.tickDragHandle:not(:disabled) {
+  cursor: grab;
+  touch-action: none;
+}
+
+.tickDragHandle:not(:disabled):active {
```
eglitise: Also I'm not sure this style is needed, since the grabbing cursor is still applied without it (both in the Electron app and browser version in Firefox)

lucasmariano003-wq:
Updated through `d40b40c`.

Applied the requested simplifications: the grip is now a direct `IconGripVertical` with `handleRef` and the tick ID as its accessible label; the button, tooltip, translation key, `group`, and `isSortable` checks are removed. The drop handler checks cancellation, a valid target index, and a nonzero offset. I also simplified the styles, removed card hover, clarified the plugin configuration, and updated the variable names, tests, and documentation. The direct `@dnd-kit/dom` dependency is removed; a clean `npm ci` resolves version 0.5.0 through `@dnd-kit/react`.

Two small adjustments remain based on Chromium browser testing:

- With one action, the SVG is hidden using `visibility="hidden"` but stays mounted. Removing it makes dnd-kit fall back to the card container and mark it `aria-disabled`, exposing its controls as disabled too. Keeping the hidden handle fixes this, including when deleting an action during a drag.
- Focus restoration runs only after a successful keyboard move. Without it, dropping the SVG with `dropAnimation: null` leaves focus on `document.body`. Escape already preserves focus and needs no manual restoration.

Validation passed: 207 unit tests, 7 integration tests, lint, formatting, and all three builds. Browser checks covered mouse and keyboard reordering, destination focus, Escape, transitions between one and two actions, deletion during drag, editable single-action fields, and save/reopen with action values and coordinates preserved. All nine published files match the tested revision. Electron was built but not launched.

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorTickCard.jsx
```diff
@@ -59,23 +59,45 @@ const GestureEditorTickCardHeaderButtons = ({tick, pointers, setPointers, select
...
+  pointers,
+  setPointers,
+  selectedTick,
+  selectTick,
+  unselectTick,
+}) => {
+  return (
+    <Card
+      className={isDropTarget ? `${styles.tickCard} ${styles.tickCardDropTarget}` : styles.tickCard}
+      // Keep the hidden handle mounted so dnd-kit does not disable the whole single-action card.
```
eglitise: nit: the card isn't disabled, it's only marked as `aria-disabled`, which is an important distinction

eglitise (approved):
Nice work 👍 
Please also run `npm run lint:fix` to resolve the lint warnings.

lucasmariano003-wq:
Done in `568ba1c`. Ran `npm run lint:fix`, formatted the concise component return, and corrected the comment to say the card is marked as `aria-disabled`, rather than disabled.

`npm run lint -- --deny-warnings` and formatting pass, as do all 207 unit tests, 7 integration tests, and browser/Electron/plugin builds. The published tree was fetched back and matches the tested revision.

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorTickCard.jsx
```diff
@@ -59,10 +59,30 @@ const GestureEditorTickCardHeaderButtons = ({tick, pointers, setPointers, select
...
+}) => (
   <Card
-    hoverable={true}
-    className={styles.tickCard}
+    className={isDropTarget ? `${styles.tickCard} ${styles.tickCardDropTarget}` : styles.tickCard}
+    // Keep the hidden handle mounted so dnd-kit does not mark the whole card as aria-disabled.
+    title={
+      <IconGripVertical
+        size={18}
+        aria-label={tick.id}
```
mykola-mokhnach, line 80: The drag handle's `aria-label` is the raw `tick.id` rather than a translated, descriptive label. The other icon buttons in this file (`togglePickerLabel`, `deleteLabel`) use `t()`-translated labels — for consistency and localization, this should probably do the same (e.g. something like `t('dragToReorder')`) instead of exposing the internal id to screen readers.

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorTickCard.jsx
```diff
@@ -59,10 +59,30 @@ const GestureEditorTickCardHeaderButtons = ({tick, pointers, setPointers, select
...
+  pointers,
+  setPointers,
+  selectedTick,
+  selectTick,
+  unselectTick,
+}) => (
   <Card
-    hoverable={true}
-    className={styles.tickCard}
+    className={isDropTarget ? `${styles.tickCard} ${styles.tickCardDropTarget}` : styles.tickCard}
```
mykola-mokhnach, line 75: `hoverable={true}` was dropped from the `Card` here with no replacement, which removes antd's hover affordance (shadow/cursor) from every tick card. Was this intentional, or a side effect of the className rewrite?

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorPointerTabContents.jsx
```diff
@@ -40,30 +45,50 @@ const AddNewTickButton = ({id, pointers, setPointers}) => {
...
-);
+}) => {
+  const dragDisabled = pointer.ticks.length < 2;
+  const {ref, handleRef, isDropTarget} = useSortable({
+    id: tick.id,
+    index,
+    disabled: dragDisabled,
+    // Replace the defaults to remove OptimisticSortingPlugin and commit the new order
+    // only on drop. Disable the drop animation, which is confusing with position-based
+    // tick IDs that are reassigned when the order changes.
```
mykola-mokhnach, line 64: Nit: this 3-line comment could be tightened to 1-2 lines, e.g.:
```
// Commit the new order only on drop (no optimistic sorting), and skip the drop
// animation since it's confusing with position-based tick IDs.
```

mykola-mokhnach (commented):
A few small things I noticed while reviewing the drag-and-drop reordering — nothing blocking, just cleanup suggestions.

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorPointerTabContents.jsx
eglitise: This is fine, it was intentionally expanded to explicitly mention the removal of OptimisticSortingPlugin, since otherwise the purpose of this block is unclear: https://github.com/appium/appium-inspector/pull/3152#discussion_r4000628183

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorTickCard.jsx
eglitise: Intentional change: https://github.com/appium/appium-inspector/pull/3152#discussion_r4000637048

app/common/renderer/components/SessionInspector/GesturesTab/GestureEditor/GestureEditorTickCard.jsx
eglitise: Agreed - I had suggested removing the translation key earlier, but if the `aria-label` is to be retained, it's better to have a key still. Something like `"dragToReorder": "Drag to Reorder"`

eglitise: I'll add this in a follow-up PR.

PR #3155, opened 2026-09-08
fix: adjust offset calculation for Safari webviews

eglitise:
Decided to work on this for a bit, as it would help addressing comments in #3145.

* Webview offset calculations are extracted into a separate method for better clarity. Arguably they could also be further split for iOS and Android, but I didn't want to spend too much time on this.
* Android calculations were working fine, so they were not changed
* iOS calculations now check if the current active app is Safari, not just if it was set as the `browserName`
* iOS was checked in portrait mode (all 3 tab styles) and landscape mode, on iPhone 17 and 17e simulators. Landscape still has an issue with the scale factor being too large, but the offsets are at least correct now.

It's still far from a properly usable mode when scrolling the page, but again, the goal here was just to tune the offsets.

### DIFF
--- a/app/common/renderer/lib/appium/inspector-driver.js
+++ b/app/common/renderer/lib/appium/inspector-driver.js
@@ -16,9 +16,10 @@ const {TAP, SWIPE, GESTURE} = SCREENSHOT_INTERACTION_MODE;
 // Selector for the Android webview - includes the correct top and bottom boundaries
 const ANDROID_WEBVIEW_SELECTOR = 'android.webkit.WebView';
 const IOS_SAFARI_BUNDLE_ID = 'com.apple.mobilesafari';
-// iOS Safari portrait tabbar is shown in all tab styles, but only relevant for the top style
-const IOS_SAFARI_PORTRAIT_TABBAR_SELECTOR =
-  '**/XCUIElementTypeOther[`name == "CapsuleViewController"`]/XCUIElementTypeOther';
+// iOS Safari portrait topbar, includes system statusbar and Safari tabbar (in top tab style)
+const IOS_SAFARI_PORTRAIT_TOPBAR_SELECTOR =
+  '**/XCUIElementTypeOther[`name CONTAINS "SafariWindow"`]' +
+  '/XCUIElementTypeOther/XCUIElementTypeOther/XCUIElementTypeOther[2]';
 // iOS Safari landscape tabbar is only shown when scrolled upwards, and is auto-hidden otherwise
 const IOS_SAFARI_LANDSCAPE_TABBAR_SELECTOR = '**/XCUIElementTypeOther[`name == "Toolbar"`]';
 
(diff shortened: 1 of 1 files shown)


app/common/renderer/lib/appium/inspector-driver.js
```diff
@@ -14,10 +15,13 @@ const {TAP, SWIPE, GESTURE} = SCREENSHOT_INTERACTION_MODE;
 
 // Selector for the Android webview - includes the correct top and bottom boundaries
 const ANDROID_WEBVIEW_SELECTOR = 'android.webkit.WebView';
-// Selector for the iOS status bar and Safari address bar - not always present
-const IOS_TOP_CONTROLS_SELECTOR =
+const IOS_SAFARI_BUNDLE_ID = 'com.apple.mobilesafari';
+// iOS Safari portrait topbar, includes system statusbar and Safari tabbar (in top tab style)
```
mykola-mokhnach, line 19: this definitely does not look very reliable %)

app/common/renderer/lib/appium/inspector-driver.js
```diff
@@ -529,4 +454,83 @@ export default class InspectorDriver {
...
+   * Calculates and applies offsets for the current webview.
+   * Not applicable in non-webview contexts.
+   */
+  async recalculateWebviewViewports(windowSize, currentContext) {
+    let webviewTopOffset, webviewLeftOffset;
+    const automationName = this.driver.capabilities?.automationName?.toLowerCase();
+
+    if ([DRIVERS.UIAUTOMATOR2, DRIVERS.ESPRESSO].includes(automationName)) {
+      // on Android, find the root webview element and use its X and Y startpoints
+      const webview = await this.fetchElement({
```
mykola-mokhnach, line 468: actually on Android there is a more reliable way to detect a web view: https://github.com/appium/appium-uiautomator2-driver/blob/0e2aa5668cb50c7e2bbe2332f6561e3374526c4b/lib/commands/viewport.ts#L275

app/common/renderer/lib/appium/inspector-driver.js
```diff
@@ -529,4 +454,83 @@ export default class InspectorDriver {
...
+          try {
+            // to minimize the endpoint call which gets error in newer chromedriver.
+            const sessionDetails = await this.driver.getSession();
+            // in case driver does not support mobile:getSystemBars
+            webviewTopOffset = sessionDetails.viewportRect.top;
+          } catch {}
+        }
+      }
+    } else if (automationName === DRIVERS.XCUITEST) {
+      // mobile:activeAppInfo exists since XCUITest 2.126.0 (pre-Appium 2)
```
mykola-mokhnach, line 491: on ios https://github.com/appium/appium-xcuitest-driver/blob/a09d53d56ab6d07af451335fda7b1aaff7a2e517/lib/commands/web-native-bridge.ts#L401 could be used to fetch web view coordinates

app/common/renderer/lib/appium/inspector-driver.js
mykola-mokhnach: currently it's not exposed via public api, but we can make it available via execute method

eglitise: Sure, it would be nice to have something in a similar format to what XCUITest does

app/common/renderer/lib/appium/inspector-driver.js
eglitise: True, but it's more accurate than the currently used selector

app/common/renderer/lib/appium/inspector-driver.js
eglitise: this would be a bit trickier to integrate since it only works in web context, whereas the current calculations are all done in native context. but I agree it's worth keeping in mind

PR #3165, opened 2026-09-12
feat: add dropdown for selecting active window strategy (UiAutomator2)

eglitise:
Following up to #2922, this PR adds a similar element in the app header, allowing users to change the strategy for selecting the active application window. It makes use of the `enableTopmostWindowFromActivePackage` and `enableMultiWindows` settings to support selection between 3 separate strategies.

Additionally, #2922 incorrectly implemented a dependency on `enableMultiWindows` (which is not related to switching displays), so it is fixed here. The display switcher now only depends on the `currentDisplayId` setting, and as a result, both the display and window switcher can be intuitively combined:

https://github.com/user-attachments/assets/29892c8b-6681-41a9-a19d-dd7fd9ef36ec

And because the enabled states depend on the setting values, as introduced in #3131, they can be set at session start:

https://github.com/user-attachments/assets/10f6c64f-e458-49c3-996e-0baf92bcabfd


app/common/renderer/components/SessionInspector/Header/DriverControls/UiA2DisplayControls.jsx
```diff
@@ -0,0 +1,94 @@
...
+            value: id,
+            label: name ? `${name} (ID ${id})` : id,
+          }))}
+        />
+      )}
+    </Space.Compact>
+  );
+};
+
+export default UiA2DisplayControls;
```
mykola-mokhnach, line 94: I usually try to avoid default exports and only use named ones

mykola-mokhnach (approved):

app/common/renderer/components/SessionInspector/Header/DriverControls/UiA2DisplayControls.jsx
eglitise: Defaults tend to be more accepted for React where there's usually only one export per file. Named exports are still supported just fine, but it's best to keep consistent.

